PR #20, opened 2019-08-07
New r.object.thickness addon module

pzatelli:
r.object.thickness evaluates minimum, maximum and mean thickness of objects of a given category on a raster map, in map units and pixels. It can be used to set the moving window width for a filter, such as the ones used in r.neighbors.
It has been developed to automatically evaluate the minimum filter size for removing text and symbols from digitized historical maps with the r.fill.category addon.
Examples of its usage are reported in a paper currently under review for open access publication.


neteler:
Please also add the module to the list in
grass7/raster/Makefile

PR #25, opened 2019-08-31
Py3, overwrite test, speedup

ninsbl:
Adjust map() function to Python 3
Replace use of mask with r.reclass (now user mask can be present)
Add overwrite to test
Speedup by dropping r.univar and mask removal, (testsuite now finishes in 3 instead of 8 seconds on my laptop)
Testsuite runs successfully after changes


ninsbl:
OK, comments removed. Saw at that occasion that I forgot to generate the random points from reclass map (instead of the input map). This is now fixed too.

PR #30, opened 2019-09-15
WIP: Simplify Makefile to build all addon subdirs

ninsbl:
Please test this modified example Makefile for raster addons in GRASS 7.
After the modification, all subdirs are listed and thus build by default, so addon-devs do not have to add new modules manually (and deleted / moved addons do not have to be removed from the Makefile). If it is an acceptable solution, Makefiles for vector imagery and so on can be adjusted accordingly.

It would however probably require to find a solution for "deprecated" addons.

Not sure if it works with all versions of make...

See: https://trac.osgeo.org/grass/ticket/3901


grass7/raster/Makefile
```diff
@@ -13,130 +13,7 @@ OLD = \
 # unclear
```
neteler:
```suggestion
```

grass7/raster/Makefile
```diff
@@ -13,130 +13,7 @@ OLD = \
...
 #	r.convert \
```
neteler:
```suggestion
```

neteler:
The same works for the vector/Makefile etc., pls consider to expand this PR

ninsbl:
> The same works for the vector/Makefile etc., pls consider to expand this PR

Yes, it would be the plan to apply this to the other AddOn types (vector, imagery, ...) as well.

Now I also committed a solution for filtering out deprecated AddOns; if not building those AddOns by default is the preferred solution. However, maybe it is better to add a DEPRECATED section to the manual instead (and have addons and manuals built)?

neteler:
Please also update misc/Makefile, then it should be complete (and there is a conflict in grass7/raster/Makefile)

PR #31, opened 2019-09-20
i.sentinel.import: fix register_output for S2 L2A products

landam:
This PR fixes `i.sentinel.import` when creating `register_output` for Sentinel 2 L2A products. See #29

### DIFF
--- a/grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
+++ b/grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
@@ -273,8 +273,8 @@ class SentinelImporter(object):
             if node is not None:
                 try:
                     # check S2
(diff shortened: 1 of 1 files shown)


grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
```diff
@@ -320,11 +320,12 @@ def create_register_file(self, filename):
                     ts=timestamp_str
                 ))
                 if has_band_ref:
+                    band_ref = re.match(
+                        r'.*_B([0-1][0-9]).*', map_name
```
veroandreo:
```suggestion
                        r'.*_B([0-1][0-9]).*', map_name
```
does this also consider B08A??

landam: @veroandreo Good catch, fixed in  a600ba5

PR #40, opened 2019-10-10
wqfile -> open

jachym:
`file` is not available in python 3, use `open`


grass7/vector/v.lidar.mcc/v.lidar.mcc.py
```diff
@@ -109,7 +109,7 @@ def main():
     # this would be done easily with StringIO
     # but it doesn't work with subprocess
     if not grass.debug_level():
-        nuldev = file(os.devnull, 'w')
+        nuldev = opennn(os.devnull, 'w')
```
neteler:
```suggestion
        nuldev = open(os.devnull, 'w')
```

PR #45, opened 2019-10-14
Simplify Makefile to build all i.sentinel subdirs

neteler:
Approach following #30

If ok, also the Makefiles of 
- imagery/i.modis
- raster/r.futures
- raster/r.green
- raster/r.pi
to be updated accordingly.


landam:
@petrasovaa Any opinion? Thanks.

PR #49, opened 2019-10-27
Make MATRIX_PROPERTIES a list

neteler:
fixes "ValueError: 9 is not in list", issue https://github.com/NikosAlexandris/i.fusion.hpf/issues/14

### DIFF
--- a/grass7/imagery/i.fusion.hpf/constants.py
+++ b/grass7/imagery/i.fusion.hpf/constants.py
@@ -29,7 +29,7 @@ RATIO_RANGES = (
 
 KERNEL_SIZES = (5, 7, 9, 11, 13, 15)
 
-MATRIX_PROPERTIES = zip(RATIO_RANGES, KERNEL_SIZES)
+MATRIX_PROPERTIES = list(zip(RATIO_RANGES, KERNEL_SIZES))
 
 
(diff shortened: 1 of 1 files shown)


NikosAlexandris:
Markus, before we merge, what is different in Python 3? Does the error "ValueError: 9 is not in list" not occur with Python 2? Maybe "fix MATRIX_PROPERTIES for Python 3" is a misleading message?

hellik:
see e.g.

https://docs.python.org/3/library/functions.html#zip

>  zip(*iterables)
> 
>     Make an iterator that aggregates elements from each of the iterables.

and e.g.

https://stackoverflow.com/questions/31683959/the-zip-function-in-python-3

```
Unlike in Python 2, the zip function in Python 3 returns an iterator. Iterators can only be exhausted (by something like making a list out of them) once. 
```

neteler:
> Markus, before we merge, what is different in Python 3? Does the error "ValueError: 9 is not in list" 
not occur with Python 2?

I think so, at least I didn't get this in the past before migrating to Python 3 (i.e., GRASS GIS 7.8+).

> Maybe "fix MATRIX_PROPERTIES for Python 3" is a misleading message?

No idea, I just need a working version in G78.

PR #50, opened 2019-10-28
i.sentinel.download: using the polygon of the map instead of the region

AnikaBettge:
- I changes the code so that the polygon of the map of AOI is used instead of only the region of this map

Is it okay to replace the region part or should I use a flag for the polygon or region variant?

### DIFF
--- a/grass7/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
+++ b/grass7/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
@@ -120,6 +120,10 @@
 #% description: List filtered products and exit
 #% guisection: Print
 #%end
+#%flag
+#% key: b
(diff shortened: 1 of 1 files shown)


neteler:
In the end we have three cases related to the region:

- map extent of map=vector_AOI
- current region
- new: true polygon in the map=vector_AOI

Probably we need all options - the question is how to "communicate" this to the user.

veroandreo:
> In the end we have three cases related to the region:
> 
>     * current region (default)
> 
>     * bbox of vector map given by user (which can be different from current region)
> 
>     * new: true vector polygon of the map given by user
> 
> 
> Probably we need all options - the question is how to "communicate" this to the user.

I would keep the default as is and use a flag for the new feature (cool by the way :)) so if the user passes a vector map with the AOI and sets the flag, then the borders of the polygon are used.

PR #52, opened 2019-11-01
r.null.all: add wrapper around r.null for multiple maps

wenzeslaus:
r.null.all uses g.list wrapper to select all raster maps or a subset.
g.list syntax options are supported but are driven by an option,
not flags with an explicit default value for selecting all maps.

r.null flags are supported as flags, but the dry run mode of r.null.all
does not take these into account.

### DIFF
--- a/grass7/raster/r.null.all/r.null.all.html
+++ b/grass7/raster/r.null.all/r.null.all.html
@@ -45,7 +45,7 @@ r.null.all setnull=0 exclude=".*1" matching=extended -d
 
 <em>
   <a href="r.null.html">r.null</a>,
-  <a href="g.proj.all.html">g.proj.all</a>
+  <a href="g.proj.all.html">g.proj.all</a>,
   <a href="g.copyall.html">g.copyall</a>,
   <a href="g.rename.many.html">g.rename.many</a>
 </em>
--- a/grass7/raster/r.null.all/r.null.all.html
+++ b/grass7/raster/r.null.all/r.null.all.html
@@ -1,0 +1,57 @@
+<h2>DESCRIPTION</h2>
+
+<em>r.null.all</em> manages NULL (no-data) values in all raster maps
+in the current mapset. Selection can be modified using <b>pattern</b>
+and <b>exclude</b> options. The option <b>matching</b> specifies
+the type of search pattern use. Python users will find the extended
+regular expression syntax (marked as <em>extended</em>) as most
+familiar, while Bash users may want to use <em>wildcards</em>
+(glob patterns).
+
+<h2>EXAMPLE</h2>
+
+All the following examples are using the <b>-d</b> flag to run in a
+<em>dry run</em> mode so that no maps are actually modified.
+
+Set all values 1 to NULL in all raster maps in the current mapset:
+
+<div class="code"><pre>
+r.null.all setnull=1 -d
+</pre></div>
+
+Change all NULL to zero in all raster maps in the current mapset
+which begin with letter t and their name contains at least one other
+character (using the extended regular expressions):
+
+<div class="code"><pre>
+!r.null.all null=0 pattern="^t.+" m=extended -d
+</pre></div>
+
+Set all values 0 to NULL in raster maps in the current mapset
+which do not end with the digit 1 (using the wildcards syntax):
+
+<div class="code"><pre>
+r.null.all setnull=0 exclude="*1" matching=wildcards -d
+</pre></div>
+
+Set all values 0 to NULL in raster maps in the current mapset
+which do not end with the digit 1 (using extended regular expressions):
+
+<div class="code"><pre>
+!r.null.all setnull=0 ex=".*1" m=extended -d
+</pre></div>
+
+<h2>SEE ALSO</h2>
+
+<em>
+  <a href="r.null.html">r.null</a>,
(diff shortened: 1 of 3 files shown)


grass7/raster/r.null.all/r.null.all.py
```diff
@@ -0,0 +1,190 @@
...
+#               This program is free software under the GNU General
+#               Public License (>=v2). Read the file COPYING that
+#               comes with GRASS for details.
+#
+#############################################################################
+
+#%module
+#% description: Manages NULL values of raster maps in a mapset or their subset
+#% keyword: raster
+#% keyword: map management
```
neteler: Probably the order of keyworks should follow that of r.null (?) or both be aligned.

grass7/raster/r.null.all/r.null.all.py
```diff
@@ -0,0 +1,190 @@
...
+# COPYRIGHT:    (C) 2019 by Vaclav Petras and the GRASS Development Team
+#
+#               This program is free software under the GNU General
+#               Public License (>=v2). Read the file COPYING that
+#               comes with GRASS for details.
+#
+#############################################################################
+
+#%module
+#% description: Manages NULL values of raster maps in a mapset or their subset
```
neteler:
```suggestion
#% description: Manages NULL values of raster maps in a mapset or their subset.
```

grass7/raster/r.null.all/r.null.all.html
```diff
@@ -0,0 +1,57 @@
...
+
+<div class="code"><pre>
+!r.null.all setnull=0 ex=".*1" m=extended -d
+</pre></div>
+
+<h2>SEE ALSO</h2>
+
+<em>
+  <a href="r.null.html">r.null</a>,
+  <a href="g.proj.all.html">g.proj.all</a>
```
neteler:
```suggestion
  <a href="g.proj.all.html">g.proj.all</a>,
```

grass7/raster/r.null.all/r.null.all.html
```diff
@@ -0,0 +1,57 @@
...
+  <a href="g.copyall.html">g.copyall</a>
```
neteler:
```suggestion
  <a href="g.copyall.html">g.copyall</a>,
```

grass7/raster/r.null.all/r.null.all.html
```diff
@@ -0,0 +1,57 @@
...
+
+<div class="code"><pre>
+r.null.all setnull=0 exclude="*1" matching=wildcards -d
+</pre></div>
+
+Set all values 0 to NULL in raster maps in the current mapset
+which do not end with the digit 1 (using extended regular expressions):
+
+<div class="code"><pre>
+!r.null.all setnull=0 ex=".*1" m=extended -d
```
neteler:
```suggestion
r.null.all setnull=0 exclude=".*1" matching=extended -d
```
(please avoid abbreviations in manual page examples)

grass7/raster/r.null.all/r.null.all.html
```diff
@@ -0,0 +1,57 @@
...
+<div class="code"><pre>
+r.null.all setnull=1 -d
+</pre></div>
+
+Change all NULL to zero in all raster maps in the current mapset
+which begin with letter t and their name contains at least one other
+character (using the extended regular expressions):
+
+<div class="code"><pre>
+!r.null.all null=0 pattern="^t.+" m=extended -d
```
neteler:
```suggestion
r.null.all null=0 pattern="^t.+" matching=extended -d
```
(please avoid abbreviations in manual page examples)

grass7/raster/r.null.all/r.null.all.html
```diff
@@ -0,0 +1,57 @@
...
+
+<em>r.null.all</em> manages NULL (no-data) values in all raster maps
+in the current mapset. Selection can be modified using <b>pattern</b>
+and <b>exclude</b> options. The option <b>matching</b> specifies
+the type of search pattern use. Python users will find the extended
+regular expression syntax (marked as <em>extended</em>) as most
+familiar, while Bash users may want to use <em>wildcards</em>
+(glob patterns).
+
+<h2>EXAMPLE</h2>
```
neteler:
```suggestion
<h2>EXAMPLES</h2>
```

grass7/raster/r.null.all/r.null.all.py
wenzeslaus: r.null has only `raster, null data`, so I think we need to fix it.

The questions here are more general: 1) Should be we have synonyms in keywords, i.e., both `null data` and `no-data` 2) Should we repeat the words from the name of module, i.e., `null` or `null data`? 3) Should different spellings be included, e.g., `no-data` and `NoData`?   I'm saying yes first two questions, so that's why I included `raster, map management, null data, no-data`. What do you think?

grass7/raster/r.null.all/r.null.all.py
wenzeslaus: We should probably fix this through out the code base. In addons, most don't have the period, in core, most have but not all. `--script` gives the period.

    grep -IrnE -m 1 description: path/to/code | grep "[\.]$" | wc

The question is if we want the period there. It is not even a sentence.

If I recall correctly, the current status is that the code processing these just adds (or perhaps also) removes the period as needed.

PR #55, opened 2019-11-12
Make most grass7 modules compilable

HuidaeCho:
Migrate to 7.9 and python3 (selected modules only)


neteler:
The shebang changes
```
- #!/usr/bin/env python
+ #!/usr/bin/env python3
```
are not (yet) desired since
- the addons need to remain compatible to Python 2 for a while (to support GRASS GIS 7.x < 7.8.0), and
- g.extension takes care of such shebang update in GRASS GIS >= 7.8.0.

Would you mind to remove these related changes from your PR?

Concerning deprecated modules, this Makefile simplification approach could be adopted: #45

HuidaeCho:
I see. Many of those Python 2 scripts failed because I didn't have required modules installed for Python 2. Looks like I need my own workaround. I'll revert the shebang changes.

PR #56, opened 2019-11-12
i.sentinel.download: catch WKT too long with -b flag

neteler:
RFC 2616 HTTP/1.1 imposes a 2000 char limit on WKT when sent via HTTP GET (sentinelsat).

Raise fatal error when AOI map (-b flag) has too many vertices, recommending to use 'v.generalize' to simplify the boundaries.

### DIFF
--- a/grass7/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
+++ b/grass7/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
@@ -173,8 +173,12 @@ def get_aoi(vector=None):
     if '+proj' not in kv:
         gs.fatal('Unable to get AOI: unprojected location not supported')
     geom_dict = gs.parse_command('v.out.ascii', format='wkt', **args)
-    geom = geom = [key for key in geom_dict][0]
+    num_vertices = len(str(geom_dict.keys()).split(','))
+    geom = [key for key in geom_dict][0]
     if kv['+proj'] != 'longlat':
+        gs.message(_("Generating WKT from AOI map ({} vertices)...").format(num_vertices))
+        if num_vertices > 500:
+            gs.fatal(_("AOI map has too many vertices to be sent via HTTP GET (sentinelsat). Use 'v.generalize' to simplify the boundaries"))
         coords = geom.replace('POLYGON((', '').replace('))', '').split(', ')
         poly = 'POLYGON(('
         poly_coords = []
@@ -184,6 +188,11 @@ def get_aoi(vector=None):
             for key in coord_latlon:
                 poly_coords.append((' ').join(key.split('|')[0:2]))
         poly += (', ').join(poly_coords) + '))'
+        # Note: poly must be < 2000 chars incl. sentinelsat request (see RFC 2616 HTTP/1.1)
+        if len(poly) > 1850:
+            gs.fatal(_("AOI map has too many vertices to be sent via HTTP GET (sentinelsat). Use 'v.generalize' to simplify the boundaries"))
+        else:
+            gs.message(_("Sending WKT from AOI map to ESA..."))
         return poly
     else:
         return geom
--- a/grass7/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
+++ b/grass7/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
@@ -167,11 +167,10 @@ def get_aoi(vector=None):
     if vector:
         args['input'] = vector
 
-    topology = gs.parse_command('v.info', map=vector, flags='t')
-    if int(topology['areas']) < 1:
+    if gs.vector_info_topo(vector)['areas'] <= 0:
         gs.fatal(_("No areas found in AOI map <{}>...").format(vector))
-    elif  int(topology['areas']) > 1:
-        gs.warning(_("More than one areas found in AOI map <{}>. \
+    else:
+        gs.warning(_("More than one area found in AOI map <{}>. \
                       Using only the first area...").format(vector))
 
     # are we in LatLong location?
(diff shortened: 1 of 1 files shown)


grass7/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
```diff
@@ -173,8 +173,12 @@ def get_aoi(vector=None):
     if '+proj' not in kv:
         gs.fatal('Unable to get AOI: unprojected location not supported')
     geom_dict = gs.parse_command('v.out.ascii', format='wkt', **args)
```
ninsbl: Looks good. One suggestions for a check and warning message,
in case users provide a unsuitable vector map (with more than one or no areas)...

```
topology = gs.parse_command('v.info', map=vector, flags='t')
if int(topology['areas']) < 1:
    gs.fatal(_("No areas found in AOI map <{}>...").format(vector))
elif  int(topology['areas']) > 1:
    gs.warning(_("More than one areas found in AOI map <{}>. \
Using only the first...").format(vector))
```

neteler: Thanks, PR updated accordingly.

I just wonder if a limit is needed in the v.in.ascii call to really only take the first area?

grass7/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
```diff
@@ -167,14 +167,25 @@ def get_aoi(vector=None):
     if vector:
         args['input'] = vector
 
+    topology = gs.parse_command('v.info', map=vector, flags='t')
+    if int(topology['areas']) < 1:
+        gs.fatal(_("No areas found in AOI map <{}>...").format(vector))
+    elif  int(topology['areas']) > 1:
+        gs.warning(_("More than one areas found in AOI map <{}>. \
+                      Using only the first area...").format(vector))
```
neteler: I wonder if a modification of the `v.out.ascii` call is needed to really restrict to  "Using only the first area" ?

ninsbl: I dont`t think so, cause you limit it to the first geometry here:
```geom = [key for key in geom_dict][0]```

An alternative would be to use a convex hull (v.hull) over input geometries, or loop over polygons...

In principle it would be possible that the first geometry is not an area.
If you want to be 100% sure, check that too:
```if not geom.startswith('POLYGON'):
    gs.fatal(_(''Geometry used as AOI is no area.))```
But that is a corner case I guess...

File: grass7/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py [hunk 3]
```diff
@@ -167,14 +167,25 @@ def get_aoi(vector=None):
     if vector:
         args['input'] = vector
 
+    topology = gs.parse_command('v.info', map=vector, flags='t')
```
landam: There is `vector_info_topo()` in GRASS scripting library. Then no casting is needed.

grass7/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
```diff
@@ -167,14 +167,25 @@ def get_aoi(vector=None):
...
+    elif  int(topology['areas']) > 1:
+        gs.warning(_("More than one areas found in AOI map <{}>. \
+                      Using only the first area...").format(vector))
+
     # are we in LatLong location?
     s = gs.read_command("g.proj", flags='j')
     kv = gs.parse_key_val(s)
     if '+proj' not in kv:
         gs.fatal('Unable to get AOI: unprojected location not supported')
     geom_dict = gs.parse_command('v.out.ascii', format='wkt', **args)
```
landam: In the future I would love to seem pyGRASS to be used more widely. This would a nice showcase. But it's not blocker for this PR.

grass7/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
neteler: I also thought of convex hull but it is not necessarily the right approach - depends on the use case.

grass7/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
neteler: ok, will check!

grass7/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
neteler: Agreed - the first case I have addressed. More suggestions welcome.

grass7/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
AnikaBettge: Alternative idea: loop over all areas

grass7/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
```diff
@@ -167,14 +167,25 @@ def get_aoi(vector=None):
     if vector:
         args['input'] = vector
 
+    if gs.vector_info_topo(vector)['areas'] <= 0:
+        gs.fatal(_("No areas found in AOI map <{}>...").format(vector))
+    else:
+        gs.warning(_("More than one area found in AOI map <{}>. \
+                      Using only the first area...").format(vector))
```
neteler:
```suggestion
                      Using only the first area... (in case of adjacent areas consider to use 'v.dissolve')").format(vector))
```

neteler:
I'll merge the current state now

PR #65, opened 2019-11-20
fix write_metadata and o-flag

ninsbl:
Address: https://trac.osgeo.org/grass/ticket/3979


neteler:
Unfortunately I still get the same error (see trac ticket)

PR #68, opened 2019-11-24
I sentinel preproc enhance

ninsbl:
Pass import parameters from i.sentinel.import and fix import issue


ninsbl:
If there are no objections I will merge tonight...

PR #76, opened 2019-12-18
i.sentinel.import: added extent parameter

neteler:
With extent=region the import can be reduced to the current region rather than full scene import.

### DIFF
--- a/grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.html
+++ b/grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.html
@@ -10,6 +10,10 @@ in the <b>input</b> directory. The number of scene files can be optionally
 reduced by the <b>pattern_file</b> option. In this option, a regular expression 
 for filtering the file names should be given, e.g. "MSIL2A.*T32VNR_2019" for 
 importing only level 2A products for tile T32VNR from 2019.
+<p>
+By default <em>i.sentinel.import</em> imports the full scene. Optionally, the
+import can be reduced to the computational region extent with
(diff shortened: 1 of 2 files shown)


grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.html
```diff
@@ -10,6 +10,10 @@ <h2>DESCRIPTION</h2>
 reduced by the <b>pattern_file</b> option. In this option, a regular expression 
 for filtering the file names should be given, e.g. "MSIL2A.*T32VNR_2019" for 
 importing only level 2A products for tile T32VNR from 2019.
+<p>
+By default <em>i.sentinel.import</em> imports the full scene. Optionally, the
+import can be reduced to the computational region extent with
+<b>extent=region</b>.
```
neteler:
```suggestion
<b>extent=region</b>. If the current region exceeds the extent of the Sentinel scene, then
uncovered pixels are stored as "no data" (NULL) values.
```

neteler:
Time test:

```
g.region raster=ortho_2001_t792_1m -p res=10 -a
i.sentinel.import input=~/tmp/ pattern='B(03|04)' memory=2000 -r extent=region
```

- without `extent=region` (original behaviour, full scene): 440 sec
- with `extent=region` (new & optional, partial scene): 39 sec

neteler:
> I would probably clarify that this does not apply to external data (-l) but "importing" is probably clear enogh, and it will not have reasonable effect anyway...

Right - I have now made `extent` and `-l` mutually exclusive.

Merging now, thanks for your comments!

PR #79, opened 2019-12-30
updated r.sim.terrain

baharmon:
Fixed bug in net difference calculation, added copy elevation to current mapset to prevent temporal framework registration error, and refactored code.


neteler:
Feel free to merge (since you are the author of this addon)

baharmon:
Markus, I don't have write access so I can't merge.

neteler:
> Markus, I don't have write access so I can't merge.

I see. Confusingly you were missing from the addons team in GH while being listed in contributors.csv. I have sent you a GH invitation, after acceptance from your side it should work.

PR #83, opened 2020-01-17
Bufferstats

lucadelu:
Some improvements in v.rast.bufferstats

### DIFF
--- a/grass7/vector/v.rast.bufferstats/v.rast.bufferstats.py
+++ b/grass7/vector/v.rast.bufferstats/v.rast.bufferstats.py
@@ -121,6 +121,10 @@ To Dos:
 #% description: Update columns if they already exist
 #%end
 
+#%flag
+#% key: r
+#% description: Remove columns without data
+#%end
 
 #%option G_OPT_F_OUTPUT
 #% description: Name for output file (if "-" output to stdout)
@@ -275,6 +279,7 @@ def main():
     update = flags['u']
     tabulate = flags['t']
     percent = flags['p']
+    remove = flags['r']
 
     mymapset = Mapset().name
     # Do checks using pygrass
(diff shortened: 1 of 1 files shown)


grass7/vector/v.rast.bufferstats/v.rast.bufferstats.py
```diff
@@ -98,16 +99,32 @@
 #% required : no
 #%end
 
+#%option
+#% key: resolution
+#% type: integer
+#% description: The region resolution to get better result in the analysis
```
neteler: This description is not very clear yet

ninsbl: The original idea was to use and align to the resolution of the current region... If the module dos not follow the current region I would prefer to change it in a way that it does...

grass7/vector/v.rast.bufferstats/v.rast.bufferstats.py
```diff
@@ -608,6 +634,18 @@ def main():
         # write results to file
         out.close()
 
+    if remove:
+        print("in remove")
```
ninsbl: This can be removed or a GRASS message can be used?

grass7/vector/v.rast.bufferstats/v.rast.bufferstats.py
lucadelu: > The original idea was to use and align to the resolution of the current region... If the module dos not follow the current region I would prefer to change it in a way that it does...

Ok, I'm going to remove this feature

grass7/vector/v.rast.bufferstats/v.rast.bufferstats.py
lucadelu: Removed and improved the next sentence.

lucadelu:
For my point of view it is ready to be merged

ninsbl:
Yes, agreed. Please merge. And thanks for your effort...

ninsbl:
Hi again, by coincidence I had an application for the module myself and now I tested the new percentage option a bit. It seems that percentage is computed for the entire current computational region (including all null cells) and not just for the area inside the buffer. And percentages for null cells are not reported, So, percentages in total are often at ~78%. That might be confusing for users.

Souldn`t we define percentages as relative to the area within each buffer?

veroandreo:
> Hi again, by coincidence I had an application for the module myself and now I tested the new percentage option a bit. It seems that percentage is computed for the entire current computational region (including all null cells) and not just for the area inside the buffer. And percentages for null cells are not reported, So, percentages in total are often at ~78%. That might be confusing for users.
> 
> Souldn`t we define percentages as relative to the area within each buffer?
I would expect so given that the module is "bufferstats" :)

PR #89, opened 2020-02-04
i.sentinel.import: avoid multiple unzipping on single run

landam:
Currently `i.sentinel.import` unzips input data several times on single run when `-n` flags is not given. Potentially 4 times (import imagery files, cloud mask, write metadata, create register file)! This PR forces unzipping input data only on the beginning (when importing imagery files). 

BTW, I wonder why `-n` is not a default behaviour?


landam:
> A more elegant way, that avoids unzipping and writing to disk, would be to use GDALs virtual file systems: https://gdal.org/user/virtual_file_systems.html
> So, basically `input=filename` in L296 would have to be replaced by `input="/vsizip/{}".format(filename) ` (and accordingly for L318)
> 
> Don`t know if it actually would be possible to read directly from sentinelhub with `/vsicurl/vsizip/`...

Let's create separated issue/PR for that.

PR #96, opened 2020-02-23
addons: Python 3 reformatting

neteler:
- tab/space indentation mess fixed
- print() fixes
- HTML code fixes


neteler:
This should fix numerous compilation issues here: https://wingrass.fsv.cvut.cz/grass78/x86_64/addons/grass-7.8.2/logs/

(to be verified on Monday, at 8:30 pm Prague time at next cronjob run)

PR #98, opened 2020-02-24
r.lfp: Port to python3

HuidaeCho:
Port to Python 3


grass7/raster/r.lfp/r.lfp.py
```diff
@@ -1,11 +1,12 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
neteler:
```suggestion
#!/usr/bin/env python
```

Please keep python here (g.extension is taking care of that during install), for consistency.
In future, when our support for Py2 will be dropped (once GRASS GIS 7.6 is out of support) then we can globally enforce python3 in the shebang of all addons.

HuidaeCho: So is it g.extension (or the user) who decides Python 2 or 3? Does it mean no mix of 2 and 3? Just curious.

neteler: It is different:
- in GRASS GIS < 7.8, g.extension will install it as-is
- in GRASS GIS >= 7.8, g.extension will replace `#!/usr/bin/env python` with `#!/usr/bin/env python3`

Like this a mixture is avoided and addons can still be installed in older GRASS GIS versions.

HuidaeCho: Which means, in terms of syntax, we should stick to Python 2 and avoid Python 3 only syntax. For example, no `f"{var}"`.

neteler: AFAIK yes... however, we should define a date to phase out Python 2.

PR #105, opened 2020-02-26
Replace norwegian character

ninsbl:


ninsbl:
Sorry, was an accident due to lack of git experience

PR #108, opened 2020-02-28
r.richdem.*: lazy import RICHDEM so manuals get built

ninsbl:
Currently, manuals for r.richdem.* are not built because richdem cannot be imported on built server.
Lazy importing richdem should solve this...


metzm:
Looks good to me. Should it be merged?

awickert:
Many thanks, all! I missed this while in the field, and appreciate the work to get the modules going.

PR #115, opened 2020-03-05
Adapt v.centerline for python 3

micha-silver:
Tested by clipping a small section of streams layer in nc_basic_spm_grass7.
Then that section was buffered and the buffer boundry was split at the ends,
creating two "bank lines". These lines were then used as input to v.centerline.


mlennert:
Thanks a lot for this Micha. This can be merged IMHO.

micha-silver:
I did only a few minor changes. Probably best to go ahead and merge, then we'll see if there are any ricochets that require further editing.

metzm:
I am merging this PR as suggested by @micha-silver.

mlennert:
Thanks a lot, Markus !

PR #120, opened 2020-03-18
r.maxent.lambdas manual: fix encoding and HTML

neteler:
* fix wrong encoding:
    * `iconv -f ISO-8859 -t UTF-8 r.maxent.lambdas.html > new.html ; mv new.html r.maxent.lambdas.html`
* fix wrong HTML structure


neteler:
Should fix the broken compilation (see https://wingrass.fsv.cvut.cz/grass78/x86_64/logs/)

PR #122, opened 2020-03-18
i.sentinel.mask fix parameter rules for only cloud computig

AnikaBettge:
Only computing the clouds does not need the mtd_file. For this the rules are modified


neteler:
I take liberty to merge, pls reopen if you disagree, @RobiFag

PR #123, opened 2020-03-19
i.sentinel.import: use footprint to delete nodata border from imported bands

griembauer:
work in progress: adds option to pass a vector footprint which is downloaded from i.sentinel.download. The footprint is then used to remove no-data from the imported bands.

### DIFF
--- a/grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
+++ b/grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
@@ -500,27 +500,28 @@ class SentinelImporter(object):
                     ))
                 fd.write(os.linesep)
 
-    def set_footprint(self,fp):
-        reg_name = "tmpregion_%s"%(str(os.getpid()))
-        gs.run_command('g.region',save=reg_name)
-        fp_dict={}
-        idlist = [x for x in gs.parse_command('v.db.select',map=fp)]
+    def set_footprint(self, fp):
+        reg_name = "tmpregion_%s" % (str(os.getpid()))
+        gs.run_command('g.region', save=reg_name)
+        fp_dict = {}
+        idlist = [x for x in gs.parse_command('v.db.select', map=fp)]
         cat_idx = idlist[0].split('|').index('cat')
         ident_idx = idlist[0].split('|').index('identifier')
 
         for item in idlist[1:]:
             tmp_key = item.split('|')[ident_idx]
-            key = "%s_%s"%(tmp_key.split('_')[5],tmp_key.split('_')[2])
+            key = "%s_%s" % (tmp_key.split('_')[5], tmp_key.split('_')[2])
             fp_dict[key] = item.split('|')[cat_idx]
-        fp_rast_name = "fp_rast_%s"%(str(os.getpid()))
+        fp_rast_name = "fp_rast_%s" % (str(os.getpid()))
         for map in imported_map_list:
-            gs.run_command('g.region',raster=map)
-            gs.run_command('v.to.rast',input=fp,cats=fp_dict['%s_%s'%(map.split('_')[0],map.split('_')[1])],use="val",value=1,memory=options['memory'], output=fp_rast_name)
-            gs.run_command('r.mapcalc',expression='tmp_%s = if(isnull(%s),null(),%s)'%(map,fp_rast_name,map))
-            gs.run_command('g.rename',raster='tmp_%s,%s'%(map,map))
-            gs.run_command('g.remove',type="raster",name=fp_rast_name,flags='f')
-        gs.run_command('g.region',region=reg_name)
-        gs.run_command('g.remove',type="region",name=reg_name,flags='f')
+            gs.run_command('g.region', raster=map)
+            gs.run_command('v.to.rast', input=fp, cats=fp_dict['%s_%s' % (map.split('_')[0], map.split('_')[1])], use="val",
+                           value=1, memory=options['memory'], output=fp_rast_name)
+            gs.run_command('r.mapcalc', expression='tmp_%s = if(isnull(%s),null(),%s)' % (map, fp_rast_name, map))
+            gs.run_command('g.rename', raster='tmp_%s,%s' % (map, map))
+            gs.run_command('g.remove', type="raster", name=fp_rast_name, flags='f')
+        gs.run_command('g.region', region=reg_name)
+        gs.run_command('g.remove', type="region", name=reg_name, flags='f')
 
 def main():
     global imported_map_list
--- a/grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
+++ b/grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
@@ -511,12 +511,11 @@ class SentinelImporter(object):
             key = "%s_%s" % (tmp_key.split('_')[5], tmp_key.split('_')[2])
             fp_dict[key] = item.split('|')[cat_idx]
         fp_rast_name = "fp_rast_%s" % (str(os.getpid()))
+        gs.use_temp_region()
         for map in imported_map_list:
             wherestring = "cat = "+ fp_dict['%s_%s'%(map.split('_')[0],map.split('_')[1])]
             reg = gs.parse_command("v.db.select", flags="r", map=fp, where=wherestring)
-            gs.use_temp_region()
             gs.run_command('g.region', align=map, n=reg["n"], s=reg["s"], e=reg["e"], w=reg["w"])
-
             gs.run_command('v.to.rast', input=fp, cats=fp_dict['%s_%s' % (map.split('_')[0], map.split('_')[1])], use="val",
                            value=1, memory=options['memory'], output=fp_rast_name)
             gs.run_command('r.mapcalc', expression='tmp_%s = round(if(isnull(%s),null(),%s))' % (map, fp_rast_name, map))
--- a/grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
+++ b/grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
@@ -52,6 +52,11 @@
 #% descriptions: region;extent of current region;input;extent of input map
 #% guisection: Filter
 #%end
+#%option G_OPT_V_INPUT
+#% key: footprints
+#% description: Name of vector input map with Sentinel-2 footprints to limit import extent to not-null data.
+#% required: no
+#%end
 #%option
 #% key: memory
 #% type: integer
@@ -112,6 +117,8 @@ from zipfile import ZipFile
 import grass.script as gs
 from grass.exceptions import CalledModuleError
 
+imported_map_list = []
+
 class SentinelImporter(object):
     def __init__(self, input_dir, unzip_dir):
         # list of directories to cleanup
@@ -143,7 +150,7 @@ class SentinelImporter(object):
                 shutil.rmtree(dirpath)
             except OSError:
                 pass
-            
+
     def filter(self, pattern=None):
         if pattern:
             filter_p = r'.*{}.*.jp2$'.format(pattern)
@@ -262,12 +269,12 @@ class SentinelImporter(object):
             gs.fatal(_("Flag -r requires GDAL library: {}").format(e))
         dsn = gdal.Open(filename)
         trans = dsn.GetGeoTransform()
-        
+
         ret = int(trans[1])
         dsn = None
 
         return ret
-    
+
     def _raster_epsg(self, filename):
         try:
             from osgeo import gdal, osr
@@ -277,7 +284,7 @@ class SentinelImporter(object):
 
         srs = osr.SpatialReference()
         srs.ImportFromWkt(dsn.GetProjectionRef())
-    
+
         ret = srs.GetAuthorityCode(None)
         dsn = None
 
@@ -289,6 +296,7 @@ class SentinelImporter(object):
 
     def _import_file(self, filename, module, args):
         mapname = self._map_name(filename)
+        imported_map_list.append(mapname)
         gs.message(_('Processing <{}>...').format(mapname))
         if module == 'r.import':
             args['resolution_value'] = self._raster_resolution(filename)
@@ -492,6 +500,8 @@ class SentinelImporter(object):
                     ))
                 fd.write(os.linesep)
 def main():
+    global imported_map_list
+
     importer = SentinelImporter(options['input'], options['unzip_dir'])
 
     importer.filter(options['pattern'])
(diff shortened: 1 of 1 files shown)


ninsbl:
Would'nt it be more efficient to import/link the QA bands (MSK_) rather than imporing and converting a vector map?

neteler:
Are they identical to the vector footprint?

ninsbl:
My bad, there are no raster QA bands in Sentinel-2 LC1.
However, NoData is supposed to be 0. So, a GDAL VRT (for linked data with r.external) or r.null (+ evtl r.clip) for imported data could be used...

griembauer:
@ninsbl Unfortunately, there are also areas within S-2 data that can be 0, e.g. water areas (in general, dark features). Using r.null therefore may leave gaps in the data.

grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
```diff
@@ -52,6 +52,11 @@
 #% descriptions: region;extent of current region;input;extent of input map
 #% guisection: Filter
 #%end
+#%option G_OPT_V_INPUT
+#% key: footprints
+#% description: Name of vector input map with Sentinel-2 footprints to limit import extent to not-null data.
```
neteler:
```suggestion
#% description: Name of vector input map with Sentinel-2 footprints to limit import extent to non NULL data
```

grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
```diff
@@ -491,7 +499,32 @@ def create_register_file(self, filename):
...
+        fp_rast_name = "fp_rast_%s"%(str(os.getpid()))
+        for map in imported_map_list:
+            gs.run_command('g.region',raster=map)
+            gs.run_command('v.to.rast',input=fp,cats=fp_dict['%s_%s'%(map.split('_')[0],map.split('_')[1])],use="val",value=1,memory=options['memory'], output=fp_rast_name)
+            gs.run_command('r.mapcalc',expression='tmp_%s = if(isnull(%s),null(),%s)'%(map,fp_rast_name,map))
+            gs.run_command('g.rename',raster='tmp_%s,%s'%(map,map))
+            gs.run_command('g.remove',type="raster",name=fp_rast_name,flags='f')
+        gs.run_command('g.region',region=reg_name)
+        gs.run_command('g.remove',type="region",name=reg_name,flags='f')
+
```
neteler:
```suggestion

    def set_footprint(self, fp):
        reg_name = "tmpregion_%s" % (str(os.getpid()))
        gs.run_command('g.region', save=reg_name)
        fp_dict = {}
        idlist = [x for x in gs.parse_command('v.db.select', map=fp)]
        cat_idx = idlist[0].split('|').index('cat')
        ident_idx = idlist[0].split('|').index('identifier')

        for item in idlist[1:]:
            tmp_key = item.split('|')[ident_idx]
            key = "%s_%s" % (tmp_key.split('_')[5], tmp_key.split('_')[2])
            fp_dict[key] = item.split('|')[cat_idx]
        fp_rast_name = "fp_rast_%s" % (str(os.getpid()))
        for map in imported_map_list:
            gs.run_command('g.region', raster=map)
            gs.run_command('v.to.rast', input=fp, cats=fp_dict['%s_%s' % (map.split('_')[0], map.split('_')[1])], use="val",
                           value=1, memory=options['memory'], output=fp_rast_name)
            gs.run_command('r.mapcalc', expression='tmp_%s = if(isnull(%s),null(),%s)' % (map, fp_rast_name, map))
            gs.run_command('g.rename', raster='tmp_%s,%s' % (map, map))
            gs.run_command('g.remove', type="raster", name=fp_rast_name, flags='f')
        gs.run_command('g.region', region=reg_name)
        gs.run_command('g.remove', type="region", name=reg_name, flags='f')

```

Reformatted to PEP8 (using PyCharm)

grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
```diff
@@ -504,8 +537,12 @@ def main():
                          "when -{} flag given").format('p'))
         importer.print_products()
         return 0
-    
+
     importer.import_products(flags['r'], flags['l'], flags['o'])
+
+    if options['footprints']:
+        importer.set_footprint(options['footprints'])
```
ninsbl: Could you include set_footprint() into import_products()? If the footprint is used to limit the computational region, that can speed-up projection (r.import) quite a bit and maybe also import without projection (r.in.gdal), especially for small footprints. So, import procedure would be:
 1. Set computational region to footprint (if provided and not options["extent"] == "region", in the latter case the overlapping area could be used as temp region)
 2. import
 3. Clip to footprint if provided
 4. round if import is done with r.import (r.proj) (See: #127)

grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
```diff
@@ -504,8 +537,12 @@ def main():
                          "when -{} flag given").format('p'))
         importer.print_products()
         return 0
-    
+
     importer.import_products(flags['r'], flags['l'], flags['o'])
+
+    if options['footprints']:
```
ninsbl: Could you include set_footprint() into import_products(), setting and limiting the temporary region to the footprint could speedup the import, esp. when done with projection (r.import). So the procedure would be:
1. set temp region to footprint (if provided and options["extent"] != "region")
2. import
3. Clip to footprint if provided or round if import was done this r.import (r.proj) (see: #127)

grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
```diff
@@ -491,7 +499,32 @@ def create_register_file(self, filename):
                         br='S2_{}'.format(band_ref)
                     ))
                 fd.write(os.linesep)
+
+    def set_footprint(self,fp):
```
ninsbl: If you use: `gs.use_temp_region()` (see: https://grass.osgeo.org/grass78/manuals/libpython/script.html?highlight=use_temp#script.core.use_temp_region) it is not necessary to capture the current region and it is also safer as an atexit handler is Integrated into "use_temp_region".

grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
```diff
@@ -491,7 +499,32 @@ def create_register_file(self, filename):
...
+        cat_idx = idlist[0].split('|').index('cat')
+        ident_idx = idlist[0].split('|').index('identifier')
+
+        for item in idlist[1:]:
+            tmp_key = item.split('|')[ident_idx]
+            key = "%s_%s"%(tmp_key.split('_')[5],tmp_key.split('_')[2])
+            fp_dict[key] = item.split('|')[cat_idx]
+        fp_rast_name = "fp_rast_%s"%(str(os.getpid()))
+        for map in imported_map_list:
+            gs.run_command('g.region',raster=map)
```
ninsbl: Here you could do:
```
reg = gs.parse_command("v.db.select", flags="r", map=fp, where="cats in (fp_dict['%s_%s'%(map.split('_')[0],map.split('_')[1])])")
gs.use_temp_region()
gs.run_command('g.region', align=map, n=reg["n"], s=reg["s"], e=reg["e"], w=reg["w"])
```
That zooms the computational region to the footprint, but Pixel alignment follows the input map

grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
```diff
@@ -491,7 +499,32 @@ def create_register_file(self, filename):
...
+            gs.run_command('v.to.rast',input=fp,cats=fp_dict['%s_%s'%(map.split('_')[0],map.split('_')[1])],use="val",value=1,memory=options['memory'], output=fp_rast_name)
+            gs.run_command('r.mapcalc',expression='tmp_%s = if(isnull(%s),null(),%s)'%(map,fp_rast_name,map))
+            gs.run_command('g.rename',raster='tmp_%s,%s'%(map,map))
```
ninsbl: Does renaming work here? I would expect it to fail cause map exsists...

grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
```diff
@@ -491,7 +499,32 @@ def create_register_file(self, filename):
...
+            gs.run_command('g.remove',type="raster",name=fp_rast_name,flags='f')
+        gs.run_command('g.region',region=reg_name)
```
ninsbl: If you use gs.use_temp_region(), here you would do:
```
gs.use_temp_region()
```
instead...

grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
```diff
@@ -491,7 +499,32 @@ def create_register_file(self, filename):
                         br='S2_{}'.format(band_ref)
                     ))
                 fd.write(os.linesep)
+
+    def set_footprint(self,fp):
+        reg_name = "tmpregion_%s"%(str(os.getpid()))
+        gs.run_command('g.region',save=reg_name)
+        fp_dict={}
+        idlist = [x for x in gs.parse_command('v.db.select',map=fp)]
```
ninsbl: If the map identifyer could be used in a where clause, this would not be necessary, cause `where="identifyer = {}".format(map_identifyer)` could be added to the `v.db.select` call for getting the footprint region as well as to `v.to.rast` 
```

File: grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py (hunk 7 above)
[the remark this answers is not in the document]
Comment (AnikaBettge) [reply grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py]: It should work: tmp_map ist renamed to map

File: grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py (hunk 4 above)
Comment (griembauer) [reply grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py]: to 1.) We might want to allow both extent=region AND footprints, e.g. if we are just interested in a small area (that might be covering an edge of a footprint)

File: grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py (hunk 6 above)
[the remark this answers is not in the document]
Comment (LeGuide) [reply grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py]: Thanks for the hint, it is updated now.

File: grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py [hunk 10]
```diff
@@ -491,7 +499,32 @@ def create_register_file(self, filename):
...
+
+        for item in idlist[1:]:
+            tmp_key = item.split('|')[ident_idx]
+            key = "%s_%s"%(tmp_key.split('_')[5],tmp_key.split('_')[2])
+            fp_dict[key] = item.split('|')[cat_idx]
+        fp_rast_name = "fp_rast_%s"%(str(os.getpid()))
+        for map in imported_map_list:
+            gs.run_command('g.region',raster=map)
+            gs.run_command('v.to.rast',input=fp,cats=fp_dict['%s_%s'%(map.split('_')[0],map.split('_')[1])],use="val",value=1,memory=options['memory'], output=fp_rast_name)
+            gs.run_command('r.mapcalc',expression='tmp_%s = if(isnull(%s),null(),%s)'%(map,fp_rast_name,map))
```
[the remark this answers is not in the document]
LeGuide: This is included now

grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
```diff
@@ -491,7 +499,32 @@ def create_register_file(self, filename):
                         br='S2_{}'.format(band_ref)
                     ))
                 fd.write(os.linesep)
+
+    def set_footprint(self,fp):
+        reg_name = "tmpregion_%s"%(str(os.getpid()))
```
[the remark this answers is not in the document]
LeGuide: now included (see above)

grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
LeGuide: now included (see above)

grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
LeGuide: now included (see above)

grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
LeGuide: Thanks, this now included as well

grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
LeGuide: set_footprint() is now moved into import_products()

metzm:
> @ninsbl Unfortunately, there are also areas within S-2 data that can be 0, e.g. water areas (in general, dark features). Using r.null therefore may leave gaps in the data.

According to the metadata, nodata are encoded as 0. Can you provide an example (without reprojection), were supposedly valid pixels are 0? That would be a bug in the original S2 data.

griembauer:
> > @ninsbl Unfortunately, there are also areas within S-2 data that can be 0, e.g. water areas (in general, dark features). Using r.null therefore may leave gaps in the data.
> 
> According to the metadata, nodata are encoded as 0. Can you provide an example (without reprojection), were supposedly valid pixels are 0? That would be a bug in the original S2 data.

An example is the scene S2B_MSIL2A_20190627T104029_N0212_R008_T32ULC_20190627T135004, uuid: bb9423ca-e076-41aa-830d-6afa8661df89. I used i.sentinel.download and i.sentinel.import to import it into GRASS, using a 32632 location (I checked that it is imported with r.in.gdal, hence no reprojection). An example is the Band B08_10m. Around coordinates 323167/5735880 there are some lakes that have pixels with value 0 (I reclassified 0 to -10000 to make it visible). These pixels are not indicated in the quality masks for the Band08 in the GRANULE/QI_DATA folder. These pixels are 0 only for B08, but other pixels within the lakes have the value 0 at other bands. 
S2_B08_Value0

neteler:
@griembauer there is the new i.zero2null (#206) available, developed by @metzm - would that help here?

griembauer:
Yes, this is a much cleaner solution and it makes this PR obsolete. Will close PR.

ninsbl:
Sorry, @griembauer . I hope my comments did not discourage you from merging this valuable improvement. It can speed-up esp. import with reprojection for scenes where only small parts contain data...
Would be nice to get this implemented...

griembauer:
> Sorry, @griembauer . I hope my comments did not discourage you from merging this valuable improvement. It can speed-up esp. import with reprojection for scenes where only small parts contain data...
> Would be nice to get this implemented...

No worries, in fact we realized that the vector footprints are not precise enough and so they should not be used to delimit the import region (footprint in grey): 
footprint_offset2

ninsbl:
Ah, OK, good to know. I did not notice.
So, g.region with zoom option might be more appropriate to limit the extent...

PR #126, opened 2020-03-20
New version of r.learn.ml

stevenpawley:
#### Important changes

- r.learn.ml is now split into two separate modules for training (r.learn.train) and prediction (r.learn.predict).
- New models added and various parameter name changes.
- Requires scikit-learn >= 0.20 and pandas >= 1.0.
- Feature importances now use method built into scikit-learn (requires scikit-learn >0.22).
- Categories automatically applied to classification maps from categories (if present) in training_map.
- Training points can be performed on numeric fields and text fields (labels are encoded automatically in alphabetical order).


AnikaBettge:
@stevenpawley are in r.learn.predict.py the parameter n_jobs and the flag -f taken used by the model or is there no way to specify them?

mlennert:
While it is great to use the latest and greatest, please note that pandas 1.0 came out less than 2 months ago which means that people in environments where updates do not happen often and easily will not be able to use of r.learn.ml for a while.

neteler:
Thanks for your efforts!

We extensively use the "old" version and rely on its parameter interface (when things are stable we will migrate to the new version). Would it be possible to keep old and new version in parallel for a while?

stevenpawley:
> @stevenpawley are in r.learn.predict.py the parameter n_jobs and the flag -f taken used by the model or is there no way to specify them?

The n_jobs parameter is passed to the permutation_importance method. n_jobs is not used separately in the prediction, but most of the classifiers/regressors already make use of parallel processes assuming they were trained with n_jobs=-1 or -2. For the few algorithms that are single threaded such as LinearDiscriminantAnalysis, QuadraticDiscriminantAnalysis, GaussianNB and single decision trees, and LinearRegression with a single target, and the user really needed parallel processing (they are generally really fast anyway), then I figured that it would be cleaner just to wrap r.learn.predict into the pygrass.modules.grid.grid module.

stevenpawley:
> While it is great to use the latest and greatest, please note that pandas 1.0 came out less than 2 months ago which means that people in environments where updates do not happen often and easily will not be able to use of r.learn.ml for a while.

Thanks for this comment. I could make a modification and support the use of an older version of pandas. I started using pandas >1 because it finally has a StringDType and I wanted to maintain the a specific string type for an application that I was working on. However, this doesn't really affect r.learn.ml if you are using through the GUI.

stevenpawley:
I also like Markus's idea of running two "new" and "old" versions in parallel for a while given that there are so many changes. Perhaps I should rename the "new" version as "r.learn.ml2" and will submit as a new add-on?

PR #128, opened 2020-03-23
os.path_exists -> os.path.exists in r.learn.ml

pesekon2:
Fix wrong syntax of `os.path.exists` in `r.learn.ml`


pesekon2:
Not needed if #126 will completely change `r.learn.ml`. If we go for Moritz's idea of two `r.learn.ml`s, it is still relevant.

neteler:
I merge in order not not have broken code in the repo if not needed :-)

PR #129, opened 2020-03-23
Created r.learn.ml2

stevenpawley:
New version of "r.learn.ml" created as addon "r.learn.ml2"


neteler:
Thanks so much, @stevenpawley - now testing will be much easier while keeping our existing workflow up & running!

PR #137, opened 2020-04-02
i.sentinel: easier computing shadows

AnikaBettge:
i.sentinel.mask needs the MTD_TL.xml file. The used information are also written by i.sentinel.import with r.support in the metadata of the bands.
Since, the metadata cannot be parsed properly due to wrong line breaks, you can now create a json dump metadata file by i.sentinel.import and use it in i.sentinel.mask.


grass7/imagery/i.sentinel/i.sentinel.mask/i.sentinel.mask.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
neteler, line 1:
```suggestion
#!/usr/bin/env python
```
g.extension will take care of this

grass7/imagery/i.sentinel/i.sentinel.mask/i.sentinel.mask.py
```diff
@@ -285,10 +296,14 @@ def main ():
 
     # Check input and output for shadow mask
     if not flags["c"]:
-        if mtd_file == '':
-            gscript.fatal('Metadata file is required for shadow mask computation. Please specified it')
-        if not os.path.exists(mtd_file):
-            gscript.fatal('Metadata file <{}> not found. Please select the right .xml file'.format(mtd_file))
+        if mtd_file == '' and metadata_file == '':
+            gscript.fatal('Metadata (file) is required for shadow mask computation. Please specified it')
```
neteler, line 300:
```suggestion
            gscript.fatal('Metadata (file) is required for shadow mask computation. Please specifiy it')
```

grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
```diff
@@ -66,6 +66,11 @@
 #% description: Name for output file to use with t.register
 #% required: no
 #%end
+#%option G_OPT_M_DIR
+#% key: metadata
+#% description: Name of directory into which Sentinel metadata json dumps are saved
```
neteler, line 71: How about predefining a directory? I'd like that, it could go into LOC/MAP/json/ or so and being used by  i.sentinel.mask.py as well (hence, less "thinking" for the user where to store this file which should really be next to the data)

grass7/imagery/i.sentinel/i.sentinel.mask/i.sentinel.mask.py
```diff
@@ -118,6 +118,12 @@
 #% required : no
 #% guisection: Metadata
 #%end
+#%option G_OPT_F_INPUT
+#% key: metadata
+#% description: name of Sentinel metadata json dump
```
neteler, line 123:
```suggestion
#% description: Name of Sentinel metadata json dump
```

see also comment above for predefined path

grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
AnikaBettge: @neteler should the file then always be written, or should another flag be added?

grass7/imagery/i.sentinel/i.sentinel.mask/i.sentinel.mask.py
```diff
@@ -222,6 +231,8 @@ def main ():
                     gscript.fatal(error_msg)
                 elif a[0] == 'MTD_TL.xml' and not mtd_file:
                     mtd_file = a[1].strip()
+                elif a[0] == 'metadata' and not metadata_file:
+                    metadata_file = a[1].strip()
```
RobiFag, line 235: Sorry but it's not so clear to me what this lines do when executed. This part was modified by @ninsbl and I'm not sure to fully understand.

ninsbl: Have not checked deeply, but my understanding is that when a file with input information is given, this line extracts the path to the metadata file.

Should probably be added to the documentation...

ninsbl:
FYI: I do have a (not yet committed) raster based computation of cloud shadows (derived from this module) that can speedup the process significantly, esp. for higher cloud coverage with many cloud areas...
Will try to get a test version out ASAP.

grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
AnikaBettge: I added a flag -f for this

grass7/imagery/i.sentinel/i.sentinel.mask/i.sentinel.mask.py
AnikaBettge: I added one sentence in the html

RobiFag: Thanks, now it's clearer to me!

grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
metzm: In order to maintain accordance with the current GRASS database structure, such a file should go to LOCATION_NAME/MAPSET/cell_misc/<raster map name>/ for individual rasters or to LOCATION_NAME/MAPSET/group/<group name>/ for imagery groups.

neteler:
@AnikaBettge please let me know when you think it is ready for merge,

AnikaBettge:
@neteler I think it is now ready for merge

grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
neteler, line 1:
```suggestion
#!/usr/bin/env python
```

grass7/imagery/i.sentinel/i.sentinel.mask/i.sentinel.mask.py
```diff
@@ -233,6 +244,8 @@ def main ():
                     bands[a[0]] = a[1].strip()
             if len(txt_bands) < 7:
                 gscript.fatal(('One or more bands are missing in the input text file.\n Only these bands have been found: {}').format(txt_bands))
+            if mtd_file and metadata_file:
+                gscript.fatal(('metadata file and mtd_file are given in the input text file.\n Only one of these should be set.'))
```
neteler, line 248:
```suggestion
                gscript.fatal(('Metadata json file and mtd_file are both given as input text files.\n Only one of these should be specified.'))
```

veroandreo:
This is a great improvement and I know this is merged already, but how are we supposed to easily know the name of the json dump to use it as input for `i.sentinel.mask`? Do we need to browse in the mapset files or does it get printed when `i.sentinel.import` finishes? An extra example for this use case would be highly appreciated and would also make the improvement more visible and discoverable by users

PR #138, opened 2020-04-04
WIP: Add i.cloud.shadow

ninsbl:
This is work in progress, not yet fully tested and not yet ready for review (though comments are welcome any time of course).
.
It is a slightly modified raster implementation of the cloud shadow detection in i.sentinel.mask. This module could be used with cloud and shadow estimates from other imagery as well.

It should speedup the process especially for many and complex clouds as data is never converted to vector, and shift is applied with r.region which takes less than a second (compared to v.transform).

See #137 

@RobiFag I listed you as a co-author as this is more or less a modification of your code. If you are uncomfortable with that let me know and I will fix that.


mlennert:
Would it make sense to merge this with i.sentinel.mask to make it just one module instead of two ?

ninsbl:
> Would it make sense to merge this with i.sentinel.mask to make it just one module instead of two ?

That is definately an option I am willing to implement.
The question is, are there other sensors where that kind of cloud shadow detection could be useful.
For Landsat FMASK and i.landsat.acca cloud shadow detection is included. So, it is in a way a question if we prefer more complex, sensor specific modules over more generic ones... No, strong opinion from my side. I am happy to implement it in the prefered way.

That said, this module could be integrated in i.sentinel.mask as well (drawback is that it adds a dependency)...

Other opinions, @RobiFag, @veroandreo @neteler, @AnikaBettge?

RobiFag:
If I understand correctly, this module extracts cloud shadows starting from an existing cloud map and an existing dark pixels map. The shadow map is obtained using an improved procedure similar to the "cleaning" procedure of i.sentinel.mask (intersection between cloud and shadow maps as a function of sun zenith, azimuth and cloud height ). 
In this way, the cloud and dark pixels maps can be computed using algorithms/software other than i.sentinel.mask that is strictly related to Sentinels bands, making the module more general and also suitable for other sensed images.
I think it could be a module in itself but maybe it could be also useful to modify the "cleaning" procedure of i.sentinel.mask introducing your improvements even if the addition of dependencies is not the best for me.

RobiFag:
> @RobiFag I listed you as a co-author as this is more or less a modification of your code. If you are uncomfortable with that let me know and I will fix that.

@ninsbl thank you so much! Actually, I'm glad to be mentioned as a co-author of your work!!

veroandreo:
AFAIU, this PR would mean a speed improvement for the identification of clouds' shadows because it's raster based instead of vector based. In that sense, I tend to think it should be integrated in `i.sentinel.mask`. Moreover, I'd like to see a benchmark of both approaches :)

Since this PR is only for enhanced cloud shadows detection, I don't really see it as a separate module. Instead, this could also be integrated into `i.landsat.acca` though it would be great if the later was updated to Landsat 8.

If we talk about a more general cloud + cloud shadow detection module, I would then go for a wraper for fmask, for example, that could take different types of images (i.e., L5, L7, L8 and S2) as input and returned cloud and cloud shadows masks.

ninsbl:
OK, then I will integrate the raster based function in i.sentinel.mask and discard this module draft.

A benchmark is a good point. for very few and small clouds, the vector based process may be faster. However, with more and complex clouds (many vertices) it can become pretty slow where a raster based process can get a significant edge in terms of performance...  
Maybe worth keeping the vector based approach if there are few clouds...

ninsbl:
Replaced by: #1002

Issue #140, opened 2020-04-05
i.sentinel.mask should remove all the *_double files


veroandreo:
`i.sentinel.mask` creates a floating point version of S2 bands for its calculations. 

IMO, it should remove them at exit maybe with a message or so. Think of users processing many S2 scenes with this add-on, then their mapsets (and disks) will be full of undesired intermediate maps.

@RobiFag what do you think?

RobiFag:
Maybe the *_double file can be defined as temp files that are removed at the end of the process by default but they can also be saved using the -t flag.
I put it on my TO DO list..I will make the changes in the next days!

veroandreo:
Thanks, Robi! :)

RobiFag:
@neteler I made changes requested by Vero. Maybe it's a trivial question but, can I push them directly or is it better to open a pull request?

veroandreo:
Ciao @RobiFag, if trivial you could push, but recommended is to commit through PR from a fix/feature branch

PR #145, opened 2020-04-15
Update i.sentinel.mask.py in order to remove rescaled files (*_double) 

RobiFag:
I have updated the i.sentinel.mask.py file in order to remove at the end of the process all the *_double file created with the -s flag as requested by @veroandreo in #140


veroandreo:
thanks Roby!

PR #163, opened 2020-05-05
i.sentinel.import: fix invalid format

landam:
`i.sentinel.import` fails with

```
Writing metadata to maps...
  File "C:\Users\martin\AppData\Roaming\GRASS7\addons/scripts/i.sentinel.import.py", line 561, in <module>
    sys.exit(main())
  File "C:\Users\martin\AppData\Roaming\GRASS7\addons/scripts/i.sentinel.import.py", line 545, in main
    importer.write_metadata()
  File "C:\Users\martin\AppData\Roaming\GRASS7\addons/scripts/i.sentinel.import.py", line 454, in write_metadata
    timestamp_str = timestamp.strftime("%-d %b %Y %H:%M:%S.%f")
ValueError: Invalid format string
```

### DIFF
--- a/grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
+++ b/grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
@@ -451,7 +451,10 @@ class SentinelImporter(object):
                     continue
 
                 timestamp = meta['timestamp']
-                timestamp_str = timestamp.strftime("%-d %b %Y %H:%M:%S.%f")
+                try:
+                    timestamp_str = timestamp.strftime("%d %b %Y %H:%M:%S.%f")
+                except ValueError as e:
+                    gs.fatal("{}: {}".format(timestamp, e))
                 descr_list = []
                 for dkey in meta.keys():
                     if dkey != 'timestamp':


wenzeslaus (changes requested):
What the try-except block is trying to do? No need to transform programmer's error causing a *desired* traceback into fatal error which misses any origin info. If the point is to catch issues with whatever is in the `meta` variable and report them to the user, then more elaborate message is needed to tell the user what is wrong. However, since `timestamp` is already a valid object (without checking the documentation) it seems to me that `strftime` should always succeed here. If it is not clear if it is a valid object, then perhaps place where `meta['timestamp']` is created is a better place to do that check. If the concern is the format string, that's a programmer's error because that is a literal in the program user does not know anything about.

ninsbl:
The timestamp should be checked already here (that it is not empty):
https://github.com/OSGeo/grass-addons/blob/41ba0cc25685b37f96ac93d3790df8644a159c82/grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py#L423

Maybe better/safer to use dateutil.parse here:
https://github.com/OSGeo/grass-addons/blob/8a48acd82e1c069786d623317d93286989798b02/grass7/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py#L380
instead of hardcoded format parsing?

landam:
Closed by #705

PR #165, opened 2020-05-09
r.accumulate: Update the manual

HuidaeCho:
I just noticed that using the same NC sample data set, the `r.watershed` module now produces slightly different results from when I first created the manual. I got different drainage directions causing the one outlet that I replaced only generating a very short longest flow path. Either r.watershed or the NC data set has changed?


metzm:
Looks good to me. Please provide a title and description with the motivation for this change.

HuidaeCho:
> Looks good to me. Please provide a title and description with the motivation for this change.

Sure. See above. Thanks for reviewing.

PR #180, opened 2020-05-25
Update r.width.funct.py

marcocurreli:
fix errors:
 unsupported operand type(s) for %: 'NoneType' and 'tuple' (rows 103...111)
 wx backend was deprecated in Matplotlib 2.0 and will be removed in the future. Use wxagg instead.

### DIFF
--- a/grass7/raster/r.width.funct/r.width.funct.py
+++ b/grass7/raster/r.width.funct/r.width.funct.py
@@ -43,7 +43,7 @@
 import sys
 import os
 import matplotlib #required by windows
-matplotlib.use('wx') #required by windows
+matplotlib.use('wxagg') #required by windows
 import matplotlib.pyplot as plt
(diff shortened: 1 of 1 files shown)


neteler:
fixes #179

grass7/raster/r.width.funct/r.width.funct.py
```diff
@@ -43,7 +43,7 @@
 import sys
 import os
 import matplotlib #required by windows
-matplotlib.use('wx') #required by windows
+matplotlib.use('wxagg') #required by windows
```
neteler, line 46:
```suggestion
matplotlib.use('WXAgg') #required by windows
```
(as e.g. here: https://github.com/OSGeo/grass/blob/master/gui/wxpython/iscatt/plots.py#L33)

PR #190, opened 2020-06-01
r.accumulate: Add an iterative algorithm (default; memory safe) for longest flow paths

HuidaeCho:


HuidaeCho:
This module will deprecate `r.lfp` written in Python.

PR #193, opened 2020-06-01
r.lfp deprecated by r.accumulate; removed

HuidaeCho:


neteler:
Hi, just a remark:

Usually we don't remove a module entirely but put a file `DEPRECATED` into the directory.
See for example
https://github.com/OSGeo/grass-addons/blob/e1f26fae7d2531d9a6f6a562ea799d83dd5682ee/grass7/imagery/i.sentinel/Makefile#L5
for the respective handling in the `Makefile`.

Of course the decision is with the author!

HuidaeCho:
> Hi, just a remark:
> 
> Usually we don't remove a module entirely but put a file `DEPRECATED` into the directory.
> See for example
> 
> https://github.com/OSGeo/grass-addons/blob/e1f26fae7d2531d9a6f6a562ea799d83dd5682ee/grass7/imagery/i.sentinel/Makefile#L5
> 
> for the respective handling in the `Makefile`.
> 

I thought about that, but, in this case, `r.lfp` is a much inefficient duplicate of `r.accumulate` in terms of functionality, so I decided to remove it because there is an alternative. I can add it back if that's a better practice. Is it a simple git trick to revert it or should I add a new folder just like a new module?

HuidaeCho:
> > Hi, just a remark:
> > Usually we don't remove a module entirely but put a file `DEPRECATED` into the directory.
> > See for example
> > https://github.com/OSGeo/grass-addons/blob/e1f26fae7d2531d9a6f6a562ea799d83dd5682ee/grass7/imagery/i.sentinel/Makefile#L5
> > 
> > for the respective handling in the `Makefile`.
> 
> I thought about that, but, in this case, `r.lfp` is a much inefficient duplicate of `r.accumulate` in terms of functionality, so I decided to remove it because there is an alternative. I can add it back if that's a better practice. Is it a simple git trick to revert it or should I add a new folder just like a new module?

Ah! Found the Revert button up there.

PR #196, opened 2020-06-01
Deprecate r.lfp instead of removing it

HuidaeCho:
Reverts OSGeo/grass-addons#193


HuidaeCho:
@neteler It's still installed with DEPRECATED. Is it normal?

neteler:
> @neteler It's still installed with DEPRECATED. Is it normal? Looks like this file is only for informational purposes only?

AFAIK it is used on the WinGRASS addon build server. 
Probably g.extension should be made sensitive to it as well and issue a warning in case the file is present.

PR #198, opened 2020-06-06
i.modis.import creates temporary data in current directory

landam:
## To reproduce

```
i.modis.import input=~/geodata/modis/MCD19A2.A2019031.h18v03.006.2019036210121.hdf -t --o
ls *.tif -w1
MCD19A2.A2019031.h18v03.single_AOD_QA.tif
MCD19A2.A2019031.h18v03.single_Optical_Depth_055.tif
```
## Important on Windows

This is causing serious problem on Windows (here current working directory is set to C:\ by default).

i-modis-win-bug


landam:
See https://github.com/OSGeo/grass-addons/pull/199

PR #199, opened 2020-06-06
i.modis.import: change cwd to input data dir

landam:
## To reproduce

```
i.modis.import input=~/geodata/modis/MCD19A2.A2019031.h18v03.006.2019036210121.hdf -t --o
ls *.tif -w1
MCD19A2.A2019031.h18v03.single_AOD_QA.tif
MCD19A2.A2019031.h18v03.single_Optical_Depth_055.tif
```
## Important on Windows

This is causing serious problem on Windows (here current working directory is set to C:\ by default).

i-modis-win-bug

### DIFF
--- a/grass7/imagery/i.modis/i.modis.import/i.modis.import.py
+++ b/grass7/imagery/i.modis/i.modis.import/i.modis.import.py
@@ -382,6 +382,9 @@ def single(options, remove, an, ow, fil):
             outname = outname.replace(' ', '_')
             execmodis = convertModisGDAL(str(hdf), outname, spectr, res,
                                          wkt=str(projwkt))
+
+        # produce temporary files in input folder
+        os.chdir(basedir)
         try:
             execmodis.run(quiet=True)
         except:
@@ -462,6 +465,9 @@ def mosaic(options, remove, an, ow, fil):
                     res = None
                 execmodis = convertModisGDAL(str(hdf), out, spectr, res,
                                              wkt=str(projwkt), vrt=True)
+
+            # produce temporary files in input folder
+            os.chdir(basedir)
             try:
                 execmodis.run(quiet=True)
             except:


neteler:
Wouldn't it be better to use the standard `.tmp` directory of GRASS here, in the mapset? 
We sometimes have the input directory as read only network directory, would that still work with this PR?

landam:
> Wouldn't it be better to use the standard `.tmp` directory of GRASS here, in the mapset?
> We sometimes have the input directory as read only network directory, would that still work with this PR?

No, you have to have write permission to the directory where temporary data is stored (see original problem on Windows).

Yes, I was thinking about `.tmp`, but there is `-t` flag.

```
  -t   Preserve temporary files (TIF and HDF mosaic)
```

In the case of `-t` flag it doesn't make sense to store temporary data in `.tmp` directory. 

Option 1: store temporary data in input data dir (this PR)

Option 2: store temporary data in `.tmp` when `-t` is not given otherwise in input data dir

@neteler @lucadelu What do you think?

neteler:
Will there be a fallback in case the input data directory is read only as it sometimes occurs?

landam:
> Will there be a fallback in case the input data directory is read only as it sometimes occurs?

Do you mean when `-t` flag given? @neteler What about this? 

1) no -t flag -> use `.tmp`
2) -t flag -> use input data dir (if not writable fails with error)

OR

1) no -t flag -> use `.tmp`
2) use `.tmp`, after import copy temporary data to input data dir (if not writable fails with error, but data will imported anyway)

lucadelu:
> > Will there be a fallback in case the input data directory is read only as it sometimes occurs?
> 
> Do you mean when `-t` flag given? @neteler What about this?
> 
>     1. no -t flag -> use `.tmp`
> 
>     2. -t flag -> use input data dir (if not writable fails with error)
> 

I like more this option

landam:
> Wouldn't it be better to use the standard `.tmp` directory of GRASS here, in the mapset?
> We sometimes have the input directory as read only network directory, would that still work with this PR?

It needs more effort especially for mosaics (create VRT file in different directory where HDF files are located).

landam:
I took liberty to merge this PR in order to make this addon functional on Windows. @neteler Also related issues https://github.com/OSGeo/grass-addons/issues/203 open.

PR #204, opened 2020-06-12
Add CI with build and test

wenzeslaus:
Couple of things to keep in mind:

1. The build does not use the current build script because they mix 6 and 7 and are not parametrized enough.
2. The make step does not return non-zero in case of failure because module failure is not counted as overall failure. That does not go well with GitHub Actions or Travis style CI.
3. There is only a small number of tests, i.e., the test coverage is small. 
4. Most of the tests are failing, but many can be probably fixed by reviewing the dataset used.


wenzeslaus:
Closing and replacing by #205.

Created from master by mistake. GitHub Actions won't kick in this case, so the PR is not useful for seeing the changes (even the master branch on fork won't have them running). Changing the "from" branch in a created PR is not possible (even for a draft).

PR #205, opened 2020-06-12
Add CI with build and test

wenzeslaus:
Couple of things to keep in mind:

1. The build does not use the current build script because they mix 6 and 7 and are not parametrized enough.
2. The make step does not return non-zero in case of failure because module failure is not counted as overall failure. That does not go well with GitHub Actions or Travis style CI.
3. There is only a small number of tests, i.e., the test coverage is small. 
4. Most of the tests are failing, but many can be probably fixed by reviewing the dataset used.


wenzeslaus:
Although the checks are not visible here and are not running in this repo, they are running on the fork, so merging this assuming that the the checks will kick in for this repo too.

neteler:
Thanks for your efforts, @wenzeslaus!

PR #206, opened 2020-06-13
i.zero2null

metzm:
fix nodata encoding for Sentinel-2 imagery, and possibly others
the value zero does not make sense for reflectance being the proportion of incoming light reflected back
zero is only possible for absolute black, all incoming light is completely absorbed (black holes can do that)


wenzeslaus:
That awesome black holes note would be worth having in the manual!

PR #220, opened 2020-06-21
Major update build_xml: argparse, python3 support

landam:


neteler:
@landam @tmszi Am I right that #656 has replaced this PR?

tmszi:
> @landam @tmszi Am I right that #656 has replaced this PR?

Yes you are right. PR #656 was based on this PR.

PR #231, opened 2020-06-29
cronjob grasslxd: fix broken tar flag order

neteler:
- fixes lack of updates due to wrong position of --exclude-vcs in tar command
- var name sync


neteler:
PR addresses compilation part of https://github.com/OSGeo/grass-website/issues/177

PR #237, opened 2020-07-05
r.accumulate: Add -N flag for faster flow accumulation (with this flag is slower)

HuidaeCho:
By default without the -N flag, flow accumulation is faster than with this flag. Nullifying cells with no flow accumulation (-N flag) adds too much overhead (`Rast_set_[cfd]_null_value` + null file). Subsequent analyses are not affected.

### DIFF
--- a/grass7/raster/r.accumulate/accumulate.c
+++ b/grass7/raster/r.accumulate/accumulate.c
@@ -7,7 +7,8 @@ static double trace_up(struct cell_map *, struct raster_map *,
                        struct raster_map *, char **, char, int, int);
 
 void accumulate(struct cell_map *dir_buf, struct raster_map *weight_buf,
-                struct raster_map *accum_buf, char **done, char neg)
+                struct raster_map *accum_buf, char **done, char neg,
+                char null)
 {
     int row, col;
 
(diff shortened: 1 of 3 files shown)


neteler:
Hi, I darkly think to remember that minor and capital flags may collide on some operating systems.. didn't we eliminate that in the GRASS GIS 7.0 cleanup? Cannot check right now... I may be wrong, though.

HuidaeCho:
> Hi, I darkly think to remember that minor and capital flags may collide on some operating systems.. didn't we eliminate that in the GRASS GIS 7.0 cleanup? Cannot check right now... I may be wrong, though.

Are you sure? `r.stats` has `[acnACN]` flags, which work fine on Windows. I also confirmed `-N` works on Windows. I think it's the filename that is case-insensitive, not arguments. Any links to this discussion?

neteler:
> > Hi, I darkly think to remember that minor and capital flags may collide on some operating systems.. didn't we eliminate that in the GRASS GIS 7.0 cleanup? Cannot check right now... I may be wrong, though.
> 
> Are you sure?

No, not at all.

> `r.stats` has `[acnACN]` flags, which work fine on Windows. I also confirmed this `-N` flag works on Windows. I think it's the filename that is case-insensitive, not arguments. Any links to this discussion?

I think I remembered the parameter discussion from 2014:

https://lists.osgeo.org/pipermail/grass-dev/2014-September/070936.html
(and related thread)

So, it should be all fine, sorry for the noise.

PR #259, opened 2020-07-24
r.colors.out_sld: do not write NaN in SLD without -n flag

AnikaBettge:
Using the generated SLD file with currently released version leads to an ERROR in Geoserver:
```
Error rendering coverage on the fast path
java.lang.UnsupportedOperationException: NaN values can only be set inside a single-point Range
NaN values can only be set inside a single-point Range
```
due to the NULL value line in the sld `<sld:ColorMapEntry color="#ffffff" opacity="0" quantity="NaN"/>`.

To avoid this I added a flag `n` which propagates NULL only if `-n` is set.

### DIFF
--- a/grass7/raster/r.colors.out_sld/r.colors.out_sld.html
+++ b/grass7/raster/r.colors.out_sld/r.colors.out_sld.html
@@ -7,6 +7,9 @@ the Styled Layer Description (SLD) format according to OGC standard.
 requires that the input map is read and may thus take a bit longer than the
 export of contious color rules (ramp).
 
+<p> Only if the flag <b>n<\b> is given, the NaN values are written into the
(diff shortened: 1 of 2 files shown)


grass7/raster/r.colors.out_sld/r.colors.out_sld.html
```diff
@@ -7,6 +7,9 @@ <h2>DESCRIPTION:</h2>
 requires that the input map is read and may thus take a bit longer than the
 export of contious color rules (ramp).
 
+<p> Only if the flag <b>n<\b> is given, the NaN values are written into the
+sld file, otherwise they lead to errors e.g. by unsing the sld in geoserver.
```
neteler, line 11:
```suggestion
generated SLD file which leads e.g. in GeoServer to an error when using such SLD.
```

PR #262, opened 2020-07-29
Add relativeorbitnumber as parameter to i.sentinel.download

marlinmm:
This allows the user to download only files of matching relative orbit number, useful for backscatter comparison analysis and some other stuff.  Can be applied to Sentinel-1 and -2 data, but orbit number ranges differ (S-1: 1 to 175, S-2: 1 to 143)


marlinmm:
It didnt quiet work before, should work now is intented... But it was only tested on Sentinel-1 data, I don't know if this is an issue

neteler:
@AnikaBettge could you please take a look?

AnikaBettge:
A message when the orbit numbers are no longer in their respective ranges would be nice.

marlinmm:
> A message when the orbit numbers are no longer in their respective ranges would be nice.

I added a small check to tell the user, if the number given is above the allowed range and also added info about the allowed ranges to the description of the field.  Works for S1 and S2. 
 
I guess, the user has to know the correct numbers anyway, for the function to be useful.

grass7/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
```diff
@@ -101,6 +101,13 @@
 #% guisection: Filter
 #%end
 #%option
+#% key: relativeorbitnumber
+#% type: integer
+#% multiple: no
+#% description: Relative orbit number to download (Sen-1: from 1 to 175; Sen-2: from 1 to 143)
```
neteler, line 107:
```suggestion
#% description: Relative orbit number to download (Sentinel-1: from 1 to 175; Sentinel-2: from 1 to 143)
```

(just to avoid abbreviations for nowcomers)

neteler:
Thanks for you contribution!

PR #58, opened 2019-11-14
r.in.srtm.region: support also locations with not longlat projection

AnikaBettge:
Set region to support also locations with not longlat projection.
Depending on the change in r.in.srtm (https://github.com/OSGeo/grass/pull/194).

PR #266, opened 2020-08-20
r.in.nasadem: no longlat projection

AnikaBettge:
Until now r.in.nasadem was only working in locations with longlat projection. This changes create a temporary location with epsg:4326 and import the nasadem and then reproject it to the target loaction. (It work similar to https://github.com/OSGeo/grass-addons/pull/58)

### DIFF
--- a/grass7/raster/r.in.nasadem/r.in.nasadem.py
+++ b/grass7/raster/r.in.nasadem/r.in.nasadem.py
@@ -92,10 +92,20 @@
 #% description: Resolution of output raster map (required if location projection not longlat)
 #% guisection: Output
 #%end
+#%option
+#% key: memory
+#% type: integer
+#% required: no
+#% multiple: no
+#% label: Maximum memory to be used (in MB)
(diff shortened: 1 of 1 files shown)


neteler (approved):
Looks very good to me, thanks.

I have made the test
- create longlat location, run the original r.in.nasadem
- create UTM49S location
- reproject NASADEM into it
- run this PR
- compare differences with r.mapcalc:
```

r.univar diff -e
 100%
total null and non-null cells: 19455472
total null cells: 9729847

Of the non-null cells:
----------------------
n: 9725625
minimum: -7.62939e-06
maximum: 1.52588e-05
range: 2.28882e-05
mean: 7.06017e-12
mean of absolute values: 1.33359e-11
standard deviation: 0
variance: 0
variation coefficient: 0 %
sum: 6.866455078125e-05
1st quartile: 0
median (odd number of cells): 0
3rd quartile: 0
90th percentile: 0
```

IMHO perfect.

PR #287, opened 2020-09-27
r.valley.bottom improvements

stevenpawley:
I've provided some tweaks to r.valley.bottom in order to speed up the computation, and (somewhat) reduce the disk space that is consumed by temporary files during execution. The bottle-neck is the elevation_percentile calculation, which can be now executed using the grass.pygrass.grid.GridModule is the argument `n_jobs` is > 1. This speeds up the computation on my system (16 cores), especially on huge rasters.

Also, r.slope.aspect is used instead of a r.mapcalc statement to calculate slope, because r.slope.aspect now has the "e" flag to avoid edge-effects.

The other improvement relates to removing temporary raster maps sooner during execution, so the disk space that is used should be smaller.

Ideally, the elevation percentile function needs to be written in C. I might attempt this next as a separate GRASS addon, and then we could use that.


grass7/raster/r.valley.bottom/r.valley.bottom.py
```diff
@@ -46,196 +46,235 @@
...
-def get_percentile(L, input, radius=3, window_square=False):
-    """Calculates the percentile whichj is the ratio of the number of points of
-    lower elevation to the total number of points in the surrounding region
+def elevation_percentile(L, input, radius=3, window_square=False, n_jobs=1):
+    """Calculates the percentile which is the ratio of the number of points of lower
+    elevation to the total number of points in the surrounding region
+
+    Notes
+    -----
+    Note this function is currently a bottle-neck is the module because r.mapcalc
```
hellik, line 245: just a small typo: "Note this function is currently a bottle-neck **is** the module" -> "Note this function is currently a bottle-neck **of** the module"

grass7/raster/r.valley.bottom/r.valley.bottom.py
```diff
@@ -46,196 +46,235 @@
...
+        The GRASS raster map (elevation) to perform calculation on.
+
     radius : int
-        Neighborhood radius (in pixels)
-    window_square : bool. Optional (default is False)
-        Boolean to use square or circular neighborhood
-    
+        The neighborhood radius (in pixels).
+
+    window_square : bool (opt). Ddefault is False
```
hellik, line 260: small typo: Ddefault => Default

hellik (commented):
some small typos, otherwise the PR looks good. I'll test it in the next days.

PR #288, opened 2020-09-27
Improvements to r.valley.bottom

stevenpawley:
I've provided some tweaks to r.valley.bottom in order to speed up the computation, and (somewhat) reduce the disk space that is consumed by temporary files during execution. The bottle-neck is the elevation_percentile calculation, which can be now executed using the grass.pygrass.grid.GridModule is the argument n_jobs is > 1. This speeds up the computation on my system (16 cores), especially on huge rasters.

Also, r.slope.aspect is used instead of a r.mapcalc statement to calculate slope, because r.slope.aspect now has the "e" flag to avoid edge-effects.

The other improvement relates to removing temporary raster maps sooner during execution, so the disk space that is used should be smaller.

Ideally, the elevation percentile function needs to be written in C. I might attempt this next as a separate GRASS addon, and then we could use that.


neteler:
Hi @stevenpawley
I see that this PR contains 3 (?) new modules:

- Added r.neighbors.tiled
- Added r.percentile
- Addded r.tpi

Exciting!

Sorry for bothering may I suggest to separate them out into own PRs? That would be more atomic...

stevenpawley:
Ah sorry!

I totally forget that I had an existing PR. I was aiming to keep the other modules really for my own use, for now.

I’ll create a new PR for the r.valley.bottom changes.

Cheers,

Steve

PR #306, opened 2020-10-06
change inequality signs in v.what.strds.timestamp

### LINKED ISSUE #305: [Bug] v.what.strds.timestamp not including the first timestamp
**Name of the addon**

v.what.strds.timestamp

**Describe the bug**

The `where` condition used for the vector feature selection [here](https://github.com/OSGeo/grass-addons/blob/master/grass7/vector/v.what.strds.timestamp/v.what.strds.timestamp.py#L109-L110) uses `> start_time` (strict inequality). It can be a problem in the case of vector features stamped with a day timestamp and raster maps time-spanning only one day as no vector features would be selected.

Exempli gratia: A vector feature is timestamped with a value of `2020-05-03`. We have STRDS containing two rasters stamped as `2 May 2020 00:00:00 / 3 May 2020 00:00:00` and `3 May 2020 00:00:00 / 4 May 2020 00:00:00`. The vector feature would not get any value.

**Expected behavior**

As a non-strict inequality is used [here](https://github.com/OSGeo/grass-addons/blob/master/grass7/vector/v.what.strds.timestamp/v.what.strds.timestamp.py#L208) for the STRDS selection `tempwhere`, I would expect the same approach (allowing `>=`) also in the code line mentioned above.

 - Operating System: Debian GNU/Linux bullseye/sid
 - GRASS GIS version 7.9 dev (commit f0966f1969c73a3797c3a87343add28d6fcaa0ba)

pesekon2:
fix #305

### DIFF
--- a/grass7/vector/v.what.strds.timestamp/v.what.strds.timestamp.py
+++ b/grass7/vector/v.what.strds.timestamp/v.what.strds.timestamp.py
@@ -96,7 +96,7 @@ def sample_relative(input, layer, timestamp_column, column, t_raster,
     start = t_raster["start_time"]
     end = t_raster["end_time"]
     raster_map = '{}@{}'.format(t_raster["name"], t_raster["mapset"])
-    where += """(julianday({0}) > date('{1}') AND \
+    where += """(julianday({0}) >= date('{1}') AND \
                 julianday({0}) < date('{2}'))""".format(timestamp_column, start, end)
 
 def sample_absolute(input, layer, timestamp_column, column, t_raster,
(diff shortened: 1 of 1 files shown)


ninsbl (approved):

neteler:
(pls put the module/library name first as we use this for ChangeLogs)

pesekon2:
Oops. 

You mean to change the commit message? That could be done with `commit --amend`, but that is actually very discouraged as it rewrites the repo history -> everyone that has pulled meanwhile would have to change his local repo history manually.

neteler:
Nono, I just referred to the PR title. All fine...

pesekon2:
Ah okay. Good, thanks.

PR #315, opened 2020-10-14
v.db.pyupdate: Update vector attributes with Python

wenzeslaus:
Uses v.db.select JSON output, Python eval and exec, and db.execute to compute new attributes.

### DIFF
--- a/grass7/vector/v.db.pyupdate/Makefile
+++ b/grass7/vector/v.db.pyupdate/Makefile
@@ -1,0 +1,7 @@
+MODULE_TOPDIR = ../..
+
+PGM = v.db.pyupdate
+
+include $(MODULE_TOPDIR)/include/Make/Script.make
+
+default: script
--- a/grass7/vector/v.db.pyupdate/testsuite/test_v_db_pyupdate.py
+++ b/grass7/vector/v.db.pyupdate/testsuite/test_v_db_pyupdate.py
@@ -1,0 +1,75 @@
+#!/usr/bin/env python3
+
+"""Test of v.db.pyupdate module
+
+Author: Vaclav Petras
+"""
+
+import json
+
+from grass.gunittest.case import TestCase
+from grass.gunittest.main import test
+
+import grass.script as gs
+
+
+class TestVDbPyUpdate(TestCase):
+    """Test v.db.pyupdate module"""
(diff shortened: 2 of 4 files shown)


ninsbl:
Nice!

neteler:
Would this new module eventually replace `v.db.update` in core, in order to get rid of the not really supported [SQLite extended functions](https://grass.osgeo.org/grass-stable/manuals/v.db.update.html#updating-of-columns-with-on-the-fly-calculation-(sqlite-extended-functions))?

mlennert:
On 14/10/20 17:38, Markus Neteler wrote:
> Would this new module eventually replace |v.db.update| in core, in order 
> to get rid of the not really supported SQLite extended functions 

Please do not get rid of the current v.db.update. As Vaclav mentions in 
the man page the SQL approach is faster (and probably also less memory 
hungry) than the Python approach in v.db.pyupdate.

An option could be to merge the two by adding the relevant python 
related paramters to v.db.update...

I have no idea what proportion of users of v.db.update actually need the 
extended functions.

wenzeslaus:
Thanks for feedback, everybody!

@neteler, as @mlennert says, this is much slower, an order of magnitude slower in fact. Even if we would make this the primary update module - which I'm _not_ arguing for in any way now - SQL-based _v.db.update_ has its advantages like that you can use SQL and keep the processing in the database engine in terms of speed but also semantics and functionality including backend specific functionality.

Is this an alternative to _v.db.update_ with `sqliteextra`? Yes, absolutely! On a controlled system for math functions, `sqliteextra` is probably a better match because of the performance and no, or known, limitations. _v.db.pyupdate_ may work better for an average use case and common user, but it will be really beneficial for complex updates. However, it will probably be always tricky to deal with performance and non-trivial types.

@mlennert, merging these two is an interesting idea. I considered this as an alternative, but you are right that from user perspective there might not be such a big difference and it might be more understandable (on the other hand, we have also the trend of splitting and wrapping overly complicated modules). From code perspective, it is a very different idea, but at the end, it is just a different way of creating an UPDATE statement, only the first one is trivial and the second one is very costly.

I will try to follow with another PR with some additional documentation, maybe also timings, memory consumption, tests, and improvements.

PR #321, opened 2020-10-21
Modifications to i.image.bathymetry to return original functionality

dnewcomb:
This will restore the previous options for adaptive and bisquare kernel as described in the manual page.

### DIFF
--- a/grass7/imagery/i.image.bathymetry/i.image.bathymetry.py
+++ b/grass7/imagery/i.image.bathymetry/i.image.bathymetry.py
@@ -4,6 +4,7 @@
 #
 # MODULE: i.image.bathymetry
 # AUTHOR(S): Vinayaraj Poliyapram <vinay223333@gmail.com> and Luca Delulucchi
+#            Scripting fixes 10/19/2020, Doug Newcomb
 #
 # PURPOSE:   Script for estimating bathymetry from optical satellite images
 # COPYRIGHT: (C) Vinayaraj Poliyapram and by the GRASS Development Team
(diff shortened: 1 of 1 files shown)


grass7/imagery/i.image.bathymetry/i.image.bathymetry.py
```diff
@@ -290,7 +296,10 @@ def main():
                     '(Rapid_ref.sdf))\n'
             r_file.write(ref_file)
             l = []
-            predict = g.read_command("g.tempfile", pid=os.getpid()).strip() + '.txt'
+            #predict = g.read_command("g.tempfile", pid=os.getpid()).strip() + '.txt'
+            #replace dynamic variable of predict with a static filename until g.tempfile formatting
+            #is cross platform
+            predict="bathytemp.txt"
```
lucadelu, line 302: could we use tempfile library to get a temporary name? something like os.path.join(tempfile.gettempdir(),  os.getpid()).strip() + '.txt') should work

lucadelu (commented):
It seems good, I suggest only a small change

grass7/imagery/i.image.bathymetry/i.image.bathymetry.py
ninsbl: Or something like:
https://grass.osgeo.org/grass76/manuals/libpython/script.html#script.core.tempname
?

lucadelu: yes, maybe better

neteler: Better refer to at least "stable", i.e.

https://grass.osgeo.org/grass78/manuals/libpython/script.html#script.core.tempname

PR #322, opened 2020-10-26
i.sentinel.download: enable footprint saving for Sentinel-1 data

griembauer:
This PR fixes the following issue in `i.sentinel.download`: 
Using `i.sentinel.download` with `producttype=GRD` or `producttype=SLC` (Sentinel-1 product types) and `footprints=...` fails, because a cloud cover percentage parameter is expected in `save_footprints` which is not available for Sentinel-1 data.


lucadelu (approved):

neteler:
Thanks!

PR #338, opened 2020-11-17
g.proj.identify: lazy import of osgeo library

lucadelu:


neteler:
I suggest to merge it.

PR #344, opened 2020-11-18
r.landscape.evol: try to resolve string to byte problem

lucadelu:
This should solve the problem reported by @cmbarton in #156


isaacullah:
I have fixed this issue with #873. Issue is now closed.

PR #361, opened 2020-12-10
m.csv.clean: Cleaning of CSV files

wenzeslaus:
Uses Python csv package to read CSV file, runs series of cleanups, esp. for
column names in the first row and also removes extra whitespace from cells.

It can prefix numbers used as column names (such as ZIP code or DOY)
and it can identify dates and reformat them as needed using Python datetime package.

The interface is designed so that user can process large number of column without
specifying them which means that there is some guess work inside to make that happen.
This applies esp. to columns containing date and columns with missing
column name in header.


neteler:
(nice original names in the NOTES!)

wenzeslaus:
Now, I'm wondering if the name should be _m.csv.clean_, not _m.csv.cleanup_.

> (nice original names in the NOTES!) 1st_place_medal +1

There was a request for more easter eggs.

PR #362, opened 2020-12-10
v.in.csv: Easy import of CSV as vector points

wenzeslaus:
Imports CSV as vector points with reprojection using Python pyproj package.

Fixes column names in the header, can limit number of imported lines,
allows to set column types as integers or floats by specifying
column names.

### DIFF
--- a/grass7/vector/v.in.csv/Makefile
+++ b/grass7/vector/v.in.csv/Makefile
@@ -1,0 +1,7 @@
+MODULE_TOPDIR = ../..
+
+PGM = v.in.csv
+
+include $(MODULE_TOPDIR)/include/Make/Script.make
+
+default: script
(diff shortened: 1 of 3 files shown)


mlennert:
Great ! Only thing: is it really necessary to add pyproj as a dependency ? I know it makes life easier for development, but requires the user to install an external where one would think that we can handle all this in GRASS GIS internally ?

wenzeslaus:
I'm leaving any non-pyproj for the future. At least the header needs to be sanitized in this module first, so at this point at least partial reading of the CSV is required anyway here, hence making pyproj the simpler solution (as @mlennert mentioned). Maybe when there is an API for the location-reprojection excercise in GRASS like Session or something...

PR #369, opened 2020-12-19
v.in.pygbif: lazy import of 'osgeo' module

tmszi:
**Error message:**

```
  File "C:/Users/landa/grass_packager/grass784/x86_64/addons/v.in.pygbif/scripts/v.in.pygbif.py", line 185, in <module>
    from osgeo import ogr, osr
ModuleNotFoundError: No module named 'osgeo'
```

Log file:
https://wingrass.fsv.cvut.cz/grass78/x86_64/addons/latest/logs/v.in.pygbif.log

### DIFF
--- a/grass7/vector/v.in.pygbif/v.in.pygbif.py
+++ b/grass7/vector/v.in.pygbif/v.in.pygbif.py
@@ -182,8 +182,6 @@ import sys
 
 # import os
 import math
-from osgeo import ogr, osr
-from osgeo import __version__ as gdal_version
 import grass.script as grass
 from grass.pygrass.vector import Vector
 from grass.pygrass.vector import VectorTopo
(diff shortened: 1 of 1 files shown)


grass7/vector/v.in.pygbif/v.in.pygbif.py
```diff
@@ -832,5 +830,12 @@ def main():
 
 # Run the module
 if __name__ == "__main__":
+    try:
+        from osgeo import ogr, osr
+        from osgeo import __version__ as gdal_version
+    except ImportError:
+        grass.fatal(_("Unable to load GDAL Python bindings (requires "
+                      "package 'python-gdal' being installed)"))
```
ninsbl, line 838:
```suggestion
                      "package 'python-gdal' or python library GDAL "
                      "to be installed)."))
```

ninsbl (approved):
Thanks for catching this. A minor suggestion inline. Feel free to disregard if you prefer the original change.

grass7/vector/v.in.pygbif/v.in.pygbif.py
tmszi: Fixed in the [56c0275](https://github.com/OSGeo/grass-addons/commit/56c02750f30616cc3d5d94a37110e2f37491ec3d).

PR #374, opened 2020-12-26
i.ann.maskrcnn: update installation command in docs

pesekon2:
The installation command in docs expected the installation from local path. As the modules are part of the official AddOns repository, the path specification in the installation command is not needed anymore.


neteler (approved):
(I believe that you may just merge this right away)

PR #382, opened 2020-12-31
r.bearing.distance r.skyline v.percolate: added a new features

mwlake:
Three new modules currently being used for academic research (but of wider application) which I hope to publish with colleagues in 2021.  You can read the fully formatted man pages at [http://www.homepages.ucl.ac.uk/~tcrnmar/](http://www.homepages.ucl.ac.uk/~tcrnmar/).  Enjoy and best wishes for a saner 2021!


neteler:
Thanks for your submission! Here some feedback:

**Code checklist:**

- license check: looks ok (declared in header)
- code style
    - the indentation does not seem to follow the GRASS GIS conventions. It may be worthwhile to reformat the files with `indent`, see https://github.com/OSGeo/grass/tree/master/tools (`grass_indent.sh` + `grass_indent_ALL.sh`)
    - C++ style comments lead at least in the past to compiler errors on some platforms. Preferably written as `/* ... */ `
- HTML style
    - in the code examples we use different HTML tags
    - the `date` part should be commented out (see other modules) as we have left SVN
    - please add a North Carolina sample dataset example to `v.percolate.html`
 - there is a spurious leftover file `r.bearing.distance/main.c~`, `v.percolate.html~` etc.
 - maybe more...
 
**Code submission:**

In general it is preferred to receive multiple addons as separate pull requests but that is no obligation, just makes it easier to test and review.

marisn (commented):
There are a lot of messages guarded by `#ifdef DEBUG`. I would advise to move them to `G_debug()` as it would allow to enable debug output at runtime.
https://grass.osgeo.org/programming7/debug_8c.html

neteler:
I was able to compile all three modules but didn't actually test them.

mwlake:
Just to say that the revised code fixes all the issues identified by Markus and Maris (thanks!).  The only exceptions are:

i) Some code guarded by #ifdef DEBUG which calls similarly guarded debugging functions and so is not easily handled with G_debug().  In any case, this info. is very low level and only of interest to someone working in great detail on the algorithm.

ii) A couple of cases where Maris suggested G_message / warning fatal_error, but where the code in question is simply inserting a suitable message into a buffer which is later presented to the user by the appropriate GRASS function.

I don't have time to add examples to the man pages now, but will be able to do so later (probably in April).

All three modules have been heavily tested as they are in production use for academic research that I hope will be published this year.

neteler:
Thanks for your contribution!

benducke:
Hi. I found that running v.percolate without option "id=" makes it crash (segfault). Weirdly, I should not even be able to run it without that option: The module interface description says that "id=" is required.

But it seems to have a default value that points to unallocated memory.

This is what valgrind says:
Invalid read of size 8
    at 0x4B1E383: db_close_database (in /opt/grass/lib/libgrass_dbmiclient.7.8.so)
    by 0x4B2103E: db_close_database_shutdown_driver (in /opt/grass/lib/libgrass_dbmiclient.7.8.so)
    by 0x40526D: read_input_vector (vector.c:149)
    by 0x402749: main (main.c:268)
  Address 0x310 is not stack'd, malloc'd or (recently) free'd

PR #384, opened 2021-01-01
v.in.pygbif: fix lazy import of 'osgeo' module

tmszi:
**Error message** (during compilation on the server)**:**

```
ERROR: Unable to load GDAL Python bindings (requires package 'python-gdal'
       or Python library GDAL to be installed).
```

[Compilation log file](https://wingrass.fsv.cvut.cz/grass78/x86_64/addons/latest/logs/v.in.pygbif.log).

We need to move 'osgeo' module import into main() function body. Same as with the [g.proj.identify](https://github.com/OSGeo/grass-addons/blob/master/grass7/general/g.proj.identify/g.proj.identify.py#L145) add-on, which was compiled successfully.

### DIFF
--- a/grass7/vector/v.in.pygbif/v.in.pygbif.py
+++ b/grass7/vector/v.in.pygbif/v.in.pygbif.py
@@ -209,6 +209,13 @@ def main():
     from dateutil.parser import parse
 
     try:
+        from osgeo import ogr, osr
+        from osgeo import __version__ as gdal_version
+    except ImportError:
+        grass.fatal(_("Unable to load GDAL Python bindings (requires "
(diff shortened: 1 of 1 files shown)


ninsbl (commented):
Shouldn't then the pygbif import moved as well? I assume that it will fail next.

tmszi:
> Shouldn't then the pygbif import moved as well? I assume that it will fail next.

Yes you are right, but pygbif module is [already moved inside main() function body](https://github.com/OSGeo/grass-addons/blob/master/grass7/vector/v.in.pygbif/v.in.pygbif.py#L211).

ninsbl (approved):
I see. Please go ahead and merge.

PR #392, opened 2021-01-06
v.what.strds.timestamp: fix v.what.rast runs twice

pesekon2:


grass7/vector/v.what.strds.timestamp/v.what.strds.timestamp.py
```diff
@@ -123,12 +123,15 @@ def sample_absolute(input, layer, timestamp_column, column, t_raster,
         grass.run_command("g.region", n=treg['n'], s=treg['s'], e=treg['e'],
                           w=treg['w'], align=raster_map)
 
+        if i_flag is True:
+            flags = 'i'
+        else:
+            flags = ''
+
         # Sample spatio-temporally matching points and raster map
-        rast_what = Module('v.what.rast', map=input, layer=layer,
```
ninsbl, line 127: Wouldn't it be enough to add `, run_=False` to the module object?

ninsbl (approved):
Good catch! Thanks. See one comment inline. But this implementation would be ok as well. Just a bit more code...

grass7/vector/v.what.strds.timestamp/v.what.strds.timestamp.py
pesekon2: True. As I have just never used `run_`, I don't know anything about speed, what is the recommended way, anything - you say that you would prefer to use the `run_` argument? It's not a problem to change the PR as it is going to be squashed in any case.

ninsbl: I don`t have a strong opinion. Using `run_=False` saves us an `if-else` and personally I would probably prefer it. However, that is no big issue. So I have no objection against our solution...

PR #395, opened 2021-01-08
r.valley.bottom performance improvements

stevenpawley:
I've made some tweaks to the r.valley.bottom module; it is faster now by using simpler r.mapcalc statements for the elevation percentile, and I've cleaned-up / commented the code. I've also cut down the large number of messages which were sent to the grass console, instead providing a more concise summary similar to what SAGA-GIS provides. Otherwise the output MrVBF/MrRTF should be the same.

### DIFF
--- a/grass7/raster/r.valley.bottom/r.valley.bottom.py
+++ b/grass7/raster/r.valley.bottom/r.valley.bottom.py
@@ -505,21 +505,25 @@ def smooth_dem(DEM):
 def refine(input, region, method="bilinear"):
     """change resolution back to base resolution and resample a raster"""
 
+    # grow input
     input_padded = rand_id("padded{}".format(L + 1))
-    TMP_RAST[L].append(input_padded)
     cell_padding(input=input, output=input_padded, radius=2)
 
-    region.write()
-    input = rand_id("refined{}".format(L + 1))
+    # resample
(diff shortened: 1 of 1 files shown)


hellik (approved):

grass7/raster/r.valley.bottom/r.valley.bottom.py
```diff
@@ -106,536 +106,638 @@
...
-                grass.run_command(
-                    "g.remove", type="raster", name=f, flags="f", quiet=True)
+            if len(gs.find_file(f)["fullname"]) > 0:
+                g.remove(type="raster", name=f, flags="f", quiet=True)
+
+    current_region.write()
 
-    Region.write(current_region)
+
+def rand_id(prefix):
```
ninsbl, line 138: Did you see: https://grass.osgeo.org/grass76/manuals/libpython/script.html#script.core.tempname

stevenpawley: Thanks Stefan,

In this case I'd like a custom name with a prefix because otherwise it's hard to debug r.valley.bottom, given the many processing steps with different outputs. That could be a potential enhancement for gs.tempname, i.e. the option of a prefix (like R's 'pattern') to generate something like "tmp_{prefix}_xxxxxx"?

Steve

PR #357, opened 2020-12-02
PEP8: bulk fixes on entire grass7/ subdirectory tree with flake8

neteler:
While not being a big fan of bulk fixes I have taken liberty to apply flake8 recursively on `grass-addons/grass7/`.

The fixes are separated by commit in order to more easily review them.
Eventually, after review, all can be squashed into one commit.

Rationale: follow more closely our coding style: https://trac.osgeo.org/grass/wiki/Submitting/Python#PEP8standardStyle

Once accepted, the [GitHub Action flake8 test](https://github.com/OSGeo/grass/blob/master/.github/workflows/flake8.yml) from  core GRASS GIS may be used also here.

The script I wrote to auto-create this PR:

```
cat pep8_formatting_commit.sh 
#!/bin/sh

# requires "python3-autopep8"

# put desired error codes to be fixed into CSV file
# https://pycodestyle.pycqa.org/en/latest/intro.html#error-codes
#
# format:
# E117 over-indented
# E122 continuation line missing indentation or outdented
# ...

CODES=~/pep8_formatting_codes_addons.csv

# trick: using a non-english char as a special separator
for i in $(cat $CODES | sed 's+ +ä+g') ; do
    code="$(echo $i | cut -d'ä' -f1)"
    desc=$(echo $i | cut -d'ä' -f2- | sed 's+ä+ +g')
    echo "Processing: $code: $desc..."
    ag --python -l | xargs autopep8 --select $code --aggressive --aggressive --in-place --jobs 4
    LISTTOCOMMIT=$(git status | grep modified | tr -s ' ' ' ' | cut -d' ' -f2)
    git commit -m"PEP8: fix $code: $desc" $LISTTOCOMMIT
done
```

PR #290, opened 2020-09-28
PEP8: fix flake8 errors

neteler:
This PR fixes PEP8 errors identified with flake8 and selectively fixed with autopep8.

For easier review, the different errors and warnings are separate out into single commits.

PR #399, opened 2021-01-10
GitHub action: added flake8 tests

### LINKED ISSUE #371: [Feat] Add flake8 tests as a new GH Action like in GRASS-core
After the recent merge of  "PEP8: bulk fixes on entire grass7/ subdirectory tree with flake8" (#357) it might be a good idea to set up a new GitHub Action like https://github.com/OSGeo/grass/blob/master/.github/workflows/flake8.yml in core.

neteler:
yml file:
workflows/flake8.yml: modified from https://github.com/OSGeo/grass/blob/master/.github/workflows/flake8.yml

The .flake8 files have been copied from:
- `gui/wxpython/.flake8` file: copied from https://github.com/OSGeo/grass/blob/master/gui/wxpython/.flake8
- `.flake8.yml` file: copied from https://github.com/OSGeo/grass/blob/master/scripts/.flake8

Fixes #371


neteler:
Mhh, not sure why the following fails here while being fine in GRASS-core/scripts/ with identical `.flake8` settings/ignore file?

(random example):
```
./i.modis/i.modis.download/i.modis.download.py:272:25: F405 '_' may be undefined, or defined from star imports: datetime
            grass.fatal(_("Error during connection"))
                        ^
```

@wenzeslaus do you have a hint by chance?

wenzeslaus:
> Mhh, not sure why the following fails here while being fine in GRASS-core/scripts/ with identical `.flake8` settings/ignore file?

F405 from your example is suppressed in `gui/wxpython/.flake8` and `lib/python/.flake8` in core. F821 is everywhere exactly to suppress missing `_`. The problem you hit there is probably a start import. It doesn't know if it is F821 case, because of the start import, so it reports F405. If you don't like the warning and you get it, then ignore it. The Flake8 philosophy is not more complicated that that.

The underlying issue is, however, much more serious. We add a build-in `_`. Flake8 has a problem with that (well, we just defied the definition of a build-in). Now we need ignore all undefined variables.

Although Pylint can deal with that situation, the underlying issue of `_` as a additional build-in is still there causing a lot of troubles and anomalies elsewhere.

wenzeslaus (changes requested):
This looks like a lot of duplication duplication. The reason for multiple .flake8 files in core is very different code bases in the different directories. This should not be the case here in addons. 

Flake8 has a `--config` parameter if that would help you to run the processes in different directories in parallel.

The multiple jobs can be replaced by matrix with a list of directories.

neteler:
The updated outcome looks much better.
Still some Python errors to be fixed in some modules but that's for different PRs.

.github/workflows/flake8.yml
```diff
@@ -0,0 +1,226 @@
...
+      with:
+        python-version: 3.8
+    - name: Install
+      run: |
+        python -m pip install --upgrade pip
+        pip install flake8==3.8.0
+    - name: Run Flake8
+      run: |
+        cd grass7/db
+        flake8 --config=../.flake8 --count --statistics --show-source --jobs=$(nproc) .
```
wenzeslaus, line 26: You can reduce the duplicity further with matrix as in [gcc.yml](https://github.com/OSGeo/grass/blob/master/.github/workflows/gcc.yml#L12) in core:

```
flake8-job:
  strategy:
    matrix:
      directory:
      - db
      - display
      ...

     cd grass7/${{ matrix.directory }}
```

.github/workflows/flake8.yml
```diff
@@ -0,0 +1,226 @@
...
+
+
+  flake8-grass7-display-scripts:
+
+    runs-on: ubuntu-18.04
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Python
+      uses: actions/setup-python@v1
```
wenzeslaus, line 36: IIRC, both of these have v2 now.

.github/workflows/flake8.yml
```diff
@@ -0,0 +1,226 @@
...
+        pip install flake8==3.8.0
+    - name: Run Flake8
+      run: |
+        cd grass7/db
+        flake8 --config=../.flake8 --count --statistics --show-source --jobs=$(nproc) .
+
+
+  flake8-grass7-display-scripts:
+
+    runs-on: ubuntu-18.04
```
wenzeslaus, line 31: I would just go with 20.04 here. (It is now matured as ubuntu-latest in GH Action and so I would use it by default unless there is a reason to use something else.)

.github/workflows/flake8.yml
```diff
@@ -0,0 +1,226 @@
...
+      with:
+        python-version: 3.8
+    - name: Install
+      run: |
+        python -m pip install --upgrade pip
+        pip install flake8==3.8.0
+    - name: Run Flake8
+      run: |
+        cd grass7/temporal
+        flake8 --config=../.flake8 --count --statistics --show-source --jobs=$(nproc) .
```
wenzeslaus, line 186: This seems to use the main .flake8 file while it has its own in this PR.

grass7/temporal/.flake8
```diff
@@ -0,0 +1,46 @@
+[flake8]
+ignore =
+    E265, # block comment should start with '# '
+    E266, # too many leading '#' for block comment
+    E402, # module level import not at top of file
+    E722, # do not use bare 'except'
```
wenzeslaus, line 6: There are also `per-file-ignores` as in: https://github.com/OSGeo/grass/blob/master/lib/python/.flake8#L34 if that would be helpful here to avoid whole another file (well, unless it is two mostly non-overlapping sets).

wenzeslaus (changes requested):
I have several suggestions to make to reduce the duplicity further.

.github/workflows/flake8.yml
neteler: Sure fine with 20.04.

.github/workflows/flake8.yml
neteler: I'll give it a try.

.github/workflows/flake8.yml
neteler: Great suggestion, @wenzeslaus - seems to run much faster now after 6f28a0fbb7aabd245de51aac4b5f86330bd8ba6d

.github/workflows/flake8.yml
```diff
@@ -0,0 +1,59 @@
...
+
+on:
+- push
+- pull_request
+
+jobs:
+  flake8-grass7-scripts:
+    runs-on: ubuntu-20.04
+
+    cd grass7/${{ matrix.directory }}
```
wenzeslaus, line 11: I don't think this is supposed to work. I don't see the jobs running either.

neteler: Well, the new attempt starts but fails miserably. No idea how it is written properly...

neteler:
@wenzeslaus , all: thanks to the fixes by @mmacata it seems to be functional now!

grass7/.flake8
```diff
@@ -0,0 +1,50 @@
...
+    __pycache__,
+    .env,
+    .venv,
+    env,
+    venv,
+    ENV,
+    env.bak,
+    venv.bak,
+    ctypes,
+    pydispatch,
```
wenzeslaus, line 49: No need to exclude ctypes and pydispatch. There are no files like this. These are in core, in `lib/python`. (Same of the other flake file.)

.github/workflows/flake8.yml
```diff
@@ -0,0 +1,66 @@
...
+      with:
+        python-version: 3.8
+    - name: Install
+      run: |
+        python -m pip install --upgrade pip
+        pip install flake8==3.8.0
+    - name: Run Flake8
+      # use own local flake8 settings file
+      run: |
+        flake8 --config=grass7/gui/wxpython/.flake8 --count --statistics --show-source --jobs=$(nproc) ${{ matrix.directory }}
```
wenzeslaus, line 66: No `matrix.directory` in this job.

.github/workflows/flake8.yml
```diff
@@ -0,0 +1,66 @@
...
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  # This workflow contains two job called "flake8-grass7-scripts" and "flake8-gui-wxgui-scripts"
```
wenzeslaus, line 15:
```suggestion
```

I can see that! :-)

.github/workflows/flake8.yml
```diff
@@ -0,0 +1,66 @@
...
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+    - name: Install
+      run: |
+        python -m pip install --upgrade pip
+        pip install flake8==3.8.0
+    - name: Run Flake8 in ${{ matrix.directory }}
+      run: |
+        flake8 --config=../.flake8 --count --statistics --show-source --jobs=$(nproc) ${{ matrix.directory }}
```
wenzeslaus, line 47: You removed the `cd ...` but left `../.flake8` here. Consequently, `../` goes outside of the cloned source code. This is normal Bash, just try the same command locally and see how it behaves.

wenzeslaus (changes requested):

.github/workflows/flake8.yml
neteler: Still it may be helpful to newcomers :)

grass7/.flake8
neteler, line 49:
```suggestion
```

.github/workflows/flake8.yml
neteler: Fixes added...

.github/workflows/flake8.yml
neteler: Fixed

.github/workflows/flake8.yml
neteler: Fixed

grass7/gui/wxpython/.flake8
```diff
@@ -0,0 +1,63 @@
...
+    __pycache__,
+    .env,
+    .venv,
+    env,
+    venv,
+    ENV,
+    env.bak,
+    venv.bak,
+    ctypes,
+    pydispatch,
```
wenzeslaus, line 62: As I said, same problem is in both files.

.github/workflows/flake8.yml
wenzeslaus: Let me invoke the rule of writing comments in the source code. Don't comment on the obvious things. We don't do:

```python
# list which contains 1 and 2
a = [1, 2]
```

Surely this would be helpful to people who has newer seen Python before, but a simple search on Python syntax reveals what `a = [1, 2]` is. Same for:
```yaml
jobs:
  flake8-grass7-scripts:
    ...
  flake8-grass7-gui:
    ...
```

I'm far from advocating for minimal comments or doc (see, e.g., #20). However, things which should be documented are intention and non-obvious syntax or behavior. This is YAML, not regular expressions, Bash, or [Whitespace](https://en.wikipedia.org/wiki/Whitespace_(programming_language)).

If you want to document overall intention of this script, why the strategy, etc, that would be something different.

wenzeslaus (changes requested):

.github/workflows/flake8.yml
```diff
@@ -0,0 +1,66 @@
...
+    - name: Install
+      run: |
+        python -m pip install --upgrade pip
+        pip install flake8==3.8.0
+    - name: Run Flake8 in ${{ matrix.directory }}
+      run: |
+        flake8 --config=grass7/.flake8 --count --statistics --show-source --jobs=$(nproc) ${{ matrix.directory }}
+
+
+  flake8-gui-wxgui-scripts:
```
wenzeslaus, line 50: Name inconsistent with `flake8-grass7-scripts` and probably over-complicated. What about  `flake8-grass7-wxgui`?

I think we will revise the names as we are adding more workflows and jobs (or consolidating as in case of these two hopefully), but in any case, they are what is shown in Checks of a PR, so it should be informative for the contributor and preferably also short.

.github/workflows/flake8.yml
```diff
@@ -0,0 +1,66 @@
...
+    branches: [ master ]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  # This workflow contains two job called "flake8-grass7-scripts" and "flake8-gui-wxgui-scripts"
+
+  flake8-grass7-scripts:
+    runs-on: ubuntu-20.04
```
wenzeslaus, line 18: Here `name:` would probably do a better job of telling which piece of the matrix is running in the Checks overview. See e.g. gcc.yml in core. The current automatic one is e.g. _flake8-grass7-scripts (grass7/general)_ but it could be just `Flake8 ${{ matrix.directory }}` (so e.g. _Flake8 grass7/general_).

wenzeslaus (changes requested):
Some naming ideas.

.github/workflows/flake8.yml
neteler: fixed

.github/workflows/flake8.yml
neteler: alright!

grass7/gui/wxpython/.flake8
neteler: fixed

.github/workflows/flake8.yml
neteler: fixed

grass7/temporal/.flake8
neteler: I was simply following the GRASS-core repo

wenzeslaus: The core repo has 4 very different code bases (with different styles and errors), so that's the reason for different .flake8 files there (for now!). It is also using different job definitions there, but my reasons for that are no longer valid (Black or Pylint would now go to a separate file, for example), so I should redo it.

wenzeslaus:
Looking at the errors, you definitively want to use the `per-file-ignores`: Many errors of the same type often coming from one module. https://github.com/OSGeo/grass/blob/master/lib/python/.flake8#L34

neteler:
@wenzeslaus - does it makes sense (after reduction to a single .flake8 file as you proposed) to sync the `flake8.yml` here with that of your new PR https://github.com/OSGeo/grass/pull/1309 ?

wenzeslaus:
> does it makes sense (after reduction to a single .flake8 file as you proposed) to sync the `flake8.yml` here with that of your new PR [OSGeo/grass#1309](https://github.com/OSGeo/grass/pull/1309) ?

Sync names, version? Yes, why not. Single .flake8 file? Only when needed/useful. I think all the raster, vector, temporal, ... have similar issues, so one .flak8 file makes sense. The gui dir seems to have different ones, although perhaps a superset.

So, you have two groups of directories, those with their own .flake8 file and those with a custom .flake8 file (set of 1). This seems to fit best with defining two separate job definituions as you currently have it in the PR (you could achieve the same with matrix, but two job definitions are easier to write).

Ideally, there would be of course only one .flake8 file which would be short, but that would require fixing the issues, so not this PR.

.github/workflows/flake8.yml
```diff
@@ -0,0 +1,64 @@
...
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+
+    - name: Install
+      run: |
+        python -m pip install --upgrade pip
+        pip install flake8==3.8.4
+    - name: Run Flake8-gui-wxgui
```
wenzeslaus, line 60: Either make it like the other one, i.e., "Run Flake8 in grass7/gui/wxpython/" or even `Run Flake8` would be quite sufficient since this is name of the step, not the job.

.github/workflows/flake8.yml
```diff
@@ -0,0 +1,64 @@
...
+        pip install flake8==3.8.4
+
+    - name: Run Flake8 in ${{ matrix.directory }}
+      run: |
+        cd ${{ matrix.directory }}
+        flake8 --config=../.flake8 --count --statistics --show-source --jobs=$(nproc) .
+
+
+  flake8-gui-wxgui:
+    name: flake8 grass7/gui
```
wenzeslaus, line 46: The workflow name says "Python Flake8 Code Quality" so no need to repeat "flake8" here. Here you want to make it succinct since there is not much space for check names in the GitHub GUI.

wenzeslaus (changes requested):
Well, the checks are failing. :-) As I was saying before:

> Looking at the errors, you definitively want to use the per-file-ignores: Many errors of the same type often coming from one module. https://github.com/OSGeo/grass/blob/master/lib/python/.flake8#L34

.github/workflows/flake8.yml
neteler: sure, fixed

.github/workflows/flake8.yml
neteler, line 60:
```suggestion
    - name: Run Flake8
```

.github/workflows/flake8.yml
neteler, line 46:
```suggestion
    name: grass7/gui
```

neteler:
> Well, the checks are failing. :-) 

Yes, I see that... Numerous PEP8 errors I had already fixed in separate PRs (#290, #357, #403, #426).

Do you think the remaining problems should be part of this PR?

grass7/imagery:
```
25    E122 continuation line missing indentation or outdented
1     E131 continuation line unaligned for hanging indent
1     E712 comparison to False should be 'if cond is False:' or 'if not cond:'
1     E999 SyntaxError: invalid syntax
2     F523 '...'.format(...) has unused arguments at position(s): 0
2     F811 redefinition of unused 'transform' from line 153
1     W291 trailing whitespace
```

grass7/raster
```
275   E101 indentation contains mixed spaces and tabs
2     E112 expected an indented block
34    E114 indentation is not a multiple of four (comment)
29    E117 over-indented (comment)
2     E712 comparison to False should be 'if cond is False:' or 'if not cond:'
3     E999 SyntaxError: invalid syntax
1     F402 import 'ant' from line 15 shadowed by loop variable
2     F507 '...' % ... has 5 placeholder(s) but 7 substitution(s)
1     F509 '...' % ... has unsupported format character '='
1     F521 '...'.format(...) has invalid format string: Single '}' encountered in format string
3     F522 '...'.format(...) has unused named argument(s): overwrite
1     F523 '...'.format(...) has unused arguments at position(s): 0
10    F632 use ==/!= to compare constant literals (str, bytes, int, float, tuple)
2     F811 redefinition of unused 'test_getpenalty' from line 84
5     W191 indentation contains tabs
3     W291 trailing whitespace
```

grass7/vector
```
1     E114 indentation is not a multiple of four (comment)
1     E117 over-indented (comment)
6     E131 continuation line unaligned for hanging indent
2     E704 multiple statements on one line (def)
2     F523 '...'.format(...) has unused arguments at position(s): 0
34    F632 use ==/!= to compare constant literals (str, bytes, int, float, tuple)
1     F811 redefinition of unused 'grass' from line 67
```

grass7/gui
```
14    E101 indentation contains mixed spaces and tabs
2     E114 indentation is not a multiple of four (comment)
2     E502 the backslash is redundant between brackets
12    E999 SyntaxError: invalid syntax
4     F632 use ==/!= to compare constant literals (str, bytes, int, float, tuple)
5     F811 redefinition of unused 'GMessage' from line 35
14    W191 indentation contains tabs
```

So, is the acceptance criterion here that all checks must be successful? Then this PR may stay in "waiting" till it has happened.

wenzeslaus:
> Do you think the remaining problems should be part of this PR?

I think the approach in core was to do ignores in the GH Action PR and then start fixing in subsequent PRs. Not necessary, but probably more transparent. As I was saying before:

> > Looking at the errors, you definitively want to use the per-file-ignores: Many errors of the same type often coming from one module. https://github.com/OSGeo/grass/blob/master/lib/python/.flake8#L34

> So, is the acceptance criterion here that all checks must be successful? Then this PR may stay in "waiting" till it has happened.

For checks to pass, it is enough if all non-ignored Flake8 errors are fixed in the code in linted directories. I don't see what would be the alternative as failing checks on new PRs would be quite uninformative. If we would expect that it gets fixed in next 24 hours, then maybe it would make sense to merge with failing checks, but otherwise `ignore` and `per-file-ignores` is the way to go.

Did you ever update this branch to master branch to get fixes from PRs which were merged after this branch was created?

neteler:
Huh, accidentially e7c556c slipped in here ... how to remove it from the PR?

wenzeslaus:
> how to remove it from the PR?

No need. As the update/comment says "neteler_added a _commit to neteler/grass-addons that referenced_ this pull request." You referenced this PR in the commit - that's all. (Perhaps an unnecessary reference. We want the issues fixed regardless of this PR.)

See Commits tab for actual list of commits in a PR. See [GitFixUm](https://sethrobertson.github.io/GitFixUm/fixup.html) for undoing changes (like accidental commit to a wrong branch).

wenzeslaus (approved):

neteler:
Thanks for your patience in reviewing this PR, @wenzeslaus !

wenzeslaus:
Whew! :rocket:

PR #409, opened 2021-01-19
r.confusionmatrix: Support usage of a single class in reference

griembauer:
This PR:
- allows a successful processing of `r.confusionmatrix` even if only one class is given in the reference map (a warning will be given to the user)


anikaweinmann (approved):

neteler (approved):
Thanks!

PR #410, opened 2021-01-19
add option to i.sentinel.coverage

lucadelu:
@neteler, @anikaweinmann does it make sense to have producttype also here?

it was missing for my use case....

### DIFF
--- a/grass7/imagery/i.sentinel/i.sentinel.coverage/i.sentinel.coverage.py
+++ b/grass7/imagery/i.sentinel/i.sentinel.coverage/i.sentinel.coverage.py
@@ -63,6 +63,15 @@
 #%end
 
 #%option
+#% key: producttype
+#% type: string
+#% description: Sentinel product type to filter
+#% required: no
+#% options: SLC,GRD,OCN,S2MSI1C,S2MSI2A,S2MSI2Ap
+#% guisection: Filter
+#%end
+
+#%option
 #% key: clouds
 #% type: integer
 #% required: no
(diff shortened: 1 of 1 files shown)


anikaweinmann:
@lucadelu yes it would make sense. But for @griembauer and me it would make more sense, if the `type` would be replaced by the `producttype`.

lucadelu:
Did I understand correctly?

griembauer:
@lucadelu Thank you very much! I think this would also require some adaptions in the `scenename_split`-function. This function automatically pulls start and end date from the filename if a filename is provided by the user. However, this function can so far only handle S2-data and S1-GRD data - I am not sure how well it can handle S-1 `SLC` and `OCN` products.

lucadelu:
@griembauer thanks to catch the problem in `scenename_split` I updated it. About the problem with S1 it seems more related to  the level of product, S1 Level 1 GRD and SLC should have the same name format that is different from S1 Level 2 GRD/SLC/OCN. But I think this is another problem that should not address in this pull request.

anikaweinmann (approved):
Thank you

anikaweinmann:
> @griembauer thanks to catch the problem in `scenename_split` I updated it. About the problem with S1 it seems more related to the level of product, S1 Level 1 GRD and SLC should have the same name format that is different from S1 Level 2 GRD/SLC/OCN. But I think this is another problem that should not address in this pull request.

For me a new pull request later for this problem is fine

lucadelu:
> > @griembauer thanks to catch the problem in `scenename_split` I updated it. About the problem with S1 it seems more related to the level of product, S1 Level 1 GRD and SLC should have the same name format that is different from S1 Level 2 GRD/SLC/OCN. But I think this is another problem that should not address in this pull request.
> 
> For me a new pull request later for this problem is fine

I created a new issue to not forget about it #413

PR #421, opened 2021-02-01
i.landsat.download: swap lon/lat order due to landsatxplore update

griembauer:
In the latest release of [landsatxplore](https://github.com/yannforget/landsatxplore) the `bbox` parameter was adapted to be consistent with the documentation (xmin, ymin, xmax, ymax). See also https://github.com/yannforget/landsatxplore/pull/30/files.

This PR swaps the respective parts of `i.landsat.download` to be consistent with this update.

### DIFF
--- a/grass7/imagery/i.landsat/i.landsat.download/i.landsat.download.py
+++ b/grass7/imagery/i.landsat/i.landsat.download/i.landsat.download.py
@@ -121,10 +121,10 @@ def get_bb(vector=None):
         gs.fatal("Unable to get bounding box: unprojected location not supported")
     if kv["+proj"] != "longlat":
         info = gs.parse_command("g.region", flags="uplg", **args)
-        return (info["sw_lat"], info["nw_long"], info["nw_lat"], info["ne_long"])
+        return (info["nw_long"], info["sw_lat"], info["ne_long"], info["nw_lat"])
     else:
         info = gs.parse_command("g.region", flags="upg", **args)
-        return (info["s"], info["w"], info["n"], info["e"])
+        return (info["w"], info["s"], info["e"], info["n"])
 
 
 def main():


veroandreo:
Thanks Guido! 

Did you test it? Because I had swapped that when writing the code and opened an issue in landsatxplore about this (https://github.com/yannforget/landsatxplore/issues/23).

I have my doubts though that the new change in landsatxplore is right. I think it is swapped again (See my comment here:  https://github.com/yannforget/landsatxplore/pull/30/files#r567766086). Do you understand the same?

griembauer:
> Thanks Guido!
> 
> Did you test it? Because I had swapped that when writing the code and opened an issue in landsatxplore about this ([yannforget/landsatxplore#23](https://github.com/yannforget/landsatxplore/issues/23)).
> 
> I have my doubts though that the new change in landsatxplore is right. I think it is swapped again (See my comment here: https://github.com/yannforget/landsatxplore/pull/30/files#r567766086). Do you understand the same?

Hi Veronica, thanks for checking. I checked again at your comment in `landsatxplore`: https://github.com/yannforget/landsatxplore/pull/30/files#r567834534 where to me it seems the internal logic of `landsatxplore` is still correct. However, I did not do a test case with `i.landsat.download` yet, but I will do so and let you know the outcome

griembauer:
I used `i.landsat.download` with the newest version of `landsatxplore` to search for some Landsat-8 scenes over Bonn, Germany, and found that I only get correct results using the version of this PR. 
In https://github.com/yannforget/landsatxplore/pull/30, an "internal logic" error was fixed in `landsatxplore` by swapping the x/y order here: https://github.com/yannforget/landsatxplore/blob/cd3660bd93da839a7b51b52f6425cb6eb1551faf/landsatxplore/datamodels.py#L49 
Before the fix, `i.landsat.download` "made use" of this error by passing the bbox parameters accordingly. Since this has now again changed, the adaptions in `i.landsat.download` are necessary. 
The "double swapping" happens only for the case where landsatxplore is used with the `longitude/latitude` parameters instead of `bbox`.

veroandreo:
> > Thanks Guido!
> > Did you test it? Because I had swapped that when writing the code and opened an issue in landsatxplore about this ([yannforget/landsatxplore#23](https://github.com/yannforget/landsatxplore/issues/23)).
> > I have my doubts though that the new change in landsatxplore is right. I think it is swapped again (See my comment here: https://github.com/yannforget/landsatxplore/pull/30/files#r567766086). Do you understand the same?
> 
> Hi Veronica, thanks for checking. I checked again at your comment in `landsatxplore`: https://github.com/yannforget/landsatxplore/pull/30/files#r567834534 where to me it seems the internal logic of `landsatxplore` is still correct. However, I did not do a test case with `i.landsat.download` yet, but I will do so and let you know the outcome

Hi Guido, 

I just updated to the latest version of landsatxplore and tested the output of `i.landsat.download` in the landsat mapset of NC location. Indeed, I get the wrong path and row for the computational region set with the `elevation` map from PERMANENT: 

```
GRASS nc_spm_08_grass7/landsat:~ > i.landsat.download -l settings=credentials.txt dataset=LANDSAT_8_C1 clouds=5 start='2020-02-01' end='2020-04-30'
2 scenes found.
LC81381162020033LGN00 LC08_L1GT_138116_20200202_20200211_01_T2 2020-02-02 0.00
LC81361172020035LGN00 LC08_L1GT_136117_20200204_20200211_01_T2 2020-02-04 0.20
```

With the old landsatxplore version that had lat/long swapped (but the right order in i.landsat.download), I got it right: 
```
i.landsat.download -l settings=credentials.txt dataset=LANDSAT_8_C1 clouds=5 start='2020-02-01' end='2020-04-30'
1 scenes found.
LC80160352020058LGN00 LC08_L1TP_016035_20200227_20200313_01_T1 2020-02-27 0.37
```

Hence, your change is relevant and correct here :)
Thanks!

veroandreo (approved):

veroandreo:
Seems we were testing at the same time, @griembauer !

griembauer:
Indeed, Thanks a lot for testing too! Swapping x and y is a classic that never gets old ;)

neteler:
> Indeed, Thanks a lot for testing too! Swapping x and y is a classic that never gets old ;)

(so true, this one was started in 2006: https://wiki.osgeo.org/wiki/Axis_Order_Confusion)

PR #427, opened 2021-02-11
i.sentinel.coverage: print result when no output defined

neteler:
Instead of no output print the module now prints the result into the terminal when no output file was defined.


griembauer (approved):
Thank you, useful addition!

PR #450, opened 2021-02-25
WIP: add collection 2 support in i.landsat.download

veroandreo:
Here's a first working version of `i.landsat.download` with support for collection 2 data conforming with the new landsatxplore.

There have been some changes in terms of metadata in Landsat collections. Hence, instead of `acquisitionDate`, I have (for now) used `publishDate`. This has a problem though. While it has info for collection 1, it appears as `Unknown` for L8 in collection 2 only. Hence preventing the sort these data by date.

The option would be to use `scene['temporalCoverage']['startDate']` that is populated properly in both collection 1 and 2, but my limited python knowledge *prevents* me from modifying current lines 233-242 where date and clouds are used to sort the listed scenes. Can someone help here, please?? :pray: 

Here's my hardcoded testing script:

```
#!/usr/bin/env python

import landsatxplore.api

# Initialize a new API instance and get an access key
api = landsatxplore.api.API('your_usgs_username, 'your_usgs_password')

# Request
scenes = api.search(
    dataset='landsat_ot_c2_l2',
    longitude=-78.77428134,
    latitude=35.68792712,
    start_date='2015-01-01',
    end_date='2015-07-01',
    max_cloud_cover=80)

print('{} scenes found.'.format(len(scenes)))

sorted_scenes = sorted(scenes, key=lambda i: (i['temporalCoverage']['startDate'],i['cloudCover']))

# Output list of scenes found
print('ID', 'DisplayID', 'Date', 'Clouds')
for scene in sorted_scenes:
    print(scene['entityId'], scene['displayId'], scene['temporalCoverage']['startDate'], scene['cloudCover'])

api.logout()
```

### DIFF
--- a/grass7/imagery/i.landsat/i.landsat.download/i.landsat.download.py
+++ b/grass7/imagery/i.landsat/i.landsat.download/i.landsat.download.py
@@ -210,7 +210,7 @@ def main():
 
             try:
 
-                ee.download(scene_id=i, output_dir=outdir, timeout=options["timeout"])
+                ee.download(scene_id=i, output_dir=outdir, timeout=int(options["timeout"]))
 
             except OSError:
 
--- a/grass7/imagery/i.landsat/i.landsat.download/i.landsat.download.py
+++ b/grass7/imagery/i.landsat/i.landsat.download/i.landsat.download.py
@@ -95,8 +95,8 @@
 #% key: sort
 #% description: Sort by values in given order
 #% multiple: yes
-#% options: publishDate,cloudCover
-#% answer: cloudCover,publishDate
+#% options: acquisitionDate,cloudCover
+#% answer: cloudCover,acquisitionDate
 #% guisection: Sort
 #%end
 
@@ -106,6 +106,14 @@
 #% options: asc,desc
 #% answer: asc
 #% guisection: Sort
+#%end
+
+#%option
+#% key: timeout
+#% type: integer
+#% description: Download timeout in seconds
+#% answer: 300
+#% guisection: Optional
 #%end
 
 #%flag
@@ -202,7 +210,7 @@ def main():
 
             try:
 
-                ee.download(scene_id=i, output_dir=outdir)
+                ee.download(scene_id=i, output_dir=outdir, timeout=options["timeout"])
 
             except OSError:
 
@@ -236,10 +244,26 @@ def main():
         if options["order"] == "desc":
             reverse = True
 
-        # Sort scenes
-        sorted_scenes = sorted(
-            scenes, key=lambda i: (i[sort_vars[0]], i[sort_vars[1]]), reverse=reverse
+        # auxiliary list of dictionaries with the entries we need
+        scenes_extracted = []
+        for idx,scene in enumerate(scenes):
+            scene_dict = {
+                'idx': idx,
+                'cloudCover': float(scene['cloudCover']),
+                'acquisitionDate': datetime.strptime(scene['temporalCoverage']['startDate'], '%Y-%m-%d %H:%M:%S')
+            }
+            scenes_extracted.append(scene_dict)
+
+        # sort auxiliary list of dictionaries and apply order to original scenes-list
+        scenes_extracted_sorted = sorted(
+            scenes_extracted, key=lambda i: (i[sort_vars[0]], i[sort_vars[1]]), reverse=reverse
         )
+        sorted_idcs = [scene['idx'] for scene in scenes_extracted_sorted]
+        sorted_scenes = [scenes[i] for i in sorted_idcs]
+        # # Sort scenes
+        # sorted_scenes = sorted(
+        #     scenes, key=lambda i: (i[sort_vars[0]], i[sort_vars[1]]), reverse=reverse
+        # )
 
         landsat_api.logout()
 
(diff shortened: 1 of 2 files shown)


veroandreo:
The issue with `acquisition_date` has been fixed now in landsatxplore v0.12.1 https://github.com/yannforget/landsatxplore/commit/1a58e6052891f951b25ce95143e8b1d1993300bc. I will update this PR accordingly :)

grass7/imagery/i.landsat/i.landsat.download/i.landsat.download.py
```diff
@@ -238,7 +251,7 @@ def main():
                 print(
                     scene["entityId"],
                     scene["displayId"],
-                    scene["acquisitionDate"],
```
griembauer, line 241: I think there are several ways to go, depending on the pending `landsatxplore` release. In any case, for the nested structure of the metadata, you could use for example:
`scene['temporalCoverage']['startDate']` --> `scene['temporalCoverage']` is a dictionary itself with the two keys `startDate` and `endDate`
or 
`scene['start_time']`, 
both returning strings that might need further handling if you need them to be datetime objects (?)

griembauer: Sorry, I realised I misunderstood the issue. Will look into the sorted()-problem

veroandreo: > Sorry, I realised I misunderstood the issue. Will look into the sorted()-problem

no problem... Let's see how the `acquisition_date` problem is fixed in landsatxplore. If that is harmonized properly and delivers a properly formatted date, then we can use it directly.

grass7/imagery/i.landsat/i.landsat.download/i.landsat.download.py
```diff
@@ -81,26 +90,30 @@
...
 #% guisection: Filter
 #%end
+
 #%option
 #% key: sort
 #% description: Sort by values in given order
 #% multiple: yes
-#% options: acquisitionDate,cloudCover
-#% answer: cloudCover,acquisitionDate
+#% options: publishDate,cloudCover
```
griembauer, line 98: I would suggest to keep it as `acquisitionDate`, it is more understandable by the user (we can internally change the name later)

grass7/imagery/i.landsat/i.landsat.download/i.landsat.download.py
griembauer: Both the `date` and the `cloudCover` that we can get from the metadata are string objects. This means they have to be converted to float/datetime objects before they can be sorted in a meaningful way. Now the date might already be a `datetime` object in the new `landsatxplore` release. However for now I would suggest the following in lines 233+ (sorry I can't directly comment there, because these lines werent changed by this PR):
```
        sort_vars = options["sort"].split(",")

        reverse = False
        if options["order"] == "desc":
            reverse = True

        scenes_extracted = []
        for idx,scene in enumerate(scenes):
            scene_dict = {
                'idx': idx,
                'cloudCover': float(scene['cloudCover']),
                'acquisitionDate': datetime.strptime(scene['temporalCoverage']['startDate'], '%Y-%m-%d %H:%M:%S')
            }
            scenes_extracted.append(scene_dict)
        scenes_extracted_sorted = sorted(scenes_extracted, key=lambda i: (i[sort_vars[0]], i[sort_vars[1]]), reverse=reverse)
        sorted_idcs = [scene['idx'] for scene in scenes_extracted_sorted]
        sorted_scenes = [scenes[i] for i in sorted_idcs]

```
I can't think of a way to convert the values 'on the fly' (but probably there is :) ) so this way would go by creating an auxiliary list of dictionaries.
Later when you print, you can just directly print `scene["temporalCoverage"]['startDate']`

veroandreo:
Thanks @griembauer!! With your suggestion, it actually works! I realize now that the sorting was highly likely not working before... given that i did not convert clouds to float and so on... ops!

I also added an option `timeout` for download with a default value of 300 seconds, so it can be modified accordingly. This option was added to landsatxplore by you @griembauer :) Thanks again!

grass7/imagery/i.landsat/i.landsat.download/i.landsat.download.py
```diff
@@ -189,7 +210,7 @@ def main():
 
             try:
 
-                ee.download(scene_id=i, output_dir=outdir)
+                ee.download(scene_id=i, output_dir=outdir, timeout=options["timeout"])
```
neteler, line 213:
```suggestion
                ee.download(scene_id=i, output_dir=outdir, timeout=int(options["timeout"]))
```

neteler:
(small wish: it would be nice to get all this merged asap for easier testing on different systems)

ninsbl:
BTW: With latest G7.9 you should be able to test before merging like this:
```g.extension extension=i.landsat url=https://github.com/veroandreo/grass-addons branch=update-ilandsat-collection2 -o```
(note the _o-flag_, _branch_ and _url_))
No problem to merge newer modules like this for testing. But others that are used in production, should probably be tested before merging....

neteler:
If you refer to my comment: i processed a lot of Landsat 7 scenes yesterday using this PR. Successfully :)

neteler:
>  `g.extension extension=i.landsat url=https://github.com/veroandreo/grass-addons branch=update-ilandsat-collection2 -o`

(@ninsbl pls consider to add this nice new method to the g.extension documentation)

PR #455, opened 2021-02-28
i.landsat manuals: add timeout option in i.landsat.download and add i.landsat.qa + author

veroandreo:
This PR fixes i.landsat.download examples, adds a short explanation about new `timeout` option and includes the new `i.landsat.qa` + author to the main i.landsat manual. Shall we add further examples?

### DIFF
--- a/grass7/imagery/i.landsat/i.landsat.download/i.landsat.download.html
+++ b/grass7/imagery/i.landsat/i.landsat.download/i.landsat.download.html
@@ -56,7 +56,7 @@ and
 </table>
 
 <p>
-To connect to EarthExplorer both a <em>user</em> name and <em>password</em>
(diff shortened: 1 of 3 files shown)


neteler (approved):

ninsbl (approved):
Looking at it, maybe i.landsat.qa should also have the dataset opition, instead of two (collection and sensor), for more seamless user experience?

veroandreo:
> Looking at it, maybe i.landsat.qa should also have the dataset opition, instead of two (collection and sensor), for more seamless user experience?

Yes, that'd be more consistent, I believe. Would that be difficult to implement?

ninsbl:
> Yes, that'd be more consistent, I believe. Would that be difficult to implement?

Should be no big deal. Sensor and collection can be easily extracted from the dataset name...

PR #460, opened 2021-03-01
i.edge: add -n flag for empty input/output

metzm:
`i.edge` currently exits with fatal error if the input is empty. The new -n flag allows to create an empty output map without fatal error if the input map is empty.

### DIFF
--- a/grass7/imagery/i.edge/canny.c
+++ b/grass7/imagery/i.edge/canny.c
@@ -337,7 +337,7 @@ void performHysteresis(CELL * edges, CELL * magnitude, int low, int high,
 
 void thresholdEdges(CELL * edges, int rows, int cols)
 {
-    int i;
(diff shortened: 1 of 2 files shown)


neteler:
Tested successfully in NC location:

```
GRASS nc_spm_08_grass7/user1 >

# create empty map
r.mapcalc "lsat_pca.1 = null()"

# works as expected
i.edge input=lsat_pca.1 output=lsat_pca_1_edges
ERROR: Input map contains no data

# new flag also works as expected
i.edge input=lsat_pca.1 output=lsat_pca_1_edges -n
r.info -r lsat_pca_1_edges 
min=NULL
max=NULL

```

neteler (approved):
Thanks for the useful flag, helpful for `i.cutline` when used with tiling.

PR #468, opened 2021-03-04
CI: Add extra addon-specific dependencies

wenzeslaus:
For addons it is common to use scipy and R (and possibly other things).
This adds these to GitHub Action workflow so that the test of modules which use these
dependencies can run. It is also an intersting list of runtime dependencies which might be optional
for core and required for some addons.

These extra dependencies are installed only after core and addons compile which should
test for the compilating not requiring these extra dependencies.


stevenpawley (commented):
Hi Vaclav,

This looks like a really good improvement to get tests running for addons.

Are you hoping to maintain the requirements/extra-requirements for individual add-on's centrally, or should there be a mechanism where each add-on has a requirements.txt (if needed)? I think for Python, the case is more straightforward because one tends to use only a few packages, e.g. scikit-learn, scipy, pandas, although the r.richdem modules (which don't currently have tests) obviously need the richdem package. Addons that use R will probably depend on many different R packages though.

Steve

wenzeslaus:
> Are you hoping to maintain the requirements/extra-requirements for individual add-on's centrally, or should there be a mechanism where each add-on has a requirements.txt (if needed)?

At this point centrally (PR with tests will fail if they are missing, so that's fine from CI point of view), but some simple grabbing of `requirements.txt` for addons which have a testsuite dir would be doable. I agree that R is more tricky - just in this PR I had to increase R version to the latest although the directly installed packages were fine with a lower one. On the other hand, the latest version of R is what usually works and there is a GitHub Action for installing it.

For more complex solution, there is an ongoing discussion in [Trac 2895: Define dependencies for GRASS addons](https://trac.osgeo.org/grass/ticket/2895).

PR #420, opened 2021-01-30
refactor i.landsat.import to use gdal virtual file system and tarfile library

veroandreo:
This PR implements @ninsbl  suggestion in https://github.com/veroandreo/i.landsat/issues/16#issuecomment-742844484 to use GDAL virtual file system and the tarfile library instead of shutil. This removes the need to untar files and then remove them. The `unzip_dir` option is no longer required either.

PR #470, opened 2021-03-05
i.landsat.import: partially revert #420

veroandreo:
This PR partially reverts #420. 

It keeps tarfile instead of shutil, but removes the use of `/vsitar/` driver because it is extremely slow for reading files (to import 2 L8 scenes took more than 10 minutes). It didn't appear so slow when I first tested for #420. 

We could set the reading cache to a larger amount, but in any case, users of level 1 data would still need to uncompress the .tar.gz files to get the metadata file for `i.landsat.toar`. So, for now, I preferred to go back to uncompress from the beginning.

Any other suggestion is more than welcome! :)

### DIFF
--- a/grass7/imagery/i.landsat/i.landsat.import/i.landsat.import.py
+++ b/grass7/imagery/i.landsat/i.landsat.import/i.landsat.import.py
@@ -117,23 +117,22 @@ from grass.exceptions import CalledModuleError
 def _untar(inputdir, untardir):
 
     if not os.path.exists(inputdir):
-        gs.fatal(_("Input directory <{}> does not exist").format(inputdir))
+        gs.fatal(_("Directory <{}> does not exist").format(inputdir))
     if  not os.is_dir(inputdir):
         gs.fatal(_("<{}> is not a directory").format(inputdir))
-    elif not os.access(untardir, os.W_OK):
+    elif not os.access(inputdir, os.W_OK):
             gs.fatal(_("Directory <{}> is not writable.").format(inputdir))
 
     if untardir is None or untardir == "":
         untardir = inputdir
-
-        
     else:
         if not os.path.exists(untardir):
             gs.fatal(_("Directory <{}> does not exist").format(untardir))
         if not os.is_dir(untardir):
-           gs.fatal(_("<{}> is not a directory").format(inputdir))
+            gs.fatal(_("<{}> is not a directory").format(untardir))
         elif not os.access(untardir, os.W_OK):
-            gs.fatal(_("Directory <{}> is not writable.").format(inputdir))
+            gs.fatal(_("Directory <{}> is not writable.").format(untardir))
+
     if options["pattern_file"]:
         filter_f = "*" + options["pattern_file"] + "*.tar.gz"
     else:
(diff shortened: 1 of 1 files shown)


grass7/imagery/i.landsat/i.landsat.import/i.landsat.import.py
```diff
@@ -108,26 +114,30 @@
 from grass.exceptions import CalledModuleError
 
 
-def get_files_paths(inputdir):
+def _untar(inputdir, untardir):
 
     if not os.path.exists(inputdir):
         gs.fatal(_("Input directory <{}> does not exist").format(inputdir))
```
ninsbl, line 120:
```suggestion
        gs.fatal(_("Input directory <{}> does not exist").format(inputdir))
    if  not os.is_dir(inputdir):
        gs.fatal(_("<{}> is not a directory").format(inputdir))
    elif not os.access(untardir, os.W_OK):
            gs.fatal(_("Directory <{}> is not writable.").format(inputdir))
```

grass7/imagery/i.landsat/i.landsat.import/i.landsat.import.py
```diff
@@ -108,26 +114,30 @@
...
 
+    if untardir is None or untardir == "":
+        untardir = inputdir
+
+    if not os.path.exists(untardir):
+        gs.fatal(_("Directory <{}> does not exist").format(untardir))
+
```
ninsbl, line 127:
```suggestion
        
    else:
        if not os.path.exists(untardir):
            gs.fatal(_("Directory <{}> does not exist").format(untardir))
        if not os.is_dir(untardir):
           gs.fatal(_("<{}> is not a directory").format(inputdir))
        elif not os.access(untardir, os.W_OK):
            gs.fatal(_("Directory <{}> is not writable.").format(inputdir))
```

Not sure which checks `G_M_OPT_DIR` perform by default, but if you test for existence you should probably also check that untardir is a direcory and writable. It would probably be more elegant to place that check in a function.

ninsbl (approved):
I am a bit surprised that the virtual file system is so much slower. Maybe /vsitar/ unpacks the tar.gz file for every band...
Sorry for putting you on that path. Added two suggestions to check if input is a directory and writable...

neteler:
> I am a bit surprised that the virtual file system is so much slower. Maybe /vsitar/ unpacks the tar.gz file for every band...

While I am reluctant to ask but maybe @rouault has some insight here? I suspected the VSI_CACHE being to small but didn't really find any hints.

rouault:
The issue with a .tar.gz file is that it is the whole .tar that is gzipped compressed (not individual files independently, contrary to a .zip), so if you need to access a file toward the end of the archive you need to decompress all the content before it. There is a mechanism internally to /vsigzip/ (used by /vsitar on a .tar.gz) to create intermediate "snapshots" of the compressed stream for faster seeking, but that might be insufficient. The performance would probably be better by using VSIReadDirRecursive("/vsitar/....") and reading the files in the order they are listed.

neteler:
> The performance would probably be better by using VSIReadDirRecursive("/vsitar/....") and reading the files in the order they are listed.

Thanks for your feedback. But where would this change to be placed? Here or in GDAL?

PR #473, opened 2021-03-09
i.landsat.import: add resample option so user can select resample method

veroandreo:
Given that some artifacts appear especially when importing L7 with SLC stripping problem when using bilinear as resample method, this PR adds the option resample so users can select a preferred method according to their needs. Default answer is nearest as in r.import.


neteler (approved):
Looks good to me

ninsbl:
Should not QA bands allways be interpolated with nearest neigbors, if they are imported?

veroandreo:
> Should not QA bands allways be interpolated with nearest neigbors, if they are imported?

My tests didn't show issues using L8 data, but when you import L7 data with SLC problem, then some artifacts appear if resample=bilinear which affect the QA application later. @neteler that was the case, right?

neteler:
Yes exactly

PR #481, opened 2021-03-16
v.rast.bufferstats: support for raster with categories

lucadelu:
The change seems to be need in the case of rasters with categories name


ninsbl (approved):
Have not actively tested but looks good! Thanks for the improvement!

Issue #472, opened 2021-03-07
[Bug] Add Python GDAL bindings needed by v.in.pygbif to CI


wenzeslaus:
**Name of the addon**

v.in.pygbif

**Describe the bug**

GitHub Actions configuration misses GDAL Python bindings and thus v.in.pygbif fails.

```
Traceback (most recent call last):
  File "vector/v.in.pygbif/testsuite/test_pygbif_import.py", line 35, in test_poa_taxon_match
    self.assertModule(v_in_pygbif_match)
  File "etc/python/grass/gunittest/case.py", line 1419, in assertModule
    self.fail(self._formatMessage(msg, stdmsg))
AssertionError: Running <v.in.pygbif> module ended with non-zero return code (1)
Called: v.in_pygbif(output='gbif_poa1', taxa='Poa', basisofrecord='ALL', rank='species', flags='g', verbose=True)
See the following errors:
ERROR: Unable to load GDAL Python bindings (requires package 'python-gdal' or Python library GDAL to be installed).
```

**To Reproduce**

Explore e.g.: https://github.com/OSGeo/grass-addons/runs/2048721725 or any other recent run.

**Expected behavior**
Addons repo CI build environment in GitHub Actions (which is running tests) has all the (reasonable) dependencies.

**System description (please complete the following information):**

GitHub Actions Ubuntu 20.04 virtual machine.

**Additional context**
See also eccea34c4d268254498f09c12a8ea85ab668977a.

ninsbl:
Seems fixed in #488

PR #488, opened 2021-03-24
add GDAL and pygbif to extra_requirements

ninsbl:
Adress #472

Would be good to document how to handle dependencies on external libraries for AddOn tests.
Maybe something for the evtl. GSoC project on testing framework?

### DIFF
--- a/.github/workflows/build_grass.sh
+++ b/.github/workflows/build_grass.sh
@@ -15,6 +15,7 @@ then
 fi
 
 export INSTALL_PREFIX=$1
+export GDAL_VERSION=$(gdal-config --version)
 
 # GRASS GIS
 
--- a/.github/workflows/build_grass.sh
+++ b/.github/workflows/build_grass.sh
@@ -15,7 +15,6 @@ then
 fi
 
 export INSTALL_PREFIX=$1
-export GDAL_VERSION=$(gdal-config --version)
 
 # GRASS GIS
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,6 +87,7 @@ jobs:
 
     - name: Get extra Python dependencies
       run: |
+        export GDAL_VERSION=$(gdal-config --version)
         pip install -r grass-addons/.github/workflows/extra_requirements.txt
 
     - name: Set up R
--- a/.github/workflows/extra_requirements.txt
+++ b/.github/workflows/extra_requirements.txt
@@ -1,5 +1,5 @@
 scipy
 pandas
 scikit-learn
-GDAL==$(gdal-config --version)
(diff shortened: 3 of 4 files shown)


neteler:
> Adress #472
> 
> Would be good to document how to handle dependencies on external libraries for AddOn tests.

In mundialis, we use `requirements.txt` to document them (incl. optional version pinning), for a run of

`pip install -r requirements.txt`

Perhaps `g.extension` could simply pick up this?

> Maybe something for the evtl. GSoC project on testing framework?

Could be!

ninsbl:
> In mundialis, we use `requirements.txt` in the addon folder to document them (incl. optional version pinning), for a run of

Yes, I actualy took some inspiration from that.

This PR however, is about the CI system, where addons currently are compiled and not installed via g.extension.

Dependency handling in g.extension is probably a separate issue. I see that a simple `requirements.txt` in the AddOn dir might be the most AddOn dev friendly solution. So maybe we can unify dependency handling around that... Not sure... As a quick-fix for the current CI setup this is probably good enough. But of course not having to define dependencies several places (central CI pluss per addon) would be an advantage...

wenzeslaus (approved):

wenzeslaus:
> > Would be good to document how to handle dependencies on external libraries for AddOn tests.
> 
> In mundialis, we use `requirements.txt` in the addon folder to document them (incl. optional version pinning), for a run of
> 
> `pip install -r requirements.txt`
> 
> Perhaps `g.extension` could simply pick up this?

I agree, see also [Trac Ticket 2895: Define dependencies for GRASS addons](https://trac.osgeo.org/grass/ticket/2895), my  [comment 26](https://trac.osgeo.org/grass/ticket/2895#comment:26), but also my [comment 25](https://trac.osgeo.org/grass/ticket/2895#comment:25).

> > Maybe something for the evtl. GSoC project on testing framework?

I think there are three separate things, installation, testing framework, and tooling which runs the CI and uses the other two. The current testing framework is not a place where any of the dependency handling could live. It does not install anything. It just runs existing stuff. A flexible library to handle dependencies used by g.extension and some tool(s) which can be used in the CI is probably the way to go. +1 for GSoC, create an idea on the 2021 page.

> As a quick-fix for the current CI setup this is probably good enough.

Yes, I definitively introduced these as a quick fix assuming that 1) most dependencies are shared and the versions do not matter 2) only small portion of addons has additional unique dependencies (such as pygbif).

> But of course not having to define dependencies several places (central CI plus per addon) would be an advantage...

The ideal state perhaps is to test the dependencies too for which the CI would have to use a fresh environment for each addon. The current setup specifies well (and in one place) what environment is need to run all addons and tests test which has some value too.

PR #491, opened 2021-03-24
v.stream.order testsuite: update tests

neteler:
With this PR, the test passes (not sure if doing the right thing, though)


neteler:
Closing, maybe nonsense anyway...

PR #497, opened 2021-04-02
Python scripts: remove coding: utf-8 line

neteler:
Implements suggestion from https://github.com/OSGeo/grass/pull/1496#issuecomment-812190638


nilason:
Without changing the shebangs to python3, this might lead to complications.
If addons are supposed to work with python3 as minimum, then the shebangs can be changed along with removing the coding line.

As I noted in [ML](https://lists.osgeo.org/pipermail/grass-dev/2021-March/095054.html) the addons makes bumping python support more complicated as it is currently implemented (per major version). I didn't have the time for initiate discussion on RFC on python version support yet, but this is a major issue moving forward.

nilason:
> Without changing the shebangs to python3, this might lead to complications.
> If addons are supposed to work with python3 as minimum, then the shebangs can be changed along with removing the coding line.

On the other hand, I notice there are already .py files with `python` in shebang with no encoding line. As long as the python version used is 3 this shouldn't cause any issues. What I'm not sure about is whether addons for G7 are supported at all for python2 (and actually work).

neteler:
> On the other hand, I notice there are already .py files with `python` in shebang with no encoding line.

Yes, the encoding line was not used everywhere.

> As long as the python version used is 3 this shouldn't cause any issues. What I'm not sure about is whether addons for G7 are supported at all for python2 (and actually work).

Python 2 support is only needed for GRASS GIS 7.6 and older. Since that's slowly moving EOL (for Linux distros, see https://repology.org/project/grass/versions) and Python 2 is EOL anyway we may move on at some point.

grass7/imagery/i.landsat8.swlst/landsat8_mtl.py
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/python\<nl>\
```
nilason, line 1: Perhaps you can change this, while you're at it, to something that makes sense, like:
`#!/usr/bin/env python`

nilason (approved):
> > On the other hand, I notice there are already .py files with `python` in shebang with no encoding line.
> 
> Yes, the encoding line was not used everywhere.
> 
> > As long as the python version used is 3 this shouldn't cause any issues. What I'm not sure about is whether addons for G7 are supported at all for python2 (and actually work).
> 
> Python 2 support is only needed for GRASS GIS 7.6 and older. Since that's slowly moving EOL (for Linux distros, see https://repology.org/project/grass/versions) and Python 2 is EOL anyway we may move on at some point.

I'm good with this change, just wanted to bring the issue to attention.

grass7/imagery/i.landsat8.swlst/landsat8_mtl.py
neteler: Sure, done  (I found a few more which I have fixed as well in ee88869)

PR #509, opened 2021-04-08
g.citation: get year from file attributes

### LINKED ISSUE #489: [Bug] g.citation: Citation fails for all modules
**g.citation**

**Describe the bug**
g.citation fails for all modules with with the error message:
```The text does not contain date entry```

**To Reproduce**
```
g.extension extension=g.citation
g.citation -as
```

**Expected behavior**
g.citation should be able to provide citation for at least some modules...

 - Ubuntu 18.04
 - GRASS GIS 7.8.5

ninsbl:
Fix #489

The new solution is to take the year from the installation / creation of the module file. But since the URL i parsed properly, Modification time could (should) probably be fetched from git (with different routines for core modules and addons).

Some author names are not picked up properly. Maybe that schould be unified at some point in the manuals?

Also the citation does not look very comprehensive for e.g. bibtex. Maybe this should be used as a template: https://www.bibtex.com/e/entry-types/#misc ?

The module has doctests, sould I add a simple unittest never the less?

### DIFF
--- a/grass7/general/g.citation/g.citation.py
+++ b/grass7/general/g.citation/g.citation.py
@@ -185,6 +185,17 @@ def clean_line_item(text):
     return text
 
 
+def get_access_date_from_documentation_file_attributes(path):
+    """Extract year from file attributes of the documentation
+
+    >>> path = documentation_filename(name)
+    >>> get_year_from_documentation_file_attributes(path)
+    2011-10-14
+    """
+    date_installed = datetime.fromtimestamp(Path(path).stat().st_ctime).strftime("%Y-%m-%d")
+    return date_installed
+
+
 def get_year_from_documentation_file_attributes(path):
     """Extract year from file attributes of the documentation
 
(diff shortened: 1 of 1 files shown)


wenzeslaus (dismissed):
Accessed or Retrieved as suggested by @neteler in #489 would give more context to the year, but otherwise, I think the fix is the right solution at this point for addon modules. Release date would be better for core modules (I see only build date), but "retrieved" is still good enough.

I don't think Git would be helpful since you don't know what version is installed especially for addon modules, although there you could try to capture that with g.extension first to use it later on with g.citation.

wenzeslaus:
> Some author names are not picked up properly. Maybe that schould be unified at some point in the manuals?

Yes or, as a next step, turn it around by generating the man page record from another file after creating these files by g.citation (and still fixing the un-parsed manually).

> Also the citation does not look very comprehensive for e.g. bibtex. Maybe this should be used as a template: https://www.bibtex.com/e/entry-types/#misc ?

Whatever you find as a good example. As you could guess from the doctest, I initially tried software, but that didn't seem to be supported in many systems.

> The module has doctests, sould I add a simple unittest never the less?

Ideally, yes. Simplest would be to call the existing doctest as unittest since the function is tested by it already. There are examples of it in the code. Search for something like `testsuite/*doctest*.py`.

ninsbl:
Replaced by: #677

PR #510, opened 2021-04-12
i.variance: stepwise (and optional area weighted) resampling

NikosAlexandris:
Add optional area weighted resampling (`-w` flag) and (in this commit an always _forced_) stepwise resampling. The stepwise resampling is: reuse each time the temporary resampled map and not each and every time the original (fine resolution) raster map.  Maybe the step-wise resampling can be optional too.-

### Visual examples

Testing the module with a 4 m map of some infrared surface property.

1. Before (using always the original map)

test_variance_before_changes

2. Forcing stepwise resampling (reusing at each new step the previous lower resolution map)

test_recursive_variance_without_w

3. Forced step-wise and area weighted resampling
test_recursive_variance_with_w

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,21 +6,13 @@ pages](https://grass.osgeo.org/grass78/manuals/addons/)
 for an overview.
 
 ## Install AddOns
-## How to get write access here
 
 The simplest way to install GRASS GIS AddOns is to use the `g.extension`
 module which is part of any GRASS GIS installation:
-Read access is granted to the public, write access
-must be requested, see here for details:
-<https://trac.osgeo.org/grass/wiki/HowToContribute#WriteaccesstotheGRASSaddonsrepository>
 
 <https://grass.osgeo.org/grass-stable/manuals/g.extension.html>
-However, it is also possible to fork this repository, insert your AddOn or fix
-an existing one in a new branch and finally open
-a [pull request](https://help.github.com/en/articles/about-pull-requests).
 
 The same module can remove installed addons
-In either case, please read the submitting rules at the bottom of this page.
 
 ## This repository
 
@@ -34,6 +26,15 @@ git clone https://github.com/OSGeo/grass-addons.git grass_addons
 
 ### Write access
 
+Read access is granted to the public, write access
+must be requested, see here for details:
+<https://trac.osgeo.org/grass/wiki/HowToContribute#WriteaccesstotheGRASSaddonsrepository>
+
+However, it is also possible to fork this repository, insert your AddOn or fix
+an existing one in a new branch and finally open
+a [pull request](https://help.github.com/en/articles/about-pull-requests).
+
+In either case, please read the submitting rules at the bottom of this page.
 
 ### Compile AddOns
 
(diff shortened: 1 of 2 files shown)


NikosAlexandris:
Perhaps the step-wise resamping isn't really necessary, or else, might be depending on what a user wants to achieve.
I wanted to progressively show the 'loss' of information at each step of decreasing the resolution.

However, the area weighted average is I think likely wanted in most, if not all, use cases!?

mlennert:
> @mlennert would you have some time to have a look here?

Definitely not in the next weeks, as I'm traveling. However, this module was somewhat of a pet project coded quickly after a discussion on the approach, and I have no idea if anyone actually uses it, so please feel completely free to modify as you feel adequate.

mlennert:
You could maybe explain/document a bit the motivation behind the proposed changes.

NikosAlexandris:
> > @mlennert would you have some time to have a look here?
> 
> Definitely not in the next weeks, as I'm traveling. However, this module was somewhat of a pet project coded quickly after a discussion on the approach, and I have no idea if anyone actually uses it, so please feel completely free to modify as you feel adequate.

Thank you Moritz. This is quite of a useful tool!

NikosAlexandris:
> You could maybe explain/document a bit the motivation behind the proposed changes.

Of course. I think that at least the `-w` flag is required for most, if not all use-cases. Me too I hope to get back to it soon and update the module and the documentation.

mlennert:
Le 28 mai 2021 09:37:03 GMT+01:00, Nikos Alexandris ***@***.***> a écrit :
>> You could maybe explain/document a bit the motivation behind the proposed changes.
>
>Of course. I think that at least the `-w` flag is required for most, if not all use-cases. Me too I hope to get back to it soon and update the module and the documentation.
>

I just have to admit that I have no understanding of the impact of that approach on how to interpret the results. Any literature you can point to ?

NikosAlexandris:
> Le 28 mai 2021 09:37:03 GMT+01:00, Nikos Alexandris ***@***.***> a écrit :
> > You could maybe explain/document a bit the motivation behind the proposed changes. Of course. I think that at least the `-w` flag is required for most, if not all use-cases. Me too I hope to get back to it soon and update the module and the documentation.
> I just have to admit that I have no understanding of the impact of that approach on how to interpret the results. Any literature you can point to ?

Moritz, I'll try to explain it visually. But in any case, think of the current functionality and the ever increasing local variance as the spatial resolution increases. This is not logical. At some point there should be a break-point or a plateau and then finally a decrease of the variance, as the spatial resolution progressively increases. Simply put, you can't have local variance increasing continuously while resampling from finer to coarser pixels.  Isn't that so?

neteler:
@NikosAlexandris can you please rebase this PR to `grass8`?

wenzeslaus:
I see only change in the readme file in the Changes. Was this merged elsewhere or wrong branch force-pushed? I suggest closing this and opening a new PR.

NikosAlexandris:
> I see only change in the readme file in the Changes. Was this merged elsewhere or wrong branch force-pushed? I suggest closing this and opening a new PR.

This has been something I haven't found the time to push forward. I think the modifications I introduced are meaningful and useful. I agree, however, best to close this PR and open a new one. Best also to visually explain what's all about and have @mlennert review.

PR #512, opened 2021-04-13
i.sentinel.preproc: remove file() + black

lucadelu:


wenzeslaus (commented):
Both commits look good, although I would appreciate more info on what I should review.

I would suggest splitting the Black part from the rest of the fix (i.e., different PR, perhaps for the whole directory or module). Then you can ignore it or skip it more easily when you dig through the history.

You should also specify which version of Black you used (and of course any additional parameters if you used them). This will make it more clear when transitioning to different Black version or adding Black to CI.

griembauer (commented):
If I understand correctly, the two parts of this PR are 
a) python formatting with black
b) replacing `file()` with `open()` for reading the aeronet file

Part a) looks completely fine to me, just one comment for part b): Is closing a file after opening it with `open()` not necessary if it is only opened in read (=default) mode?

nilason:
> If I understand correctly, the two parts of this PR are
> a) python formatting with black
> b) replacing `file()` with `open()` for reading the aeronet file
> 
> Part a) looks completely fine to me, just one comment for part b): Is closing a file after opening it with `open()` not necessary if it is only opened in read (=default) mode?

Python docs, as far as I can tell, is not conclusive what happens in case of `for row in open(aeronet_file):`.  However, [it states](https://docs.python.org/3/tutorial/inputoutput.html#reading-and-writing-files): "*If you’re not using the with keyword, then you should call f.close() to close the file and immediately free up any system resources used by it.*"

Alternative solution could be something like:

```py
with open(aeronet_file, "r") as aeronet_file_fp:
    for row in aeronet_file_fp:
        # do something with row
```

lucadelu:
> Both commits look good, although I would appreciate more info on what I should review.
> 
More info using comments?

> I would suggest splitting the Black part from the rest of the fix (i.e., different PR, perhaps for the whole directory or module). Then you can ignore it or skip it more easily when you dig through the history.
> 

ok, good to know for next time

> You should also specify which version of Black you used (and of course any additional parameters if you used them). This will make it more clear when transitioning to different Black version or adding Black to CI.

black version is the one you suggested, 20.8b1

lucadelu:
> > If I understand correctly, the two parts of this PR are
> > a) python formatting with black
> > b) replacing `file()` with `open()` for reading the aeronet file
> > Part a) looks completely fine to me, just one comment for part b): Is closing a file after opening it with `open()` not necessary if it is only opened in read (=default) mode?
> 
> Python docs, as far as I can tell, is not conclusive what happens in case of `for row in open(aeronet_file):`. However, [it states](https://docs.python.org/3/tutorial/inputoutput.html#reading-and-writing-files): "_If you’re not using the with keyword, then you should call f.close() to close the file and immediately free up any system resources used by it._"
> 
> Alternative solution could be something like:
> 
> ```python
> with open(aeronet_file, "r") as aeronet_file_fp:
>     for row in aeronet_file_fp:
>         # do something with row
> ```

done!

griembauer (approved):

wenzeslaus:
> More info using comments?

In this case just filling the title and description of the PR would be sufficient.

> black version is the one you suggested, 20.8b1

Please add that to the (final) commit message (when merging) for future reference.

lucadelu:
@wenzeslaus just to be sure, is it correct that I just modify the comment like https://github.com/OSGeo/grass-addons/commit/9e9bbf502c4c3c56904d49813adb6982bda437bd ?

wenzeslaus:
> @wenzeslaus just to be sure, is it correct that I just modify the comment like [9e9bbf5](https://github.com/OSGeo/grass-addons/commit/9e9bbf502c4c3c56904d49813adb6982bda437bd) ?

Yes, modifying the second ("commit message body") field after hitting _Squash and merge_ is exactly what I meant for adding the Black version. I would just leave out the empty lines, but we don't have a set style for the body of the commit message. As you have likely noticed by now, the body usually requires manual editing of what GitHub creates automatically.

PR #520, opened 2021-04-29
i.sentinel.download: adapt API URL

griembauer:
The API Hub URL to download Sentinel-data from the Copernicus Open Access Hub has changed from https://scihub.copernicus.eu/apihub/ to https://apihub.copernicus.eu/apihub/ as of 28 April 2021. See also [announcement 1](https://scihub.copernicus.eu/news/News00865) and [announcement 2](https://scihub.copernicus.eu/news/News00865). Even though "the old URL will not be dismissed" ([announcement 1](https://scihub.copernicus.eu/news/News00865)) trying to list and download Sentinel data using the old URL yields an authentication error. Hence, this PR adapts the default URL. 

### known issues
With the adapted URL searching for data with the `-l` flag and downloading footprints works again. However, data download was so far **not successful** as all requested datasets supposedly lie in the long-term archive (even though they are only a few weeks old).  In [announcement 2](https://scihub.copernicus.eu/news/News00865) a 2-week migration period is mentioned, so hopefully this is the cause of the problem and it will be solved soon.
A quick fix for the meantime would be to return to the old URL https://scihub.copernicus.eu/dhus (which was initially adapted to the API-HUB url in https://github.com/OSGeo/grass-addons/pull/475), as it seems that querying and downloading from this URL works without any issues.
What do you recommend? 

(see also PR for sentinelsat: https://github.com/sentinelsat/sentinelsat/pull/445)

### DIFF
--- a/grass7/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
+++ b/grass7/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
@@ -480,7 +480,7 @@ def download_gcs(scene, output):
 
 
 class SentinelDownloader(object):
-    def __init__(self, user, password, api_url='https://scihub.copernicus.eu/apihub'):
+    def __init__(self, user, password, api_url='https://apihub.copernicus.eu/apihub'):
         self._apiname = api_url
         self._user = user
         self._password = password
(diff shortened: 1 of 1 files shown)


griembauer:
Update: Downloading using the new address https://apihub.copernicus.eu/apihub now seems to work.

neteler (approved):

griembauer:
It seems to depend on the scene itself, perhaps the migration is ongoing... For example downloading the following scene works using the new URL:
`i.sentinel.download settings=settings.txt output=some_dir uuid=da161b9f-d0c3-4fec-89c7-495a22ba9999`

I would suggest to merge now as the old URL does not work at all anymore and this new URL will be the new standard according to ESA. Hopefully these "long-term-archive" issues will be sorted out after the migration period...

Issue #515, opened 2021-04-26
[Bug] v.tin.to.rast : `NameError: name 'file' is not defined`


kikislater:
**Name of the addon**
v.tin.to.rast

**Describe the bug**
Plugin return a python error

```
/home/sylvain/.grass7/addons/scripts/v.tin.to.rast --verbose input=TIN output=TIN_rast
Traceback (most recent call last):
  File "/home/sylvain/.grass7/addons/scripts/v.tin.to.rast", line 188, in <module>
    main()
  File "/home/sylvain/.grass7/addons/scripts/v.tin.to.rast", line 74, in main
    nuldev = file(os.devnull, 'w')
NameError: name 'file' is not defined
Error in atexit._run_exitfuncs:
Traceback (most recent call last):
  File "/home/sylvain/.grass7/addons/scripts/v.tin.to.rast", line 65, in cleanup
    nuldev = file(os.devnull, 'w')
NameError: name 'file' is not defined
```

**To Reproduce**
1. Install v.tin.to.rast from g.extension
2. Run v.tin.to.rast
3. Input as previous output from v.triangle

**Expected behavior**
Easily run plugin

**System description (please complete the following information):**
 - Operating System: Manjaro Linux
 - Python 3.9.3

```
g.version -rge
version=7.8.5
date=2020
revision=4e3f393
build_date=2021-04-26
build_platform=x86_64-pc-linux-gnu
build_off_t_size=8
libgis_revision=2021-04-26T07:55:46+00:00
libgis_date=2021-04-26T12:00:00+04:00
proj=6.3.2
gdal=3.0.4
geos=3.8.1
sqlite=3.35.4
```

nilason:
> I suppose that the same change as this one is needed: ab83b84

Please see discussion at https://github.com/OSGeo/grass-addons/pull/512#issuecomment-819271930

kikislater:
It works thank you, PR created

PR #521, opened 2021-04-29
Use open() for opening files, Python 3 update

nilason:
Following #515 I located some more uses of `file()` to open files among grass7 addons. `file()` was removed with Python 3 and leads to fatal error. This replaces those remaining cases with `open()`.


neteler (approved):
Looks good to me, thanks

lucadelu (approved):

nilason:
Thanks!

PR #523, opened 2021-05-03
[DRAFT] first version of t.vi

lucadelu:
There is a problem with ParallelModuleQueue, it run the first MultiModule but after that it stop and it remain appended. Is there any suggestions?


metzm:
Have a look at other `t.*` modules how ParallelModuleQueue is used there.
Please rename `t.vi` to `t.rast.vi` for consistency.

grass7/temporal/t.vi/t.vi.py
```diff
@@ -0,0 +1,590 @@
...
+            )
+        else:
+            if numrast is None:
+                numrast = len(maps)
+                name = sp.get_name()
+            else:
+                if numrast != len(maps):
+                    grass.warning(
+                        _(
+                            "Number of raster in {} strds and {} strds are not the same.".format(
```
veroandreo, line 215:
```suggestion
                            "The number of raster maps in {} strds and {} strds differs.".format(
```

grass7/temporal/t.vi/t.vi.py
```diff
@@ -0,0 +1,590 @@
...
+        extent = rast[0].get_temporal_extent()
+        clean = []
+        if clouds:
+            maps = clouds.get_registered_maps_as_objects(mywhere, "start_time", None)
+            if len(maps) == 0:
+                gscript.warning(_("No clouds map for {}").format(ti[0]))
+            else:
+                cloud_exists = True
+                if len(maps) > 1:
+                    gscript.warning(_("To many clouds maps, only first will be used"))
```
veroandreo, line 453: IIUC, one single cloud map will be used to mask the full time series here, is that right? 

Maybe better warn that this is not a good idea and it might lead to unnecessary loss of valid data? Same for shadows below.

lucadelu: no, only the first map for the considered times (since we are looping for each times) is used. it is just a check that no more than one map of each input map is used. I think we can resolve this conversation, if it is fine for you. Maybe I can add a comment here

grass7/temporal/t.vi/t.vi.html
```diff
@@ -0,0 +1,25 @@
...
+This module calculate vegetation indices from several space time raster datasets,
+using <em>i.vi</em> module.<br>
+
+It can apply cloud and shadow masks stored into space time datasets (raster or vector)
+
+The output space time raster dataset can be a new dataset or an existing one, 
+in this case the new created maps will be add to the existing space time raster dataset.
+
+<h2>EXAMPLE</h2>
+
```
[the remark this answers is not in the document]
lucadelu: I can add an example with funny names, I would like to have a location/mapset for all the temporal examples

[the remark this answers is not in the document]
veroandreo: There's a mapset with MODIS LST data, but it's LST, so, not really useful here... 
But I meant moreover a real example to better understand the use cases you envision, even 2 or 3 landsat or sentinel scenes would be enough here. Should users create time series of cloud and/or cloud shadow masks? Or add cloud/cloud shadow as a band reference and then register those maps in the same time series with bands?

grass7/temporal/t.vi/t.vi.py
veroandreo: ah, ok, thanks for the explanation and sorry for my confusion. The message could then be: "Too many clouds maps for a single date/time, only the first one will be used". What do you think?

Still I wonder in which case would someone register many cloud maps for a single date/time. Maybe, when they try with different thresholds? or different flags? 

Please, add all these details to the manual.

lucadelu:
there was some problem with the rebase. going to close this and open new pull request

wenzeslaus:
Maybe names for the remotes in your local clone are different from the guide you use for rebase or merge and that's why it went wrong. Here is what I have:

```
$ git remote -v
fork	git@github.com:wenzeslaus/grass-addons.git (fetch)
fork	git@github.com:wenzeslaus/grass-addons.git (push)
upstream	https://github.com/OSGeo/grass-addons.git (fetch)
upstream	https://github.com/OSGeo/grass-addons.git (push)
```
 
So I update with:

```
git fetch upstream
git switch grass8
git rebase upstream/grass8
git switch -c add-cool-feature
```

Then you can check with `git log` (I have this set as alias called `hi` in `.gitconfig`):

```
git log --graph --pretty=format:"%Cred%h%Creset -%C(yellow)%d%Creset %s %Cgreen(%ar) %C(bold blue)<%an>%Creset" --abbrev-commit --max-count=10
```

The commit hashes should than be the same as on [the repo website](https://github.com/OSGeo/grass-addons/commits/grass8) (just the commit order is reversed in the two listings). Then you are ready to commit your changes.

You can also completely replace a remote branch with a new local branch using:

```
git push origin +new-branch:old-branch
```

Assuming that `origin` is your fork (see `git remote -v`). In my case (see above), I would replace `origin` with `fork` in the above command.

Issue #525, opened 2021-05-07
[Bug] i.fusion.hpf - change file encoding from us-ascii to utf-8


hellik:
**Name of the addon**
i.fusion.hpf 

it seems not all text/python files in the i.fusion.hpf source are utf-8 encoded

```
$ file --mime-encoding *
Makefile:                        us-ascii
README.md:                       utf-8
constants.py:                    utf-8
high_pass_filter.py:             us-ascii
i.fusion.hpf.html:               us-ascii
i.fusion.hpf.py:                 us-ascii
i_fusion_hpf_lsat7_hpf_rgb.png:  binary
i_fusion_hpf_lsat7_orig_rgb.png: binary
licenses:                        binary
test_high_pass_filter.py:        us-ascii
```

this causes a [g.extension crash](https://github.com/OSGeo/grass/pull/1496#issuecomment-812165105) in winGRASS

**Expected behavior**
no g.extension error

**solution**
encoding of files should changed to utf-8

hellik:
though not sure when looking at other addons, e.g.

```
$ file --mime-encoding *
Makefile:                 us-ascii
i.sentinel.download.html: us-ascii
i.sentinel.download.py:   us-ascii
```

source downloaded via zip file from github

neteler:
I have done a full repo check and grouped and counted files with identical encoding:

```
cd grass-addons/grass7/

# charset=us-ascii
for i in $(find . -type f) ; do file -i $i ; done | grep ascii | wc -l
2290

# charset=utf-8
for i in $(find . -type f) ; do file -i $i ; done | grep utf | wc -l
90
```

Likely it should be streamlined...

hellik:
lets have closer look into the files

```
D:\temp\testencoding\grass-addons-master\grass7\imagery\i.fusion.hpf>cat constants.py
#!/usr/bin/env python
#-*- coding:utf-8 -*-

"""
@author: Nikos Alexandris | October 2014
"""

"""
Constants for the HPFA Image Fusion Technique:
Kernel Size, Center Value, Modulation Factor (all depend on Resolution Ratio).

Sources:

- "Optimizing the High-Pass Filter Addition Technique for Image Fusion",
Ute G. Gangkofner, Pushkar S. Pradhan, and Derrold W. Holcomb (2008).

- â?oERDAS IMAGINE.â?? Accessed March 19, 2015. <==

```

these quotings are suspicious:

[...]

PR #526, opened 2021-05-08
g.extension/Windows: fix special characters

neteler:
This PR replaces numerous special characters g.extension on Windows is unable to deal with, as reported in https://github.com/OSGeo/grass/issues/1499.

The main problem appears to be ASCII vs UTF-8 encoding confusion.

### DIFF
--- a/grass7/gui/wxpython/wx.metadata/g.gui.metadata/g.gui.metadata.html
+++ b/grass7/gui/wxpython/wx.metadata/g.gui.metadata/g.gui.metadata.html
@@ -92,7 +92,7 @@ possible to use predifined templates with this editing mode.
 
 On the right side of editor there is a validator of metadata. Currently, 
 validator is able to validate two build-in profiles. 
-<!--For selection of profile “Load custom” is set by default the inspire profile.-->
+<!--For selection of profile "Load custom" is set by default the inspire profile.-->
 
 <h3>Updating GRASS native metadata </h3>
 
--- a/grass7/gui/wxpython/wx.mwprecip/mw3.py
+++ b/grass7/gui/wxpython/wx.mwprecip/mw3.py
@@ -1100,9 +1100,9 @@ class Computor():
     def computeAlphaK(self, freq, polarization):
         """@RECOMMENDATION ITU-R P.838-3
         Specific attenuation model for rain for use in prediction methods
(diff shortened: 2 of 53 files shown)


neteler:
It ideally fixes #525

ninsbl:
Will have a look into this. Meanwhile see also:
https://github.com/OSGeo/grass-addons/issues/525#issuecomment-835774706

neteler:
(fine for me to trash this PR if
we find a generic solution)

nilason:
I think ASCII-alise the files, in particular the .py files, would be a step backwards and only used as a last resort. Hopefully the suggested update https://github.com/OSGeo/grass-addons/issues/525#issuecomment-835774706 will deal with the most urgent problems on Windows.
There are a lot of ascii encoded python files both in grass and grass-addons, which may be batch converted with iconv.

hellik:
> I think ASCII-alise the files, in particular the .py files, would be a step backwards and only used as a last resort. Hopefully the suggested update [#525 (comment)](https://github.com/OSGeo/grass-addons/issues/525#issuecomment-835774706) will deal with the most urgent problems on Windows.
> There are a lot of ascii encoded python files both in grass and grass-addons, which may be batch converted with iconv.

In some Python addon scripts, there seems to be some mismatches between some (special) characters used and file encoding; this causes the crash in winGRASS g.extension.

Maybe an paragraph about to be more careful with file encoding of Python scripts and used (special) characters  should be added to the submitting guidelines?

In the example of i.fusion.hpf, it's just 1 quote character which isn't in line with the file encoding and causes the crash. 

Such things should be avoided while submitting an addon.

ninsbl (approved):
Regardles of possible changes in g.extensions, I tink this PR makes a lot of sense and increases consistency.

hellik:
> Regardles of possible changes in g.extensions, I tink this PR makes a lot of sense and increases consistency.

Agreed here

nilason:
> Regardles of possible changes in g.extensions, I tink this PR makes a lot of sense and increases consistency.

With regard to python files I disagree. This is just shovelling the dirt under the carpet. Proposed changes to g.extension actually make a whole lot of difference!

neteler:
I'll close this since a better solution is in place.

Issue #532, opened 2021-05-27
[Bug] i.segment.uspo: KeyError: 'radiues_start'


Raizencorp:
**Name of the addon**
i.segment.uspo

**Describe the bug**
Code is waiting "radiues_start" when radius_step is mentioned but raduies_start is not a valid argument

**To Reproduce**
i.segment.uspo group=yourgroup region=yourregion segmentation_method=mean_shift radius_step=1.0  raduis_start=1.0 radius_stop=10.0
run this line

**Expected behavior**
TraceBack just below

**Screenshots**
Traceback (most recent call last):
  File "/data/.grass7/addons/scripts/i.segment.uspo", line 1033, in <module>
    main()
  File "/data/.grass7/addons/scripts/i.segment.uspo", line 820, in main
    start = float(options['radiues_start'])
KeyError: 'radiues_start'

**System description (please complete the following information):**
 - OpenSuse Leap 15.2
 - GRASS GIS version 7.9.dev (2048e1016)

**Additional context**
Well i think it's a missfire from your keyboard, by just deleting the "e" in radiues_start line 820, this behavior should be corrected

veroandreo:
I fixed the typo, could you re-install the add-on and try again?

PR #533, opened 2021-05-27
i.segment.uspo: fix typo in option name as reported in #532

veroandreo:


veroandreo:
I merge this one directly since it's only a typo

PR #535, opened 2021-06-01
Modifiy the branch for archival of v6 addons

wenzeslaus:
* Removes the grass7 subdirectory.
* Renames the grass6 subdiretcory to src.
* Updates readme.
* Adds a license.
* Removes GitHub configuration.

See the individual commits to review the changes rather than all changes in Changes.


neteler:
Thanks for the initiative. The new branch https://github.com/wenzeslaus/grass-addons/tree/grass6 looks good to me.

The only changes (5 times) which I would suggest in `README.md` is

grass78 --> grass64

to reduce potential confusion.

wenzeslaus:
> README...grass78 --> grass64

Done.

wenzeslaus:
Does the rename of `grass6` subdirectory to `src` sound good in this case?

neteler:
> Does the rename of `grass6` subdirectory to `src` sound good in this case?

Since it stays in its branch `grass6` it sounds fine to me.

neteler:
May this be merged, @wenzeslaus? - then we can go ahead with the other addon branches which is a blocker for the GRASS GIS 8.0.0 release.

PR #540, opened 2021-06-08
i.sentinel.mask: adapt for non-SQlite vector databases

griembauer:
This PR adaps i.sentinel.mask such that it works in a mapset with a non-SQlite vector database connection. SQlite allows some simplifications of commands that are not possible in other systems (tested here with PostgreSQL):

- when running `v.db.addtable` to also add new columns, the datatype of the column must be indicated
- running `v.db.droptable` only works if the table exists in the database, it will otherwise throw an error. Because of this, a check is now implemented beforehand
- running `v.to.db` already creates an appropriate column so it is not necessary to create the column with `v.db.addcolumn` beforehand.

### DIFF
--- a/grass7/imagery/i.sentinel/i.sentinel.mask/i.sentinel.mask.py
+++ b/grass7/imagery/i.sentinel/i.sentinel.mask/i.sentinel.mask.py
@@ -498,12 +498,14 @@ def main ():
                 gscript.run_command('v.centroids',
                     input=tmp["shadow_temp_mask"],
                     output=tmp["centroid"], quiet=True)
-                gscript.run_command('v.db.droptable',
-                    map=tmp["centroid"],
-                    flags='f', quiet=True)
+                # drop table only if it exists
+                if len(gscript.parse_command("v.db.connect", flags="p", map=tmp["centroid"])) > 0:
+                    gscript.run_command('v.db.droptable',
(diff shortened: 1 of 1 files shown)


griembauer:
Update: Please ignore for now, more adaptions to i.sentinel.mask are to come (performance improvement)

ninsbl:
Hi, @griembauer 
When you are looking for speed improvements, maybe you could have a look at #138 for a significant speed improvement of cloud shaddow detection in cases where one has a significant amount clouds that slow vector processing. Basically, the idea is to keep things in raster all the way and use `r.region` (which takes seconds) + `r.stats` instead of `v.transform` + `v.overlap` which both can take minutes... With few clouds, the vector approach may still be faster though. Since you need the mask as raster dataset further down the processing chain, staying in the raster world is probably an advantage for the next processing steps too...
Unfortunately, I have not found the time to implement that myself. So pleasee feel free to pick from there, if relevant for you...

griembauer:
Hi @ninsbl, Thanks a lot for the link, I wasn't aware of this PR honestly, but of course it makes much more sense to take it from there! (indeed staying in the raster world as long as possible is probably the best way to go)

ninsbl:
Please see: #1002 

Is this PR still needed then?

PR #545, opened 2021-06-12
Add i.gabor

ocsmit:
This PR adds a new module (`i.gabor`) for Gabor filter bank creation/convolution with imagery data.


grass7/imagery/i.gabor/i.gabor.py
```diff
@@ -0,0 +1,204 @@
...
+# % description: Percentile threshold to extract
+# % answer: 0
+# %end
+
+import os
+import sys
+import numpy as np
+
+if "GISBASE" not in os.environ:
+    sys.stderr.write("You must be in GRASS GIS to run this program.\n")
```
HuidaeCho, line 98: Please i18nize messages using ``_()``.

HuidaeCho (changes requested):

HuidaeCho:
I cannot reproduce your result:
image

grass7/imagery/i.gabor/i.gabor.html
```diff
@@ -74,27 +74,27 @@ <h2>EXAMPLE</h2>
...
     <img src="i_gabor_imaginary.png" border=1><br>
     <p>Imaginary component of Gabor filter bank with 8 orientations.
 </center>
 
 <div class="code"><pre>
-i.gabor.py input="ortho_2001_t792_1m@user1" output="ortho_i_8o_41_5_q@user1"
-orientation='0, 22.5, 45, 67.5, 90, 112.5, 135, 157.5' size=41 wavelength=6
-threshold=90 -c -i -q
-</div></pre>
+i.gabor input=ortho_2001_t792_1m output=ortho_i_8o_41_5_q \
```
HuidaeCho, line 92: Isn't this the same command line as above?

ocsmit: Fixed now, should be reproducible.

grass7/imagery/i.gabor/i.gabor.html
```diff
@@ -79,9 +79,9 @@ <h2>EXAMPLE</h2>
 </center>
 
 <div class="code"><pre>
-i.gabor input=ortho_2001_t792_1m output=ortho_i_8o_41_5_q \
+i.gabor input=ortho_2001_t792_1m output=ortho_i_8o_41_5 \
```
HuidaeCho, line 82: Please make it completely reproducible including region settings.
```bash
g.region n= s= e= w= res=
```
It was the extent. I had to zoom into this specific area manually to be able to see the reproduced result.

image

HuidaeCho (changes requested):

HuidaeCho:
I remember you created a *bank* of filters (multiple rasters) using a basename. Now, it's only one output raster. I guess I was wrong before?

grass7/imagery/i.gabor/i.gabor.html
```diff
@@ -79,6 +79,7 @@ <h2>EXAMPLE</h2>
 </center>
 
 <div class="code"><pre>
+g.region n=220272 s=219585 w=638335 e=639313
```
HuidaeCho, line 82: Please include resolution as well. `res=1`?

HuidaeCho (changes requested):

ocsmit:
The examples create a combined raster output with the **-c** flag. Without the **-c** flag it creates individual rasters with the base name to detail the parameters of each raster.

HuidaeCho:
Got it. Please consider using the standard `G_OPT_R_BASENAME_OUTPUT` for creating multiple output rasters. You can add a separate option for the basename and use the output name by default if both need to be created (supported?). As is, it's not clear if the module can generate multiple rasters at once. I couldn't find relevant information from the manual either.

grass7/imagery/i.gabor/i.gabor.html
```diff
@@ -5,6 +5,9 @@ <h2>DESCRIPTION</h2>
 filter bank and raster map.
 The 2 dimensional Gabor filter is an orientation sensitive filter which mimics
 the cells within the human visual cortex.
+The module can either produce individual rasters for with the prefix of
```
HuidaeCho, line 8: `for with the prefix` => `with the prefix`

I believe using the basename as prefix and appending filter information would be better than the other way around.

`basename_windowsize_orientation_waelength_offset_aspect` would be better than `windowsize_orientation_wavelength_offset_aspect_basename` for organizing outputs.

HuidaeCho (changes requested):

grass7/imagery/i.gabor/i.gabor.html
ocsmit: That makes much more sense. ``windowsize_orientation_wavelength_offset_aspect_basename`` would get messy quickly.

HuidaeCho:
Maybe, it's just the color map, but the last image looks different from mine:
image

ocsmit:
Looks like my example was using 4 orientations, not 8. Updating output example. Other example out put for the combined raster is correct.

ninsbl (commented):
Thanks for the nice addition! I am looking forward to testing it in a project.

I added two minor comments to the manual. Addressing them could help users who do not know the Gabor filter in particular to easier understand relevance of the AddOn for their possible use cases.

In the "Notes" you may add that the module handles the entire input map in NumPy and thus in memory and that therefore memory can be a bottleneck for handling large raster maps at once...

But neither of my comments are essential...

grass7/imagery/i.gabor/i.gabor.html
```diff
@@ -0,0 +1,148 @@
...
+output.
+<p>
+Additionally, <em>i.gabor</em> offers statistical thresholding which will set
+values below a user-specified percentile (<b>threshold</b>) of the convolved
+image to 0 while keeping all values above the percentile the same.
+In conjunction with thresholding, a bitwise quantification (<b>-q</b>) method
+is implemented for the orientations specified by the user.
+
+<h2>NOTES</h2>
+As an orientation sensitive filter, the Gabor filter is unique in its
```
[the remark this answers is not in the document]
HuidaeCho: Absolutely agreed with @ninsbl. Like I mentioned earlier personally, explaining why this module is important is key. Otherwise, you're limiting the user base to only those who know what the Gabor filter is and why/how it is useful. As is, the manual explains the technical aspects of Gabor, but it's not clear how it can be useful beyond creating directional texture rasters as shown in the examples. I mean, now what to do with the outputs? Making sense of the outputs and how to use them would be a great addition to the manual.

HuidaeCho:
> Looks like my example was using 4 orientations, not 8. Updating output example. Other example out put for the combined raster is correct.

Still, your background color is black, but mine is purple. Your *color* is not reproducible.

ocsmit:
Thank you so much for your suggestions @ninsbl! I believe I addressed all that was brought up, but would certainly take more suggestions to improve my submission.

HuidaeCho:
The image looks good. One thing I would add is an example of post-analysis of your result. It doesn't have to be GRASS-centric, but, for example, how to use it for supervised classification in Python code using SciPy? That would make the module more attractive. Like I said, as is, no image-processing experts, I'd assume most GIS users, will know what to do now.

grass7/imagery/i.gabor/i.gabor.html
```diff
@@ -0,0 +1,166 @@
...
+classification, water body and river extraction, land use classification and
+change detection, and is especially powerful when used with Object Based Image
+Analysis (OBIA) and edge detection algorithms.
+
+
+<h2>NOTES</h2>
+
+<em>i.gabor</em> uses <a href="https://numpy.org/">NumPy</a> to handle the
+entire raster map in memory.
+This can result in a bottleneck when processing large raster maps.
```
mlennert, line 27: I would be really great if there was an option to not treat everything in memory. Even though RAM is increases quickly, so does image size and resolution and we've always tried to make GRASS GIS robust to this (although some of the other filter addons do not). Would some sort of streaming approach be feasible ?

ocsmit: I believe that would be very important to add, especially as an intention of  mine for this module was to use it with high resolution imagery. I believe it will be feasible with the GRASS RasterRow/Segement classes.

HuidaeCho:
@ocsmit Overall, the manual looks good and the `i.segment` usages will help a lot. A few minor comments: two figures have subfigures, but they are hard to see with blurry text. For example,
image
This is already zoomed it. It would be better to upload the same-size figures as the others and make them as subfigures with links to the full standalone images using HTML and CSS.
```html
<div>
  <div style="display:inline-block; width:50%;">
    <div style="text-align:center;">
      <a href="figure1.png"><img src="figure1.png" /></a>
    </div>
    <div style="text-align:center;">
      <a href="figure3.png"><img src="figure3.png" /></a>
    </div>
  </div>
  <div style="display:inline-block; width:50%;">
    <div style="text-align:center;">
      <a href="figure2.png"><img src="figure2.png" /></a>
    </div>
    <div style="text-align:center;">
      <a href="figure4.png"><img src="figure4.png" /></a>
    </div>
  </div>
</div>
```
Also, it would be useful as a technical manual if you linked to [your final report](https://here.isnew.info/gisc-4360k/i.gabor%20-%20A%20gabor%20filter%20module%20for%20GRASS%20GIS%20by%20Owen%20Smith.pdf).

Congrats on your first GRASS module!

HuidaeCho:
@ocsmit Oops! I think you merged the master branch into this PR. Please avoid that and use `git rebase upstream/master`. Looks like this last commit doesn't have any new changes? If so, `git reset --hard HEAD^; git push -f origin HEAD` to cancel this merge commit? Please check [Updating your PR from master](https://trac.osgeo.org/grass/wiki/HowToGit#UpdatingyourPRfrommaster). Are you planning to make more suggested changes?

ocsmit:
Didn't realize that had happened! I used the new option on the website to pull upstream but I had assumed it rebased and not merged. Will fix. 

When I have time I plan on implementing a more robust method for  raster map I/O so it is not so dependent on memory. At the moment I am not sure when I will get to it.

HuidaeCho:
> pull upstream

`git pull` is a shortcut for `git fetch; git merge` [git-pull](https://git-scm.com/docs/git-pull).

ocsmit:
@HuidaeCho Reverted accidental merge.

wenzeslaus:
> Please avoid that and use `git rebase upstream/master`. Looks like this last commit doesn't have any new changes?

Just FYI, now it is not a big deal whether you merge or rebase. We currently squash the commits for merge (we have that as the only enabled option on GitHub) which makes all commits including merges disappear. As we have tested, it is the only way how to keep the history clear. I use rebase by default, but I use merge when I anticipate conflicts because I need to resolve them only for the last state, not each commit.

mlennert:
> Didn't realize that had happened! I used the new option on the website to pull upstream but I had assumed it rebased and not merged. Will fix.
> 
> When I have time I plan on implementing a more robust method for raster map I/O so it is not so dependent on memory. At the moment I am not sure when I will get to it.

If all other issues are resolved, I would suggest to already merge module even without having solved the question of memory handling. I think the module is very useful as is, and merging it now does not keep you from modifying it later.

grass7/imagery/i.gabor/i.gabor.html
mlennert: @HuidaeCho This can be resolved now, or ?

ninsbl:
> If all other issues are resolved, I would suggest to already merge module even without having solved the question of memory handling. I think the module is very useful as is, and merging it now does not keep you from modifying it later.

Agreed with @mlennert here.

As for the memory issue: Would it be possible to implement the gabor filter as a function in [r.resamp.filter](https://github.com/OSGeo/grass/blob/master/raster/r.resamp.filter/main.c) or [r.neighbors](https://github.com/OSGeo/grass/pull/597/files)?

HuidaeCho:
> > Please avoid that and use `git rebase upstream/master`. Looks like this last commit doesn't have any new changes?
> 
> Just FYI, now it is not a big deal whether you merge or rebase. We currently squash the commits for merge (we have that as the only enabled option on GitHub) which makes all commits including merges disappear. As we have tested, it is the only way how to keep the history clear. I use rebase by default, but I use merge when I anticipate conflicts because I need to resolve them only for the last state, not each commit.

Good to know. Thanks.

HuidaeCho:
> > If all other issues are resolved, I would suggest to already merge module even without having solved the question of memory handling. I think the module is very useful as is, and merging it now does not keep you from modifying it later.
> 
> Agreed with @mlennert here.
> 
> As for the memory issue: Would it be possible to implement the gabor filter as a function in [r.resamp.filter](https://github.com/OSGeo/grass/blob/master/raster/r.resamp.filter/main.c) or [r.neighbors](https://github.com/OSGeo/grass/pull/597/files)?

r.resamp.filter might be a good candidate for integration, but Gabor filters are unique in that they are usually generated as a *bank* for multiple orientations. r.resamp.filter currently doesn't take additional parameters for individual filters.

grass7/imagery/i.gabor/i.gabor.html
HuidaeCho: > @HuidaeCho This can be resolved now, or ?

It is already resolved. `output=` is now prefix.

Let me add some final touches to the manual.

ocsmit:
> As for the memory issue: Would it be possible to implement the gabor filter as a function in [r.resamp.filter](https://github.com/OSGeo/grass/blob/master/raster/r.resamp.filter/main.c) or [r.neighbors](https://github.com/OSGeo/grass/pull/597/files)?

I had looked at that previously and was hesitant to add it as function for either of those  as the gabor filter bank takes orientation (multiple), aspect, wavelength (multiple), offset, etc.. to construct each kernel in the bank. I believed this would add too many additional options for both modules for just one kernel and it would be simpler to implement on its own. However as [r.neighbors](https://github.com/OSGeo/grass/pull/597/files) allows for weights to be input by the user it could be possible to take each individual gabor kernel and use them as inputs for r.neighbors. That would remove the need to read raster maps into NumPy arrays for processing, as well as the SciPy dependency for FFT convolution. 

if r.neighbors were to be used and a combined raster is being created (`-c`), then it might be tricky to create this file without writing the individual raster maps for each kernel in the bank to disk.

ninsbl:
> if r.neighbors were to be used and a combined raster is being created (`-c`), then it might be tricky to create this file without writing the individual raster maps for each kernel in the bank to disk.

Maybe you could make it an additional, kind of "lowmem" option in `i.gabor` (version 2.0)?
Create different weight files in Python, send them to `r.neighbors` and maybe run one r.neighbors-process per direction in parallel using the [ParallelModuleQueue](https://grass.osgeo.org/grass79/manuals/libpython/pygrass.modules.interface.html?highlight=parallel%20module%20qu#pygrass.modules.interface.module.ParallelModuleQueue). Resulting raster maps could then be combined to one with e.g. `r.series`? This results likely in more IO (if c-flag is set), maybe slower (maybe not)... Maybe worth a try?

ocsmit:
I think that is a great idea and certainly one that is worth a try.

grass7/imagery/i.gabor/i.gabor.py
```diff
@@ -0,0 +1,205 @@
...
+# COPYRIGHT:    (C) 2021 by Owen Smith and the GRASS Development Team
+#
+#               This program is free software under the GNU General
+#               Public License (>=v2). Read the file COPYING that
+#               comes with GRASS for details.
+#
+############################################################################
+
+# %module
+# % description: Creates Gabor filter bank for an n-dimensional image
```
HuidaeCho, line 21: @ocsmit `n-dimensional` Maybe, it's a typo? Can it really handle an arbitrary number of dimensions? There is only `gabor2d()` in the code. Did you mean `2-dimensional` or can we use it for 3D voxel segmentation, for example? If not, I would change it to `2-dimensional`.

ocsmit: Good catch,  meant `2-dimensional`.

ocsmit (commented):
Whoops...

HuidaeCho:
Since all the three reviewers (@ninsbl, @mlennert, @HuidaeCho) approved the merge of this module, I take the liberty of merging it now.

ocsmit:
Thank you @ninsbl, @mlennert, and @HuidaeCho  for all of your help!

PR #549, opened 2021-06-15
i.modis.download: more reasonable error handling

neteler:
- catch more (user) errors
- more messages improved
- user example added with .netrc file
- fix broken URL to i.modis.qc
- remove obsolete GRASS GIS version test (hence, it also works with GRASS GIS 8)

### DIFF
--- a/grass7/imagery/i.modis/i.modis.download/i.modis.download.py
+++ b/grass7/imagery/i.modis/i.modis.download/i.modis.download.py
@@ -225,12 +225,6 @@ def main():
                                 "close this GRASS GIS session"))
             if check_folder(path):
                 fold = path
-    # check the version
-    version = grass.core.version()
-    # this is would be set automatically
-    if version['version'].find('7.') == -1:
-        grass.fatal(_('GRASS GIS version 7 required'))
-        return 0
     # the product
     products = options['product'].split(',')
     # first date and delta
--- a/grass7/imagery/i.modis/i.modis.download/i.modis.download.py
+++ b/grass7/imagery/i.modis/i.modis.download/i.modis.download.py
@@ -261,7 +261,7 @@ def main():
         modisOgg.connect()
         if modisOgg.nconnection <= 20:
             # download tha tiles
-            grass.message(_("Downloading MODIS product <%s>..." % produ))
+            grass.message(_("Downloading MODIS product <%s> (%s)..." % (produ, prod['prod'])))
             modisOgg.downloadsAllDay()
             filesize = int(os.path.getsize(modisOgg.filelist.name))
             if flags['g'] and filesize != 0:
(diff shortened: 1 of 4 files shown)


grass7/imagery/i.modis/i.modis.download/i.modis.download.py
```diff
@@ -195,6 +199,8 @@ def main():
                           "the username and password"))
     # set username, password and folder by file
     else:
+        if os.path.isdir(options['settings']):
```
veroandreo, line 202: I'm not an expert, but I wonder if `isdir()` will indeed test if a path is a file? I also found that there's `isfile()` (https://docs.python.org/3/library/os.path.html). I think isdir() will always be false, either the file in the path exists or not, at least that's what my toy tests show here.

grass7/imagery/i.modis/i.modis.download/i.modis.download.py
```diff
@@ -255,7 +261,7 @@ def main():
         modisOgg.connect()
         if modisOgg.nconnection <= 20:
             # download tha tiles
-            grass.message(_("Downloading MODIS product <%s>..." % produ))
+            grass.message(_("Downloading MODIS product <%s> (%s)..." % (produ, prod['prod'])))
```
veroandreo, line 264: general question: what is the difference between `%s % ()` and the .format() way?

lucadelu: I think format should be preferable. Reading [this](https://stackoverflow.com/questions/5082452/string-formatting-vs-format-vs-f-string-literal) format seems slower...

grass7/imagery/i.modis/i.modis.download/i.modis.download.py
lucadelu: I agree

anikaweinmann, line 202:
```suggestion
        if not os.path.isfile(options['settings']):
```

lucadelu: I committed the code but I don't see it in the code....

grass7/imagery/i.modis/i.modis.download/i.modis.download.py
anikaweinmann, line 264:
```suggestion
            grass.message(_("Downloading MODIS product <{}> ({})...".format(produ, prod['prod'])))
```

neteler: ok, changed by Anika :)

grass7/imagery/i.modis/i.modis.download/i.modis.download.py
neteler: Fixed by Anika

veroandreo (approved):

PR #554, opened 2021-06-16
Apply black -l 80

HuidaeCho:


wenzeslaus:
> `black -l 80` ... `black -l 79`

But see `line-length = 88` at https://github.com/OSGeo/grass/blob/master/pyproject.toml#L2

HuidaeCho:
Is that automatically applied so I don't have to do it myself?

Personally, I didn't like the default 88 line length. Why don't we stick to [PEP 8](https://www.python.org/dev/peps/pep-0008/) if PEP compliance is important? 79 line length is also default for vi (this time without m to save typing ;-) for gq and gw.

wenzeslaus:
> Is that automatically applied so I don't have to do it myself?

When you run Black, yes. In the PR, no. It could be, e.g., Mega-Linter can do that (although I think there is currently actually some problem with not really working well for forks). I'm just afraid it would cause more confusion for the beginners when the code on their branch is changed by some bot. Locally, you can probably do a pre-commit hook (here is [an example using vim](https://ljvmiranda921.github.io/notebook/2018/06/21/precommits-using-black-and-flake8/)).

> Personally, I didn't like the default 88 line length. Why don't we stick to [PEP 8](https://www.python.org/dev/peps/pep-0008/) if PEP compliance is important? 79 line length is also default for vi (this time without m to save typing ;-) for gq and gw.

[Black documentation](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length) explains: _You probably noticed the peculiar default line length. Black defaults to 88 characters per line, which happens to be 10% over 80. This number was found to produce significantly shorter files than sticking with 80 (the most popular), or even 79 (used by the standard library). In general, [90-ish seems like the wise choice](https://www.youtube.com/watch?v=wf-BqAjZb8M&t=260s)._

I didn't go for the Bugbear’s B950 Flake8 warning additionally recommended by Black, because it seemed unnecessary complex.

Another things to consider is that Python, unlike C, does not use names like `strcmp`, instead the names are usually longer and for new code often readable full words, so the classic 72/79/80 is little harder to fit into when you want readable variable names, so 88 is little more than 79, but not radical.

It seems that Linux kernel [recently switched](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=bdc48fa11e46f867ea4d75fa59ee87a7f48be144) from 80 to 100 (for maximum line length; I don't know how exactly this applies to the code).

PR #559, opened 2021-06-22
g.projpicker: Add error message if projpicker not found on system.

ocsmit:
Really minor, but potentially very helpful as long as projpicker is not a part of core repository.


ninsbl (approved):
Looks good.

PR #560, opened 2021-06-22
Reapply black without any options

HuidaeCho:
Personally, I think this `black` formatting breaks *my* semantic indentation and even violates [PEP 8](https://www.python.org/dev/peps/pep-0008/). Also, it's not compatible with vim's `gq` and `gw` with the default line width. https://lists.osgeo.org/pipermail/grass-dev/2021-June/095226.html

### DIFF
--- a/grass7/general/g.projpicker/g.projpicker.py
+++ b/grass7/general/g.projpicker/g.projpicker.py
@@ -105,7 +105,9 @@ def main():
     try:
         import projpicker as ppik
     except ImportError:
-        grass.fatal(_("ProjPicker not installed. Use 'pip install projpicker'"))
+        grass.fatal(
+            _("ProjPicker not installed. Use 'pip install projpicker'")
+        )
 
     coords = options["coordinates"]
     operator = options["operator"]
@@ -123,7 +125,10 @@ def main():
     single = flags["1"]
     start_gui = flags["g"]
 
-    if bbox_map and grass.parse_command("g.proj", flags="g")["unit"] != "degree":
+    if (
+        bbox_map
+        and grass.parse_command("g.proj", flags="g")["unit"] != "degree"
+    ):
         grass.fatal(_("Cannot create vector in degree in a non-degree mapset"))
 
     # ppik.projpicker() appends input file contents to geometries from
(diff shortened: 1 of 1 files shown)


wenzeslaus:
Since this breaks CI, please, either revert this PR or introduce a separate settings for this code after opening a discussion on mailing list about exceptions for individual addons.

I hope you forgive me that don't want to comment on it further here as this is not the a place to discuss the style. The place to do that was https://github.com/OSGeo/grass/issues/543 as also stated in https://github.com/OSGeo/grass/pull/766#issuecomment-731637722 and also on grass-dev in [February-April](https://lists.osgeo.org/pipermail/grass-dev/2021-February/094993.html) and [June 1](https://lists.osgeo.org/pipermail/grass-dev/2021-June/095199.html) threads.

HuidaeCho:
> Since this breaks CI

Does this PR break CI? Interesting. I thought CI just does `black` with no `--line-length`? Let me check... So should I use just `black` or add any arguments? OK, I was aliasing `black` to `black -l 79`, my bad. Does it mean we're forced to apply `black` with no arguments to not break CI?

> I hope you forgive me that don't want to comment on it further here

No need to comment on this here. I already expressed my opinion on this topic [here](https://lists.osgeo.org/pipermail/grass-dev/2021-June/095221.html). Just surprised that the `black` formatting of this script was changed to something else (without `-l 79`? I tried to figure that out).

HuidaeCho:
https://github.com/OSGeo/grass-addons/actions/runs/964349314

wenzeslaus:
> I thought CI just does `black` with no `--line-length`? Let me check... So should I use just `black` or add any arguments?

There is a configuration file Black is using (it is using generic pyproject.toml), so yes, using no arguments is correct. It reads config from the config file. Most of these linting and formatting tools behave in the way that command line arguments override config file values which is what happened in you case. The change in g.projpicker files was caused by me applying a consistent Black with the settings from core in the whole addons repo.

> OK, I was aliasing `black` to `black -l 79`, my bad. Does it mean we're forced to apply `black` with no arguments to not break CI?

Each project using Black would generally use either the Black defaults or have a config file, so you should run `black` without arguments every time. Unrelated to GRASS, if you want something else for your own projects, the right way is to create a config file. You can setup another alias (`black79`, ha ha!) to format your projects without a config file, but that makes it more difficult to share with other people.

Issue #484, opened 2021-03-17
r.basin - raster resolution ISSUE


IreneM21:
I’ve tried running r.basin using an elevation raster map (in lat/long coordinates) that I had to reproject first in order to apply the tool. As a result of the reprojection, the cell dimension is not of an integer type anymore. 
When I try to run the function it stops before ending and I get the following error message:

```
> ERROR: Region resolution and raster map <my_basin_dem_30m_strahler> resolution differs. Run 'g.region raster=my_basin_dem_30m_strahler' to set proper region resolution.
> ------------------------------
> An ERROR occurred running r.basin
> Please check for error messages above or try with another pairs of outlet coordinates
```

When I use a raster map with integer cell dimension (e.g. 30 meters) everything works fine so I think that the issue is related to the raster resolution.

Is there any way to fix this problem? It would be great if users could apply the function to reprojected raster maps as well.

To reproduce:
[dem_30m.tif.zip](https://github.com/OSGeo/grass-addons/files/6159087/dem_30m.tif.zip)
g.region raster=dem_30m -p 
r.basin --o map=dem_30m prefix=my_basin coord=442190,5075482 dir=/tmp/my_basin threshold=1000

hellik:
have a look at the gdalinfo output and the pixel size:

```

Driver: GTiff/GeoTIFF
Files: dem_30m.tif
       dem_30m.tif.aux.xml
Size is 3601, 3633
Coordinate System is:
PROJCRS["WGS 84 / UTM zone 32N",
    BASEGEOGCRS["WGS 84",
        DATUM["World Geodetic System 1984",
            ELLIPSOID["WGS 84",6378137,298.257223563,
                LENGTHUNIT["metre",1]]],
        PRIMEM["Greenwich",0,
            ANGLEUNIT["degree",0.0174532925199433]],
        ID["EPSG",4326]],
    CONVERSION["UTM zone 32N",
        METHOD["Transverse Mercator",
            ID["EPSG",9807]],
        PARAMETER["Latitude of natural origin",0,
            ANGLEUNIT["degree",0.0174532925199433],
            ID["EPSG",8801]],
        PARAMETER["Longitude of natural origin",9,
            ANGLEUNIT["degree",0.0174532925199433],
            ID["EPSG",8802]],
        PARAMETER["Scale factor at natural origin",0.9996,
            SCALEUNIT["unity",1],
            ID["EPSG",8805]],
        PARAMETER["False easting",500000,
[...]

PR #567, opened 2021-06-25
r.stream.*: Avoid comparing floating point numbers using not equal

HuidaeCho:
Attempt to address #484.

### DIFF
--- a/grass7/raster/r.stream.basins/io.c
+++ b/grass7/raster/r.stream.basins/io.c
@@ -75,8 +75,8 @@ int ram_read_map(MAP * map, char *input_map_name, int check_res,
     G_get_window(&this_window);
     Rast_get_cellhd(input_map_name, mapset, &cellhd);
     if (check_res)
-	if (fabs(this_window.ew_res - cellhd.ew_res) >= 1e-10 ||
-	    fabs(this_window.ns_res - cellhd.ns_res) >= 1e-10)
+	if (fabs(this_window.ew_res - cellhd.ew_res) >= GRASS_EPSILON ||
+	    fabs(this_window.ns_res - cellhd.ns_res) >= GRASS_EPSILON)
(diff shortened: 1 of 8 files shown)


grass7/raster/r.stream.basins/io.c
```diff
@@ -75,8 +75,8 @@ int ram_read_map(MAP * map, char *input_map_name, int check_res,
     G_get_window(&this_window);
     Rast_get_cellhd(input_map_name, mapset, &cellhd);
     if (check_res)
-	if (this_window.ew_res != cellhd.ew_res ||
-	    this_window.ns_res != cellhd.ns_res)
+	if (fabs(this_window.ew_res - cellhd.ew_res) >= 1e-10 ||
+	    fabs(this_window.ns_res - cellhd.ns_res) >= 1e-10)
```
marisn, line 79: Use GRASS_EPSILON.

marisn (approved):
Change explicit value to GRASS_EPSILON in all places and then it is ready for merging.

grass7/raster/r.stream.basins/io.c
HuidaeCho: `GRASS_EPSILON` is `1e-15` and the difference was greater than that (`-2.2311e-12` EW and `9.45022e-13` NS) *yesterday*. Now, I restarted in the same mapset for testing, the error is `0`. Interesting.. there were no changes in the mapset. All I did is exited and restarted the same mapset. In short, `GRASS_EPSILON` would not have worked yesterday, but now it works because of an unknown reason. Let me try it from scratch.

HuidaeCho:
OK, I think #484 is a problem in r.basin, not this floating point comparison (was only a workaround). Anyway, this PR still makes sense.

HuidaeCho:
#484 is fixed in #568. I'll merge this PR as well.

PR #568, opened 2021-06-25
r.basin: g.region using raster to set the correct resolution

### LINKED ISSUE #484: r.basin - raster resolution ISSUE

HuidaeCho:
Resolve #484 by using raster, not the vector version of the same raster, to set the correct resolution.

### DIFF
--- a/grass7/raster/r.basin/r.basin.py
+++ b/grass7/raster/r.basin/r.basin.py
@@ -465,9 +465,7 @@ def main():
         # Rectangle containing basin
         basin_east = baricenter_slope_baricenter[33]
         basin_north = baricenter_slope_baricenter[34]
-        info_region_basin = grass.read_command(
-            "g.region", vect=options["prefix"] + "_" + mapname[0] + "_basin", flags="m"
-        )
+        info_region_basin = grass.read_command("g.region", rast=r_basin, flags="m")
 
         grass.message("g.region done")
         dict_region_basin = dict(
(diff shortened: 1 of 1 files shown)


grass7/raster/r.basin/r.basin.py
```diff
@@ -465,9 +465,7 @@ def main():
         # Rectangle containing basin
         basin_east = baricenter_slope_baricenter[33]
         basin_north = baricenter_slope_baricenter[34]
-        info_region_basin = grass.read_command(
-            "g.region", vect=options["prefix"] + "_" + mapname[0] + "_basin", flags="m"
-        )
+        info_region_basin = grass.read_command("g.region", rast=r_basin, flags="m")
```
neteler, line 468:
```suggestion
        info_region_basin = grass.read_command("g.region", raster=r_basin, flags="m")
```

(complete parameter names are preferred)

Issue #528, opened 2021-05-14
Restructure repository for version 8


wenzeslaus:
The grass-addons repository will be restructured to accommodate better a new major version of GRASS GIS. To put it simply, we don't want to add another subdirectory grass8 next to existing grass6 and grass7 directories and we are taking the opportunity of upcoming release of 8.0.0 to change to a better system, more fitting with having multiple past major versions and with using Git.

This topic was discussed previously and most recently at [PSC meeting on May 11](https://trac.osgeo.org/grass/wiki/PSC/Minutes/PSC_Meeting_20210511). Please provide feedback for the plan below and/or volunteer. Help with identification of what will break is especially welcome. Complete reworkings of the plan are possible.

Goals:

- Use branches for different and old versions as in the core repository, not subdirectories.
- Make the new structure future-proof, i.e., such that when v9 comes, it is clear what to do.
- Keep addons working for the latest release (whatever is the latest patch version) of v7.

This issue is to track the main transition, but individual issue can be created to track specific issues both here and in the core grass repository as needed. We hope that more people will help with this. So far @landam plans to update the addon compilation for Windows and @wenzeslaus plans to do the changes in the branching and GitHub Actions updates in the grass-addons repository. Please add yourself and/or TODO items here as needed by editing this issue description.

TODOs:

- [x] Create branch (from master branch) for contents of grass6 directory (named `grass6`, `grass_v6`?). Restructure the contents as needed. Delete the directory on master branch. @wenzeslaus #535
- [x] Apply Black to the grass7 code in master branch so that it is similar in v7 and v8 to make backporting easier. #537
- [x] Create new branch (from master branch) for v7 (named `grass8`? `main`?). @wenzeslaus
- [x] Move contents of grass7 directory to the top level or rename it to src in the branch for (new) v7 and v8. @wenzeslaus #571
- [x] Make changes in the compilation and checks of addons in GitHub Actions. @wenzeslaus #571
- [x] Write to grass-dev about the change. @wenzeslaus (see [grass-dev](https://lists.osgeo.org/pipermail/grass-dev/2021-June/095257.html))
- [x] Make grass8 the default branch, update to the current master branch, make the master branch read-only. @wenzeslaus
- [ ] Inform contributors in PRs.
- [ ] Update PRs if needed (for stale but desired PRs).
- [x] Update the contributing instructions for v7. @wenzeslaus 
- [x] Update the contributing instructions for v8. @wenzeslaus #636
[...]

PR #571, opened 2021-06-28
Change to new v7 and v8 structure

wenzeslaus:
- Rename grass7 dir to src, delete grass6
- Update CI to src dir
- Add license file

This PR is against the new grass7 branch which would replace the master branch for any updates to v7 series.

The new grass8 for v8 series can be then created from this updated version.

See also #528.


neteler (approved):
This massive restructuring looks good to me and will help to improve the "navigation" in the different addon versions.

PR #576, opened 2021-07-05
r.suitability.regions: New addon for suitability scores

ecodiv:
A new addon to locate regions of contiguous raster cells with high suitability scores, based on a suitability raster layer as input.


ecodiv:
Hi @wenzeslaus, would you mind reviewing the code? Otherwise, I can merge the code myself, that is fine too. Not sure though what option I should use, squash and merge or rebase and merge (not sure what the difference is).

src/raster/r.suitability.regions/r.suitability.regions.py
```diff
@@ -0,0 +1,410 @@
...
+    focalstat = options["focalstat"]
+    minsuit = float(options["minsuit"])
+    minfrag = float(options["minfrag"])
+    maxgap = float(options["maxgap"])
+    suitreg = options["suitreg"]
+
+    # Flags
+    flag_c = flags["c"]
+    if flag_c:
+        c = "c"
```
wenzeslaus, line 198: `c` should be something like `neighbor_flags`. `c = "c"` is hard to read.

src/raster/r.suitability.regions/r.suitability.regions.py
```diff
@@ -0,0 +1,410 @@
...
+            gs.run_command("g.remove", flags="f", type="raster", name=rast, quiet=True)
+
+
+def tmpname(prefix):
+    """Generate a tmp name which contains prefix
+    Store the name in the global list.
+    Use only for raster maps.
+    """
+    tmpf = prefix + str(uuid.uuid4())
+    tmpf = tmpf.replace("-", "_")
```
wenzeslaus, line 167: See new functions in gs (`grass/script/utils.py`) for this. If you look for v7 compatibility, import in `try` and use copy-pasted (possibly simplified) version in except. See e.g.:

https://github.com/OSGeo/grass-addons/blob/grass7/src/misc/m.csv.clean/m.csv.clean.py#L90

src/raster/r.suitability.regions/r.suitability.regions.py
```diff
@@ -0,0 +1,410 @@
...
+        d = "d"
+    else:
+        d = ""
+    flag_z = flags["z"]
+    flag_k = flags["k"]
+    flag_f = flags["f"]
+    flag_a = flags["a"]
+
+    # Temporary layer names
+    tmp01 = tmpname("tmp01")
```
wenzeslaus, line 212: Don't just number your variables. Name them.

src/raster/r.suitability.regions/r.suitability.regions.py
```diff
@@ -0,0 +1,410 @@
...
+        )
+        gs.run_command(
+            "r.series", input=[suitrast, tmp02a], method="maximum", output=tmp02
+        )
+    elif radius > 1:
+        gs.message("Computing neighborhood statistic")
+        try:
+            float(absminsuit)
+        except:
+            gs.error("absminsuit must by numeric or left empty")
```
wenzeslaus, line 244: This does not terminate the process. Is that intentional? Use `fatal` to terminate.

src/raster/r.suitability.regions/r.suitability.regions.py
```diff
@@ -0,0 +1,410 @@
...
+            size=radius,
+        )
+        gs.run_command(
+            "r.series", input=[suitrast, tmp02a], method="maximum", output=tmp02
+        )
+    elif radius > 1:
+        gs.message("Computing neighborhood statistic")
+        try:
+            float(absminsuit)
+        except:
```
wenzeslaus, line 243: Please, no bare except. This is TypeError I presume.

src/raster/r.suitability.regions/r.suitability.regions.py
```diff
@@ -0,0 +1,410 @@
...
+    tmp04 = tmpname("tmp04")
+    tmp05 = tmpname("tmp05")
+    tmp06 = tmpname("tmp06")
+    tmp07a = tmpname("tmp07a")
+    tmp07 = tmpname("tmp07")
+    tmp08 = tmpname("tmp08")
+    xtralayer = "{}_filledgaps".format(suitreg)
+
+    # Compute neighborhood statistic
+    if (radius > 1) & (len(absminsuit) == 0):
```
wenzeslaus, line 225: Logical AND in Python is `and`. Most people would leave out the `()` around `and` operands.

src/raster/r.suitability.regions/r.suitability.regions.py
```diff
@@ -0,0 +1,410 @@
...
+                "{0} = if({1} > {2},{3},null())".format(
+                    tmp02, suitrast, absminsuit, tmp02a
+                )
+            ),
+        )
+    else:
+        gs.run_command("g.copy", raster=[suitrast, tmp02], quiet=True)
+
+    # Convert suitability to boolean: suitable (1) or not (nodata)
+    gs.message("Creating boolean map suitable/none-suitable")
```
wenzeslaus, line 270: Some of there might be `verbose` (messages) rather than `message`.

src/raster/r.suitability.regions/r.suitability.regions.py
```diff
@@ -0,0 +1,410 @@
...
+    xtralayer = "{}_filledgaps".format(suitreg)
+
+    # Compute neighborhood statistic
+    if (radius > 1) & (len(absminsuit) == 0):
+        gs.message("Computing neighborhood statistic")
+        gs.run_command(
+            "r.neighbors",
+            flags=c,
+            input=suitrast,
+            quiet=True,
```
wenzeslaus, line 231: I think a library-level systematic solution is needed for this, but what you can do to make it better here and now is to have a `quiet` variable to have the one place where you can change True to False (or `quiet_module` or something like that).

wenzeslaus: You can make it global or you can pass it to your helper functions.

src/raster/r.suitability.regions/r.suitability.regions.py
```diff
@@ -0,0 +1,410 @@
...
+    tmp02 = tmpname("tmp02")
+    tmp02a = tmpname("tmp02a")
+    tmp03 = tmpname("tmp03")
+    tmp04 = tmpname("tmp04")
+    tmp05 = tmpname("tmp05")
+    tmp06 = tmpname("tmp06")
+    tmp07a = tmpname("tmp07a")
+    tmp07 = tmpname("tmp07")
+    tmp08 = tmpname("tmp08")
+    xtralayer = "{}_filledgaps".format(suitreg)
```
wenzeslaus, line 222: This name is unclear.

src/raster/r.suitability.regions/r.suitability.regions.py
```diff
@@ -0,0 +1,410 @@
...
+    tmprecode = tmpname("tmprecode")
+    gs.write_command(
+        "r.recode", input=input, output=tmprecode, rule="-", stdin=rules, quiet=True
+    )
+    return tmprecode
+
+
+def main(options, flags):
+
+    suitrast = options["suitrast"]
```
wenzeslaus, line 184: Similarly to GRASS option names, the current trend in Python is using real words plus some special cases such as num, x, etc. for variable names and underscores to separate words.

src/raster/r.suitability.regions/r.suitability.regions.py
```diff
@@ -0,0 +1,410 @@
...
+    absminsuit = options["absminsuit"]
+    radius = int(options["radius"])
+    if (radius % 2) == 0:
+        gs.error("Radius should be an odd positive number")
```
wenzeslaus, line 188: Same fatal vs error issue as below.

src/raster/r.suitability.regions/r.suitability.regions.py
```diff
@@ -0,0 +1,410 @@
...
+import os
+import atexit
+import uuid
+import grass.script as gs
+
+RECLASSRULES1 = """\
+1:filled gaps
+2:suitable areas
+"""
+COLORRULES1 = """
```
wenzeslaus, line 139: Purpose or another description in name would be better for readability instead of numbering those. Underscore to separate words.

wenzeslaus (changes requested):
I reviewed the Python part (not the GRASS module usage). Generally, I recommend running Flake8 and Pylint to quickly get feedback about the code, variable names, etc.

wenzeslaus:
> Not sure though what option I should use, squash and merge or rebase and merge (not sure what the difference is).

It's simple. Only squash and merge is enabled for this repository. Anything else was creating mess on the base branch.

src/raster/r.suitability.regions/r.suitability.regions.py
ecodiv: Useful new function. In this particular case, the easier (less code) solution I found is to use str(uuid.uuid4().hex) which removes the hypen from the name in one go.

src/raster/r.suitability.regions/r.suitability.regions.py
ecodiv: Implemented. But ambiguous feelings about this, as it is difficult to find names that make sense and still are not too long.

src/raster/r.suitability.regions/r.suitability.regions.py
ecodiv: Not sure to which issue below this is refering? Should I use fatal or error?

veroandreo: I guess it refers to the one above. Use gs.fatal if you want to terminate the process there when radius is not odd.

src/raster/r.suitability.regions/r.suitability.regions.py
wenzeslaus: Another complimentary way to improve this is to create the variables only right before they are needed. This gives more context for what they are and lower likelihood of using one sooner than intended.

src/raster/r.suitability.regions/r.suitability.regions.py
wenzeslaus: The particular function is not that relevant here. I meant the link as an example of using import with try-except for version compatibility.

The hex trick is implemented in `gs.append_uuid` (`python/grass/script/utils.py`) together with possibly a better alternative `append_node_pid`. Using one or the other provides an opportunity for future improvement (at one single place in the library) or for correction where a clearly identifiable function can be detected in modules, so it makes sense to use a function even in place of the short code.

https://github.com/OSGeo/grass/blob/master/python/grass/script/utils.py#L537

src/raster/r.suitability.regions/r.suitability.regions.py
ecodiv: Was just doing that  :-). Definitely the better way to go in most cases.

src/raster/r.suitability.regions/r.suitability.regions.py
ecodiv: Yes, changed that already based on your previous remark.

src/raster/r.suitability.regions/r.suitability.regions.py
ecodiv: This is intended to give some feedback on progress. Admittedly, not a great way. Ideally, there would be some way of passing on the progress bar of the individual functions used, or to create a progress bar. Any ideas on how to do that, or other clever ways to provide feedback on progress?

ecodiv:
@wenzeslaus and @veroandreo, thanks for the review and feedback. I have tried to implement all suggestions. Merging now.

PR #577, opened 2021-07-06
i.sentinel.import: Improved cloud masking

fkroeber:
Suggestion for improved cloud masking during import based on the Level-2A inherent cloud probability layer (QI_DATA/*CLDPRB_20m) and scene classification layer (IMG_DATA/R20m/*SCL_20m) instead of the previously used MSK_CLOUDS layer (QI_DATA). The latter didn't include the possibility to mask cloud shadows and areas of clouds were masked less accurately. The comparison below presents the cloud masking output based on the proposed changes in i.sentinel.import (second image) and the result of the current version of i.sentinel.import (last image).

grafik
grafik
grafik

### DIFF
--- a/src/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.html
+++ b/src/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.html
@@ -120,8 +120,8 @@ This mask is based on the cloud probability layer which is part of each Level2A 
 By specifying the <b>cloud_*</b>-options the creation of the cloud mask can be controlled.
 The default cloud masking threshold applied to the probability layer is set to 65%, which corresponds to
 the classification of such areas as most likely cloudy according to the ESA classification scheme used for the creation of the SCL layer.
-The SCL Layer itself is used if shadow masks should be included in the _MSK_CLOUDS layer. This requires to set <em>cloud_shadows=yes</em>,
-which includes shadows as part of the output layer. The area threshold parameter allows to exclude
+The SCL Layer itself is used if shadow masks should be included in the _MSK_CLOUDS layer. This requires to call the <b>-s</b> flag,
+which adds shadows as part of the _MSK_CLOUDS output layer. The area threshold parameter allows to exclude
 small areas of cloud and shadow cover, which in case of vector outputs tend to increase the computational speed.
 Note that the resolution of the _MSK_CLOUDS layer (20m) is determined by the native resolution of the
 cloud probability layer and the SCL layer, respectively.
--- a/src/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.html
+++ b/src/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.html
@@ -6,9 +6,9 @@ module.
 
 <p>
 By default <em>i.sentinel.import</em> imports all Sentinel scene files found
-in the <b>input</b> directory. The number of scene files can be optionally 
-reduced by the <b>pattern_file</b> option. In this option, a regular expression 
-for filtering the file names should be given, e.g. "MSIL2A.*T32VNR_2019" for 
+in the <b>input</b> directory. The number of scene files can be optionally
+reduced by the <b>pattern_file</b> option. In this option, a regular expression
+for filtering the file names should be given, e.g. "MSIL2A.*T32VNR_2019" for
 importing only level 2A products for tile T32VNR from 2019.
 <p>
 By default <em>i.sentinel.import</em> imports the full scene. Optionally, the
@@ -17,13 +17,13 @@ import can be reduced to the computational region extent with
 
 <p>
 Note that in the case that spatial reference system of input data differs
-from GRASS GIS location, the input data need to be reprojected with 
+from GRASS GIS location, the input data need to be reprojected with
 <em><a href="https://grass.osgeo.org/grass-stable/manuals/r.import.html">r.import</a></em>. To speed up this
 process, a higher than default value can be specified for the <b>memory</b>
 option.
 
 <p>
-In order to ignore insignificant mismatch of the spatial reference 
+In order to ignore insignificant mismatch of the spatial reference
 systems, the projection check can be suppressed with the <b>-o</b> flag.
 
 <p>
@@ -109,12 +109,22 @@ AZIMUTH_ANGLE_5=295.354861828927<br>
 </tbody>
 </table>
 
+
 <h2>NOTES</h2>
 
 <p>
-If <b>-c</b> flag is given, than also cloud mask file is imported as
-vector map if available. The name of created vector map is determined from
-input Sentinel product.
+<b>Importing cloud and cloud shadow masks:</b>
+
+If <b>-c</b> flag is given, a cloud mask with suffix _MSK_CLOUDS is additionally created.
+This mask is based on the cloud probability layer which is part of each Level2A product.
+By specifying the <b>cloud_*</b>-options the creation of the cloud mask can be controlled.
+The default cloud masking threshold applied to the probability layer is set to 65%, which corresponds to
+the classification of such areas as most likely cloudy according to the ESA classification scheme used for the creation of the SCL layer.
+The SCL Layer itself is used if shadow masks should be included in the _MSK_CLOUDS layer. This requires to set <em>cloud_shadows=yes</em>,
+which includes shadows as part of the output layer. The area threshold parameter allows to exclude
+small areas of cloud and shadow cover, which in case of vector outputs tend to increase the computational speed.
+Note that the resolution of the _MSK_CLOUDS layer (20m) is determined by the native resolution of the
+cloud probability layer and the SCL layer, respectively.
 
 <p>
 By <b>register_file</b> option <em>i.sentinel.import</em> allows
@@ -125,7 +135,7 @@ currently a register file can be created only for Sentinel-2 data. See
 example below.
 
 <p>
-<b>Metadata import</b>
+<b>Metadata import:</b>
 
 By using the <b>-j</b> flag the band metadata are additionally stored
 in JSON format (in the current mapset under <tt>cell_misc</tt>). These
--- a/src/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
+++ b/src/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
@@ -95,15 +95,6 @@
 #% answer: vector
 #% label: Create cloud mask as raster or vector product
 #%end
-#%option
-#% key: cloud_shadows
-#% type: string
-#% required: no
-#% multiple: no
-#% options: yes, no
-#% answer: yes
-#% label: Include cloud shadows in cloud masking
-#%end
 #%flag
 #% key: r
 #% description: Reproject raster data using r.import if needed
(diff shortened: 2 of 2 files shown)


src/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.html
```diff
@@ -109,12 +109,22 @@ <h2>DESCRIPTION</h2>
...
 </table>
 
+
 <h2>NOTES</h2>
 
 <p>
-If <b>-c</b> flag is given, than also cloud mask file is imported as
-vector map if available. The name of created vector map is determined from
-input Sentinel product.
+<b>Importing cloud and cloud shadow masks:</b>
```
neteler, line 116:
```suggestion
<h3>Importing cloud and cloud shadow masks</h3>
```

(like this it becomes part of the table of content)

src/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.html
```diff
@@ -125,7 +135,7 @@ <h2>NOTES</h2>
 example below.
 
 <p>
-<b>Metadata import</b>
+<b>Metadata import:</b>
```
neteler, line 138:
```suggestion
<h3>Metadata import</h3>
```

src/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
```diff
@@ -3,7 +3,7 @@
 ############################################################################
 #
 # MODULE:      i.sentinel.import
-# AUTHOR(S):   Martin Landa
+# AUTHOR(S):   Martin Landa, Felix Kröber
 # PURPOSE:     Imports Sentinel data downloaded from Copernicus Open Access Hub
 #              using i.sentinel.download.
 # COPYRIGHT:   (C) 2018-2019 by Martin Landa, and the GRASS development team
```
neteler, line 9:
```suggestion
# COPYRIGHT:   (C) 2018-2021 by Martin Landa, and the GRASS development team
```

src/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
```diff
@@ -71,6 +71,30 @@
...
+#% answer: 65
+#%end
+#%option
+#% key: cloud_output
+#% type: string
+#% required: no
+#% multiple: no
+#% options: vector,raster
+#% answer: vector
+#% label: Create cloud mask as raster or vector product
```
neteler, line 96:
```suggestion
#% label: Create cloud mask as raster or vector map
```

neteler:
Very nice work! 
Please take a look at the Python formatting (see the QA checks) which isn't `black` compliant yet.

src/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.html
```diff
@@ -109,12 +109,22 @@ <h2>DESCRIPTION</h2>
...
 
 <p>
-If <b>-c</b> flag is given, than also cloud mask file is imported as
-vector map if available. The name of created vector map is determined from
-input Sentinel product.
+<b>Importing cloud and cloud shadow masks:</b>
+
+If <b>-c</b> flag is given, a cloud mask with suffix _MSK_CLOUDS is additionally created.
+This mask is based on the cloud probability layer which is part of each Level2A product.
+By specifying the <b>cloud_*</b>-options the creation of the cloud mask can be controlled.
```
veroandreo, line 120:
```suggestion
The creation of the cloud mask can be controlled by specifying the <b>cloud_*</b> options.
```

src/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.html
```diff
@@ -109,12 +109,22 @@ <h2>DESCRIPTION</h2>
...
+The default cloud masking threshold applied to the probability layer is set to 65%, which corresponds to
+the classification of such areas as most likely cloudy according to the ESA classification scheme used for the creation of the SCL layer.
+The SCL Layer itself is used if shadow masks should be included in the _MSK_CLOUDS layer. This requires to call the <b>-s</b> flag,
+which adds shadows as part of the _MSK_CLOUDS output layer. The area threshold parameter allows to exclude
```
veroandreo, line 124:
```suggestion
which adds shadows as part of the _MSK_CLOUDS output layer. The <b>cloud_area_threshold</b> parameter allows to exclude
```

veroandreo (commented):
Great addition! I suggest to run black on the code, since there's a mix of single and double quotes. Also, IIRC, we tend to use format instead of f strings.

fkroeber:
> Great addition! I suggest to run black on the code, since there's a mix of single and double quotes. Also, IIRC, we tend to use format instead of f strings.
> 
> Will this work with products from different providers, i.e., ESA, USGS, Google Cloud?

Thanks for reviewing @veroandreo. I've formatted the code accordingly so it should be compliant now. As you can see, I've modified the addition in a way that also allows the import of cloud masks for Level 1C products (based on the mechanism which is implemented in i.sentinel.import right now). So, yes, the procedure works for Level1C and Level2A products from all providers (ESA, GoogleCloud & USGS).

neteler (approved):

veroandreo:
> > Great addition! I suggest to run black on the code, since there's a mix of single and double quotes. Also, IIRC, we tend to use format instead of f strings.
> > Will this work with products from different providers, i.e., ESA, USGS, Google Cloud?
> 
> Thanks for reviewing @veroandreo. I've formatted the code accordingly so it should be compliant now. As you can see, I've modified the addition in a way that also allows the import of cloud masks for Level 1C products (based on the mechanism which is implemented in i.sentinel.import right now). So, yes, the procedure works for Level1C and Level2A products from all providers (ESA, GoogleCloud & USGS).

Great! Would you mind adding an example (ideally with figures also) showing the usage of these new options and flags? Maybe just below the current fig with the cloud mask, in order to show the difference

fkroeber:
> > > Great addition! I suggest to run black on the code, since there's a mix of single and double quotes. Also, IIRC, we tend to use format instead of f strings.
> > > Will this work with products from different providers, i.e., ESA, USGS, Google Cloud?
> > 
> > 
> > Thanks for reviewing @veroandreo. I've formatted the code accordingly so it should be compliant now. As you can see, I've modified the addition in a way that also allows the import of cloud masks for Level 1C products (based on the mechanism which is implemented in i.sentinel.import right now). So, yes, the procedure works for Level1C and Level2A products from all providers (ESA, GoogleCloud & USGS).
> 
> Great! Would you mind adding an example (ideally with figures also) showing the usage of these new options and flags? Maybe just below the current fig with the cloud mask, in order to show the difference

I've adjusted the html documentation using a new example demonstrating the use of the flags and the results. Importing masks for a Level 2A Scene is demonstrated since cloud and shadow layers are both available here.

veroandreo:
> > > > Great addition! I suggest to run black on the code, since there's a mix of single and double quotes. Also, IIRC, we tend to use format instead of f strings.
> > > > Will this work with products from different providers, i.e., ESA, USGS, Google Cloud?
> > > 
> > > 
> > > Thanks for reviewing @veroandreo. I've formatted the code accordingly so it should be compliant now. As you can see, I've modified the addition in a way that also allows the import of cloud masks for Level 1C products (based on the mechanism which is implemented in i.sentinel.import right now). So, yes, the procedure works for Level1C and Level2A products from all providers (ESA, GoogleCloud & USGS).
> > 
> > 
> > Great! Would you mind adding an example (ideally with figures also) showing the usage of these new options and flags? Maybe just below the current fig with the cloud mask, in order to show the difference
> 
> I've adjusted the html documentation using a new example demonstrating the use of the flags and the results. Importing masks for a Level 2A Scene is demonstrated since cloud and shadow layers are both available here.

The dark grey is barely visible, would you mind using colors with a better contrast with the background? Otherwise, I'm positive for merging ;)
Thanks @fkroeber !

fkroeber:
> > > > > Great addition! I suggest to run black on the code, since there's a mix of single and double quotes. Also, IIRC, we tend to use format instead of f strings.
> > > > > Will this work with products from different providers, i.e., ESA, USGS, Google Cloud?
> > > > 
> > > > 
> > > > Thanks for reviewing @veroandreo. I've formatted the code accordingly so it should be compliant now. As you can see, I've modified the addition in a way that also allows the import of cloud masks for Level 1C products (based on the mechanism which is implemented in i.sentinel.import right now). So, yes, the procedure works for Level1C and Level2A products from all providers (ESA, GoogleCloud & USGS).
> > > 
> > > 
> > > Great! Would you mind adding an example (ideally with figures also) showing the usage of these new options and flags? Maybe just below the current fig with the cloud mask, in order to show the difference
> > 
> > 
> > I've adjusted the html documentation using a new example demonstrating the use of the flags and the results. Importing masks for a Level 2A Scene is demonstrated since cloud and shadow layers are both available here.
> 
> The dark grey is barely visible, would you mind using colors with a better contrast with the background? Otherwise, I'm positive for merging ;)
> Thanks @fkroeber !

You're absolutely right, in that it's difficult to recognize the shadows given the color-composite S2 scene in the background. However, not displaying the S2 scene or increasing its transparency makes it easy possible to detect the chosen colors. My consideration here was that it is more valuable to the normal user to have colors that are intuitive in their interpretation instead of using more eye-catching colors that are less easier associated with clouds and shadows. But if you still propose to change the colors, I can do so :)

veroandreo:
> > > > > > Great addition! I suggest to run black on the code, since there's a mix of single and double quotes. Also, IIRC, we tend to use format instead of f strings.
> > > > > > Will this work with products from different providers, i.e., ESA, USGS, Google Cloud?
> > > > > 
> > > > > 
> > > > > Thanks for reviewing @veroandreo. I've formatted the code accordingly so it should be compliant now. As you can see, I've modified the addition in a way that also allows the import of cloud masks for Level 1C products (based on the mechanism which is implemented in i.sentinel.import right now). So, yes, the procedure works for Level1C and Level2A products from all providers (ESA, GoogleCloud & USGS).
> > > > 
> > > > 
> > > > Great! Would you mind adding an example (ideally with figures also) showing the usage of these new options and flags? Maybe just below the current fig with the cloud mask, in order to show the difference
> > > 
> > > 
> > > I've adjusted the html documentation using a new example demonstrating the use of the flags and the results. Importing masks for a Level 2A Scene is demonstrated since cloud and shadow layers are both available here.
> > 
> > 
> > The dark grey is barely visible, would you mind using colors with a better contrast with the background? Otherwise, I'm positive for merging ;)
> > Thanks @fkroeber !
> 
> You're absolutely right, in that it's difficult to recognize the shadows given the color-composite S2 scene in the background. However, not displaying the S2 scene or increasing its transparency makes it easy possible to detect the chosen colors. My consideration here was that it is more valuable to the normal user to have colors that are intuitive in their interpretation instead of using more eye-catching colors that are less easier associated with clouds and shadows. But if you still propose to change the colors, I can do so :)

Well, if you don't mind, yes, please either add transparency to the S2 scene or change the colors (at least that of shadows) to something more visible. I understand your line of thoughts and agree to some point, but if someone reading the manual cannot really distinguish the shadows, then it might be detrimental for the module itself.

fkroeber:
> transparency

@veroandreo please have a look at the transparency-changed example images and approve the changes if they meet your expectations. Thanks ;)

veroandreo (approved):

PR #579, opened 2021-07-07
i.sentinel.download: changes to pandas module import

fkroeber:
Even though pandas is currently imported in the main function (l. 1027 ff.), the module doesn't seem to be available for further use in the called methods, i.e. filter_USGS (l. 1120), which in turn leads to an error when calling i.sentinel.download with producttype=USGS_EE ("Error: Unable to connect to USGS_EE: name 'pandas' is not defined"). Even though I do not understand why importing pandas at the beginning of main() doesn't work, importing pandas at the very beginning of the script (as suggested below) avoids the error. Feel free to change or discard this pull request in case there are better ways to handle the import of pandas!

### DIFF
--- a/src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
+++ b/src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
@@ -169,6 +169,7 @@ import shutil
 import sys
 import logging
 import time
+import pandas
 from collections import OrderedDict
 
 import grass.script as gs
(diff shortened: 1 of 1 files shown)


griembauer:
This was changed in https://github.com/OSGeo/grass-addons/commit/54010b44598cf41448245c75b6ce94107d02dad0#diff-8f46003ae8b0dbe657ce69e1ef7305b221abb83834dfef76ede48c24f04cc315 to importing `pandas` in the main function instead of globally. 
@ninsbl @neteler can you perhaps comment why this lazy import is necessary? As a consequence, is it then necessary to import `pandas` in the relevant method itself?

lucadelu:
@fkroeber @griembauer lazy import is needed specially for compilation, if in an environment there is no `pandas` it compile but it return error only when someone run it.

Before to modify the import would be useful to understand the error. I suggest to close this PR and open an issue for the problem

griembauer:
Thanks for the clarification @lucadelu !

PR #596, opened 2021-08-02
g.projpicker: Import projpicker from the soon-to-be core module

HuidaeCho:
Merge after https://github.com/OSGeo/grass/pull/1891


neteler:
@HuidaeCho pls assign this PR to the `grass8` branch

HuidaeCho:
> @HuidaeCho pls assign this PR to the `grass8` branch

@neteler I think you meant rebasing to grass8? Or adding a new milestone for grass8?

neteler:
At time is is submitted to addons base branch `grass7` (see above). I assume it should be changed to `grass8`, e.g. by using the "Edit" button in the Web interface next to the title.

wenzeslaus:
I think there is also some confusion about what base is being changed. The branch "against" which the PR is made is called _base_ in GitHub interface (see the _Edit_ button for a PR). However, we have also operation _rebase_ (in Git and potentially also in GitHub PR interface when merging) which re-creates all commits from a branch on top of another commit (typically same base branch, just later commit). These two share a lot in common, so perhaps that's the reason for the same name, but who knows, GitHub does have an unfortunate way of naming things IMHO (namely, a _pull request_ on GitHub is a _merge request_ on GitLab).

PR #599, opened 2021-08-20
Update contributing file to master to grass7 switch

wenzeslaus:
* The default (and thus usually the correct base branch) is grass7, not master.
* Some items related to remotes and merging were updated or corrected.
* Some parts were specialized to addons (original text is from core).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,8 +15,8 @@ on GitHub.
 
 ## Changing code and documentation
 
-This guide covers contributing to the main version of GRASS GIS Addons
-source code which is stored the master branch.
+This guide covers contributing to the latest version of GRASS GIS Addons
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -165,11 +165,11 @@ GRASS GIS maintainers or users will now review your pull request.
 If needed, they will work with you to improve your changes.
 
 Once the changes in the pull request are ready to be accepted,
-the maintainers will decide if it is more appropriate to:
+the maintainers will usually squash all your commits into one commit and merge it
+to the _main_ branch.
```
veroandreo, line 169: shouldn't it be *grass7* here too?

wenzeslaus: Right, fixed, thanks.

PR #602, opened 2021-08-27
i.sentinel.download: checksumming and lazy import

griembauer:
This PR adapts `i.sentinel.download` as follows:

- Checksumming from GCS is adapted. Some newer files use `sha3-256` method for checksumming. Also sometimes the checksum indicated in the `manifest.safe` is not the same as for the actually downloaded file, although the file content is identical. For this case, only a warning is now given to the user (instead of assuming the entire downloaded scene is invalid and is removed)

- lazy import is moved to individual functions and methods for `requests` and `pandas`. This seems to be necessary as importing in the main function is not sufficient for the import to be in the environment also for each method/function. Here I am not sure whether this is the best solution, please advise.


ninsbl:
This PR is integrated in #609...

neteler:
@griembauer pls rebase this PR to the (new) `grass8` branch

griembauer:
This PR can be closed because it has been integrated and merged in #609

PR #605, opened 2021-09-01
i.landsat.download: add limit option

griembauer:
This PR adds an optional `limit` options to `i.landsat.download` (similar to `i.sentinel.download`, https://github.com/OSGeo/grass-addons/blob/grass7/src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py#L88) to limit the maximum number of scenes to be listed/downloaded.


veroandreo (approved):
Thanks @griembauer !!

PR #607, opened 2021-09-06
i.sentinel.import: change to interval temporal type

landam:
WORKAROUND for https://github.com/OSGeo/grass/issues/1843


metzm:
It seems that `t.rast.mapcalc` behaves as documented in [OSGEO/grass#1843](https://github.com/OSGeo/grass/issues/1843), if more than one strds is given as input. Try to use `t.rast.algebra`instead.

This PR should rather not be applied.

landam:
Right, closing as invalid.

Issue #580, opened 2021-07-08
i.sentinel.import: pandas module import error


fkroeber:
Importing pandas in the main() function instead of globally was changed recently in 54010b4. As @lucadelu pointed out, this change to lazy import is needed especially for compilation.
Unfortunately, this change of import leads to an import error, since the module pandas does not seem to be available for further use in the called methods, i.e. filter_USGS (l. 1120), anymore. When calling i.sentinel.download with producttype=USGS_EE the error "Error: Unable to connect to USGS_EE: name 'pandas' is not defined" occurs.
I'd be helpful to understand why importing pandas at the beginning of main() doesn't work.
Also see #579

ninsbl:
Still an issue?

ecodiv:
Is this still an issue?  I had what seems to be a similar issue but with pyplot from matplotlib. I imported pyplot in the main() function, while using it in a function defined before main (as seems the case for pandas in i.sentinel.download). Importing it in the function where it is actually used resolved the issue. Not sure though if this is a proper solution.

PR #609, opened 2021-09-10
i.sentinel.download: add support for Sentinel-3

ninsbl:
Thiss PR adds support for downloading Sentinel-3 data. It also addresses import issuses for optional libraries...

I am also working on import functions for Sentinel-3 level 2 products (LST in particular). But the format is not so well documented and differs between products. Further more, it is not supported by GDAL out of the box, so I am not sure if it would be appropriateto add it to `i.sentinel.import` or if a separate module would be needed.

### DIFF
--- a/src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.html
+++ b/src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.html
@@ -28,6 +28,21 @@ for download from the <b>Copernicus Open Access Hub</b>:
     <li>S2MSI1C: Top-Of-Atmosphere reflectances in cartographic geometry (Level-1C)</li>
   </ul>
   </li>
+  <li>Sentinel-3 (OLCI and SLSTR instrument data products at level 2; available from April 2018 to present day)
+  (optical) <a href="https://sentinel.esa.int/web/sentinel/missions/sentinel-3/data-products">products</a>:
+  <ul>
+    <li>S3OL2LFR: Land and atmosphere geophysical parameters</li>
+    <li>S3OL2LRR: Land and atmosphere geophysical parameters at reduced resolution</li>
+    <li>S3SL2LST: Land Surface Temperature</li>
+    <li>S3SL2FRP: Fire Radiative Power</li>
+    <li>S3SR2LAN: Land Surface Height</li>
+    <li>S3SY2SYN: Surface Reflectance and Aerosol parameters over Land</li>
+    <li>S3SY2VGP: 1 km VEGETATION-Like product (~VGT-P) - TOA Reflectance</li>
+    <li>S3SY2VG1: 1 km VEGETATION-Like product (~VGT-S1) 1 day synthesis surface reflectance and NDVI</li>
+    <li>S3SY2V10: 1 km VEGETATION-Like product (~VGT-S10) 10 day synthesis surface reflectance and NDVI</li>
+    <li>S3SY2AOD: Global Aerosol parameter over land and sea on super pixel resolution (4.5 km x 4.5 km)</li>
+  </ul>
+  </li>
 </ul>
 
 <p>
--- a/src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
+++ b/src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
@@ -68,7 +68,7 @@
 #% description: Sentinel product type to filter
 #% label: USGS Earth Explorer only supports S2MSI1C
 #% required: no
-#% options: SLC,GRD,OCN,S2MSI1C,S2MSI2A,S2MSI2Ap,S3SL2LST
+#% options: SLC,GRD,OCN,S2MSI1C,S2MSI2A,S2MSI2Ap,S3OL2LFR,S3OL2LRR,S3SL2LST,S3SL2FRP,S3SY2SYN,S3SY2VGP,S3SY2VG1,S3SY2V10,S3SY2AOD,S3SR2LAN
 #% answer: S2MSI2A
 #% guisection: Filter
 #%end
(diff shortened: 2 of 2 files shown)


veroandreo:
Will test soon! Seems like a great improvement, @ninsbl !!

marisn:
Would it be possible to set band reference for each imported raster automatically? It is just a matter of calling Rast_write_bandref("map_name", "band_name") from python. See raster metadata tests for an example:  https://github.com/OSGeo/grass/blob/main/lib/raster/testsuite/test_raster_metadata.py#L94

ninsbl:
> Would it be possible to set band reference for each imported raster automatically? 

Yes, that would be fully possible and should be done. `i.sentinel.import` currently writes band references to a text file for `t.register` but not into map metadata yet. Should be done here: https://github.com/OSGeo/grass-addons/blob/31cd7ae00f454a698fcb70b38a0ff8b2eca3909a/src/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py#L911 I guess...

marisn:
> Yes, that would be fully possible and should be done. `i.sentinel.import` currently writes band references to a text file for `t.register` but not into map metadata yet. Should be done here:
> 
> https://github.com/OSGeo/grass-addons/blob/31cd7ae00f454a698fcb70b38a0ff8b2eca3909a/src/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py#L911
> I guess...

Seems like the right place. That would be good as it would make band references available for all GRASS modules, visible in r.info output etc.

ninsbl:
OK, will try to add that in a separate PR.

metzm:
Regarding Sentinel-3 LST import, I discovered that this product comes as several NetCDF files, suggesting a raster, but these NetCDF files are actually describing a vector point cloud in a very inefficient way. The vector points have irregular spacing can can not be directly converted to a raster map. The commands I used are

```
# latitudes
gdal_translate -of XYZ -unscale -ot Float32 NETCDF:"geodetic_in.nc":latitude_in lat_rc.xyz
# longitudes
gdal_translate -of XYZ -unscale -ot Float32 NETCDF:"geodetic_in.nc":longitude_in lon_rc.xyz
# lST
gdal_translate -of XYZ NETCDF:"LST_in.nc":LST lst_rc.xyz

echo "convert latitudes..."
cat lat_rc.xyz | cut -d ' ' -f3 >lat.xyz

echo "convert longitudes..."
cat lon_rc.xyz | cut -d ' ' -f3 >lon.xyz

echo "extract LST..."
cat lst_rc.xyz | cut -d ' ' -f3 >lst.xyz

echo "paste coordinates and LST..."
paste -d '|' lon.xyz lat.xyz lst.xyz >lst_ll.xyz
```
not sure how to proceed from here, `r.in.xyz` does not produce a good result.

Full scene as point cloud:

S3_LST_full_szene

zoom in to illustrate irregular spacing of the points:

S3_LST_zoom

ninsbl:
> Regarding Sentinel-3 LST import, I discovered that this product comes as several NetCDF files, suggesting a raster, but these NetCDF files are actually describing a vector point cloud in a very inefficient way. The vector points have irregular spacing can can not be directly converted to a raster map.

Yes, there are a couple of complaints about the format:
http://blog.imagico.de/sentinel-3-a-first-look-at-the-data-part-2/
https://forum.sentinel-hub.com/t/s3-slstr-lst-product-grid-anomaly/2764/4

And I do not understand how this was engineered either. May the meorologists fondness of netCDF... I also tried exporting a Sentinel-3 product from ESA SNAP and reading the resulting GeoTiff, but the result was pretty bad, frankly.

My current approach is to use netCDF4 in python to transform the data into a format readable by `r.in.xyz`. Will add that here (different PR) soon for review...

metzm:
> My current approach is to use netCDF4 in python to transform the data into a format readable by `r.in.xyz`. Will add that here (different PR) soon for review...

I had no success with `r.in.xyz`: using the nominal resolution of 1 km, I got many small gaps in the raster which would need to be filled. With a resolution of 2 km, these gaps disappear, but 2 km is too coarse...
Maybe `v.in.ascii` + `v.surf.*` ?

That S3 LST format is a real pain...

veroandreo (approved):
I've tested listing available products and downloading by `uuid` and all works as expected :)

griembauer (commented):
Thank you, great addition! I don't have experience with S-3 data but listing and downloading worked for all added products for me too. However there is still a problem with the lazy import of modules (see also #580 and #602):. These libraries are only required when downloading from GCS, so the issue gets only visible when trying to download S2-L2A data from GCS. Here I think we should harmonize the importing with #602

ninsbl:
Thanks @fkroeber . I do not have a GCS account, so I cannot test that functionalty.
I am no programmer by education, so my knowledge of this stuff is only "experience based". But from my understanding, the reason the lazy import does not work in the functions is probably due to name spaces.

We probably have to lazy imports more like here:
https://github.com/OSGeo/grass-addons/blob/050be44852a977747217b530d210f31789b4632d/src/raster/r.denoise/r.denoise.py

Can update the PR accordingly. Would be great if you then could check if that helps...

griembauer:
Thanks a lot @ninsbl this indeed solves the issue and I would agree to go that way for the lazy importing (little programming foundation from my side too ;) ). If you agree to the changes, you could also implement the checksumming changes from #602 - then we could close #602 and keep going with this PR or #611. 

You can download from GCS without a GCS account by simply setting the `datasource=GCS` but still using your standard ESA credentials in the `settings`. This way, data is queried on the ESA-HUB and then the required (openly accessible!) GCS download URL is automatically created. Google only charges for the use of Big Query, while the data themselves are openly accessible.

griembauer (approved):

ninsbl:
Thanks for the info on GCS download options.
Will try to have a look at merging #602, #609 and #611 hopefully later today...

PR #610, opened 2021-09-13
i.sentinel.import: write semantic label with r.support

ninsbl:
This adds the bandref directly to the raster map. I think if would also have been written upon registration in the temporal database but this way it is available without using temporal modules...

As discussed in #609

### DIFF
--- a/src/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
+++ b/src/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
@@ -923,7 +923,7 @@ class SentinelImporter(object):
                     "source2": img_file,
                     "history": descr,
                 }
-                # Band refrence available after version 7.9
+                # Band references/semantic labels available from GRASS GIS 8.0.0 onwards
                 if float(gs.version()["version"][0:3]) >= 7.9:
                     support_args["bandref"] = self._extract_bandref(map_name)
                 for band in bands:
--- a/src/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
+++ b/src/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
@@ -352,20 +352,22 @@ class SentinelImporter(object):
     def _map_name(filename):
         return os.path.splitext(os.path.basename(filename))[0]
 
-    def _extract_bandref(self, map_name):
-        band_ref = None
+    def _extract_semantic_label(self, map_name):
+        semantic_label = None
         try:
-            band_ref = re.match(
+            semantic_label = re.match(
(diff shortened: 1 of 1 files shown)


marisn (approved):
Looks good for a moment.
If band references will be renamed (#1868), this code also will have to change.

Regarding version dependent code – wasn't there an idea to have a separate add-ons branch for 8 thus avoiding version checking in the code?

ninsbl:
Thanks @marisn There has been (a not very active) discussion about versions and addons here:
https://www.mail-archive.com/grass-dev@lists.osgeo.org/msg60905.html

It is a minor change and we can wait until branch and option names are settled and then update before merging...

src/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
```diff
@@ -948,21 +965,12 @@ def create_register_file(self, filename):
...
-                        ).groups()
-                        band_ref = band_ref[0] if band_ref[0] else band_ref[1]
-                        band_ref = band_ref.lstrip("0")
-                    except AttributeError:
-                        gs.warning(
-                            _("Unable to determine band reference for <{}>").format(
-                                map_name
-                            )
-                        )
+                # Band refrence available after version 7.9
```
neteler, line 968:
```suggestion
                # Band references/semantic labels available from GRASS GIS 8.0.0 onwards
```

src/imagery/i.sentinel/i.sentinel.import/i.sentinel.import.py
```diff
@@ -902,14 +917,17 @@ def write_metadata(self):
                 )
                 if flags["j"] and not os.path.isdir(json_standard_folder):
                     os.makedirs(json_standard_folder)
+                support_args = {
+                    "map": map_name,
+                    "source1": ip,
+                    "source2": img_file,
+                    "history": descr,
+                }
+                # Band refrence available after version 7.9
```
neteler, line 926:
```suggestion
                # Band references/semantic labels available from GRASS GIS 8.0.0 onwards
```

marisn:
In G8 band references are renamed to semantic labels (ditto for r.support parameter). This PR should be adjusted accordingly and then it's +1 from me.

neteler:
@ninsbl pls rebase this PR to the (new) `grass8` branch

ninsbl:
OK to merge, or should I remove the version check since this goes into thegrass8 branch anyway?

wenzeslaus:
> should I remove the version check since this goes into thegrass8 branch anyway?

I would remove it to simplify the code, but you have it already there and if you anticipate a lot of messy backports, it is not bad to keep it there as long as it is useful.

PR #611, opened 2021-09-15
i.sentinel.download: print size and skip existing scenes

ninsbl:
Currently, scenes present in the `input` directory are downloaded again if they are part of a query. This PR adds a flag to skip downloading those again if the ingestion date of the scscene is not newer than the creation time of the local file.

Additionally, `-l` now also writes size of the scenes, wich can be useful information if one lists scenes before downloading...

The base branch of this PR is that of #609 (not sure how the situation ideally would be handled in git when one wants to add other functionality on top of a not yet merged PR). Will change the PR according to recommendations then.

### DIFF
--- a/src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
+++ b/src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
@@ -68,7 +68,7 @@
 #% description: Sentinel product type to filter
 #% label: USGS Earth Explorer only supports S2MSI1C
 #% required: no
-#% options: SLC,GRD,OCN,S2MSI1C,S2MSI2A,S2MSI2Ap,S3SL2LST
+#% options: SLC,GRD,OCN,S2MSI1C,S2MSI2A,S2MSI2Ap,S3OL2LFR,S3OL2LRR,S3SL2LST,S3SL2FRP,S3SY2SYN,S3SY2VGP,S3SY2VG1,S3SY2V10,S3SY2AOD,S3SR2LAN
 #% answer: S2MSI2A
 #% guisection: Filter
 #%end
@@ -796,7 +796,7 @@ class SentinelDownloader(object):
 
         # Sentinel-1 data does not have cloudcoverpercentage
         prod_types = [type for type in self._products_df_sorted["producttype"]]
-        if any(type in prod_types for type in no_cloudcoverpercentage):
+        if not any(type in prod_types for type in cloudcover_products):
             del attrs["cloudcoverpercentage"]
 
         for key in attrs.keys():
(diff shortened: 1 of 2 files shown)


src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
```diff
@@ -674,9 +682,57 @@ def list(self):
...
+        # Filter by ingestion date
+        skiprows = []
+        for idx, display_id in enumerate(self._products_df_sorted["identifier"]):
+            existing_file = [sfile for sfile in existing_files if display_id in sfile]
+            if existing_file:
+                creation_time = datetime.fromtimestamp(
+                    os.path.getctime(existing_file[0])
+                )
+                if self._products_df_sorted["ingestiondate"][idx] <= creation_time:
+                    gs.verbose(
```
griembauer, line 718: I would suggest to change this to `gs.message()` so that it is always visible.

griembauer (commented):
(Checked only the new skip download functionality, the rest of the PR is reviewed in #609 ) Thanks a lot, that's a very helpful addition! It would additionally be helpful if the skip-download functionality would also work even if no `ingestiondate` is in the metadata (e.g. when data was downloaded form USGS_EE). A warning could then be given that no check for the local version vs. online version of the file can be performed, but the file is skipped nonetheless if found in the output folder.

src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
```diff
@@ -1126,6 +1192,9 @@ def main():
     if options["footprints"]:
         downloader.save_footprints(options["footprints"])
 
+    if flags["s"]:
+        downloader.skip_existing(options["output"], "*.zip")
```
griembauer, line 1196: When downloading from GCS or USGS_EE, the result is in .SAFE format (not sure about ESA result at the moment). It would be good to scan for this file ending too

src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
ninsbl: Good point. Integrated in #609

src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
ninsbl: Right. Some users may also extract the zip, link the SAFE format (`r.external`) and then remove the zip...
Integrated in #609

ninsbl:
This PR is integrated in #609. Closing this one...

PR #613, opened 2021-09-17
fix cron_grass_build_bins scripts (78/80)

landam:
* align grass-addons to the new layout [grass7 branch] (critical)
* use already defined vars (cosmetics)
* don't run distclean twice (cosmetics)
* fix ARCH variable (critical)
* remove unknown configure options (cosmetics)

### DIFF
--- a/tools/cronjobs_osgeo_lxd/cron_grass78_releasebranch_78_build_bins.sh
+++ b/tools/cronjobs_osgeo_lxd/cron_grass78_releasebranch_78_build_bins.sh
@@ -60,7 +60,7 @@ LDFLAGSSTRING='-s'
 MAINDIR=/home/neteler
 # where to find the GRASS sources (git clone):
 SOURCE=$MAINDIR/src/
-BRANCH=releasebranch_7_$GMINOR
+BRANCH=releasebranch_${GMAJOR}_$GMINOR
 GRASSBUILDDIR=$SOURCE/$BRANCH
 TARGETMAIN=/var/www/code_and_data/
 TARGETDIR=$TARGETMAIN/grass${VERSION}/binary/linux/snapshot
@@ -99,7 +99,7 @@ CFLAGS=$CFLAGSSTRING LDFLAGS=$LDFLAGSSTRING ./configure \
   --with-geos \
   --with-odbc \
   --with-cairo --with-cairo-ldflags=-lfontconfig \
-  --with-proj --with-proj-share=/usr/share/proj \
+  --with-proj-share=/usr/share/proj \
   --with-postgres --with-postgres-includes=/usr/include/postgresql \
   --with-freetype --with-freetype-includes=/usr/include/freetype2 \
   --with-netcdf \
@@ -109,7 +109,6 @@ CFLAGS=$CFLAGSSTRING LDFLAGS=$LDFLAGSSTRING ./configure \
   --with-blas --with-blas-includes=/usr/include/atlas/ \
   --with-lapack --with-lapack-includes=/usr/include/atlas/ \
   --with-zstd \
-  --without-motif \
   --with-liblas \
   2>&1 | tee config_$DOTVERSION.git_log.txt
 
@@ -142,10 +141,9 @@ cp -f *.csv $TARGETMAIN/uploads/grass/
 
 #configure
 echo "configuring"
-$MYMAKE distclean  > /dev/null
 configure_grass || (echo "$0: an error occured" ; exit 1)
 pwd
-ARCH=`cat config.status |grep '@host@' | cut -d'%' -f3`
+ARCH=`cat include/Make/Platform.make | grep ^ARCH | cut -d'=' -f2 | xargs`
 
 ########  now GRASS is prepared ####################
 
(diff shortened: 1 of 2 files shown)


landam:
Converted to draft: `cron_grass8_HEAD_build_bins.sh` should exists only in `grass8` branch when it will be created.

wenzeslaus:
The _grass8_ branch was created and it is not the default branch (see also #528). Besides other changes needed, you may want to change base branch for this PR (_Edit_ next to the PR title).

I like the use of variables in the script, BTW!

neteler:
> Converted to draft: `cron_grass8_HEAD_build_bins.sh` should exists only in `grass8` branch when it will be created.

@landam ...hence this PR should be split into two PRs, right?
(one for `grass7` branch and one for `grass8` branch)

neteler:
I suggest to move on in the following direction:

- rename `cron_grass78_releasebranch_78_build_bins.sh` -> `cron_grass_releasebranch_build_bins.sh` and handle the version as a parameter when calling the script in cron (`78`or `80`)
- rename `cron_grass8_HEAD_build_bins.sh` -> `cron_grass8_main_build_bins.sh`

Importantly, keep all in the `grass8` default addon branch to avoid the need of switching branches during compilation. The reason is that we could encounter a race condition if, say, the  `cron_grass8_main_build_bins.sh` still runs when  `cron_grass_releasebranch_build_bins.sh 78` starts (and switches to a different branch in the same compilation directory). By running all addon compilations from the same branch after the generalization of the releasebranch_build_bins scripts this could be avoided.
Or, we have to duplicate `grass8` default addon branch and have a separate clone of the local repo pointing to the `grass7` addon branch.

Opinions?

wenzeslaus:
> ...have a separate clone of the local repo pointing to the `grass7` addon branch.

I'm not sure sure what duplicate grass8 means nor what switching now could cause race conditions, but if all you need is latest source code without any history, doing _clone_ with `--depth=1 --branch` is fairly cheap and thus can be repeated many times.

```sh
git clone --depth=1 --branch {commit_tag_or_branch_name}
```

I'm not sure if not cloning and updating instead would bring much benefit, but _fetch_ can also take `--depth=1`.

neteler:
> > ...have a separate clone of the local repo pointing to the `grass7` addon branch.
> 
> I'm not sure sure what duplicate grass8 means nor what switching now could cause race conditions

So far we had one active addon branch, now we  have two active ones (G7 and G8) in separate branches.
Depending on how long each compilation job takes on the server, one could overlap with the second depending on when the start time is set in crontab. Now, when both branches are compiled from the same "physical" directory on the server just by switching the branch in the cronjob, this could lead to a mess in case of overlapping cronjobs. However, it is hard to know a priori for how long each job runs.

Hence my suggestion, to run the compile jobs on the server in two distinct directories. Here, your `git clone with --depth=1 --branch ...` suggestion is a good idea.

wenzeslaus:
> ...compiled from the same "physical" directory on the server...overlapping cronjobs...hard to know a priori for how long each job runs...two distinct directories...`git clone --depth=1...` suggestion is a good idea.

Yes, definitively `git clone --depth=1 --branch...` used in a unique temporary directory (`mktemp -d` and traps, perhaps) for each cron job run.

neteler:
@landam @wenzeslaus @tmszi I would like to merge this PR and then move on with a new one. I think that's easier

wenzeslaus:
Merging this now makes sense to me. There is some code commented out, but more changes are needed which need to be done while testing on the build server I suppose.

With big changes and renames, would that be hard to also move rename the _tools_ directory to _utils_ like we have it in the main repo.

> ...rename `cron_grass78_releasebranch_78_build_bins.sh` -> `cron_grass_releasebranch_build_bins.sh`...

Yes.

> ...handle the version as a parameter when calling the script in cron (`78`or `80`)

I suggest using the identifiers from Git (whole branch names, tags) as much as possible or somewhat real version numbers (7, 7.8, 7.8.0), or adding multiple parameters explicitly if really needed (like `releasebranch_7_8 7.8 78 7`). I think that leads to more universal scripts which are potentially simpler.

> rename `cron_grass8_HEAD_build_bins.sh` -> `cron_grass8_main_build_bins.sh`

I'm not sure what the script is, but the new name is better.

> ...keep all in the `grass8` default addon branch to avoid the need of switching branches during compilation.

Switching the branch with scripts sounds bad. All in grass8 sounds good. Alternatively, there could be only some top-level script in the default branch which would switch branches and trigger some main in each branch, but that could be hard to navigate. We could also have a separate repo with these script which would avoid the v7 stuff in grass8 confusion (and maybe some confusing clone/update situation on the server), but perhaps not a big deal.

Ideally, the scripts be written in the way that they could run also in GitHub Actions, but that's for some later time.

neteler:
> Merging this now makes sense to me. There is some code commented out, but more changes are needed which need to be done while testing on the build server I suppose.

Yes, exactly.

> With big changes and renames, would that be hard to also move rename the _tools_ directory to _utils_ like we have it in the main repo.

Let's do that immediately after merging this PR (suggestion: now).

> > ...rename `cron_grass78_releasebranch_78_build_bins.sh` -> `cron_grass_releasebranch_build_bins.sh`...
> 
> Yes.
> 
> > ...handle the version as a parameter when calling the script in cron (`78`or `80`)
> 
> I suggest using the identifiers from Git (whole branch names, tags) as much as possible or somewhat real version numbers (7, 7.8, 7.8.0), or adding multiple parameters explicitly if really needed (like `releasebranch_7_8 7.8 78 7`). I think that leads to more universal scripts which are potentially simpler.

I agree to the simplification of these scripts. They have been developed for CSV, then updated to SVN, now git...).

> > rename `cron_grass8_HEAD_build_bins.sh` -> `cron_grass8_main_build_bins.sh`
> 
> I'm not sure what the script is, but the new name is better.

It builds the Linux weekly binary snapshot installer.

> > ...keep all in the `grass8` default addon branch to avoid the need of switching branches during compilation.
> 
> Switching the branch with scripts sounds bad. All in grass8 sounds good.

(which may not work with the GRASS 7 addons as this requires to actually switch the branch)

> Alternatively, there could be only some top-level script in the default branch which would switch branches and trigger some main in each branch, but that could be hard to navigate. We could also have a separate repo with these script which would avoid the v7 stuff in grass8 confusion (and maybe some confusing clone/update situation on the server), but perhaps not a big deal.

I am quite used to simply checkout the branches into different directories on the server (with `git ... --depth=1 --branch ... ` limitation).

> Ideally, the scripts be written in the way that they could run also in GitHub Actions, but that's for some later time.

I shall keep in mind for the next PR.

wenzeslaus:
> > Switching the branch with scripts sounds bad. All in grass8 sounds good.
> 
> (which may not work with the GRASS 7 addons as this requires to actually switch the branch)

It seems to me that to do that build, you have one grass-addons clone with the grass8 branch just to get access to the scripts. Than you clone again to get grass7, grass8 and whatever you need. In other words, there is a main grass-addons clone which is used just to get the tools/utils directory with scripts, i.e., it is used as if it would be a separate repo with scripts. Anyway, that's what I though you are aiming for.

neteler:
Ok, I will consider that.

My suggestion for now
- merge this PR
- rename `tools` to `utils`
- create a new PR, continuing this one

neteler:
Unfortunately this fails due to the branch protection, how to solve that?

```
~/software/grass-addons [grass8|✔] 
18:33 $ git cherry-pick   0c6e694
[grass8 8ab9671f] fix cron_grass_build_bins scripts (78/80) (#613)
 Author: Martin Landa <landa.martin@gmail.com>
 Date: Wed Dec 22 18:32:34 2021 +0100
 2 files changed, 45 insertions(+), 30 deletions(-)
 
/software/grass-addons [grass8 ↑·1|✔] 
18:35 $ git push upstream grass8
Enumerating objects: 11, done.
Counting objects: 100% (11/11), done.
Delta compression using up to 12 threads
Compressing objects: 100% (2/2), done.
Writing objects: 100% (6/6), 1.51 KiB | 1.51 MiB/s, done.
Total 6 (delta 5), reused 4 (delta 4), pack-reused 0
remote: Resolving deltas: 100% (5/5), completed with 5 local objects.
remote: error: GH006: Protected branch update failed for refs/heads/grass8.
remote: error: Required status check "flake8" is expected. Changes must be made through a pull request.
```

neteler:
Merged into `grass7` branch.

(fwd-ported to `grass8` in #648)

PR #635, opened 2021-11-23
d.region.grid: Add new tool to display grids

wenzeslaus:
Unlike d.grid, d.region.grid (a wrapper around two d.grid calls) can display grid lines for the a computation region without
the user first getting the resolution manually. It also respects that the resolution might be different in each direction.
Finally, it also supports displaying grid based on a resolution of a raster map.

It has limitations inherited from d.grid, so it can't limit the grid extent (d.grid draws full rendering extent/region)
and it can't draw marks instead of lines (d.grid has one resolution and uses that to draw a point grid).

Two possible lines of uses cases are drawing region grid where the EW and NS resolutions differ
(requires two d.grid calls) and drawing grid of raster or region (no additional calls needed to get the
resolution, unless you are in GUI, but there you can at least avoid handling the numbers directly).


src/display/d.region.grid/d.region.grid.html
```diff
@@ -0,0 +1,151 @@
+<h2>DESCRIPTION</h2>
+
+<em>d.region.grid</em> plots a grid defined by the a computational region
```
veroandreo, line 3:
```suggestion
<em>d.region.grid</em> plots a grid defined by the computational region
```

src/display/d.region.grid/d.region.grid.html
```diff
@@ -0,0 +1,151 @@
...
+or by a raster map.
+The computational region can be the current computational region or a saved
+computational region.
+
+<h2>EXAMPLES</h2>
+
+<h3>Comparing grids of two rasters</h3>
+
+To compare how grids (resolutions) of two rasters align,
+here an digital elevation model and a Landsat image,
```
veroandreo, line 13:
```suggestion
here a digital elevation model and a Landsat image,
```

src/display/d.region.grid/d.region.grid.html
```diff
@@ -0,0 +1,151 @@
...
+g.region region=rural_1m grow=-250 -p
+</pre></div>
+
+Start a monitor (skip this in the GUI):
+
+<div class="code"><pre>
+d.mon start=cairo width=600 height=400 output=two_rasters.png
+</pre></div>
+
+Optionally, show the digital elevation model an its shaded relief raster:
```
veroandreo, line 30:
```suggestion
Optionally, show the digital elevation model and its shaded relief raster:
```

src/display/d.region.grid/d.region.grid.html
```diff
@@ -0,0 +1,151 @@
...
+<div class="code"><pre>
+d.region.grid raster=lsat7_2002_10 color=black
+</pre></div>
+
+<center>
+    <img src="d_region_grid_two_rasters.png">
+    <p><em>Figure: Grids of two rasters which are not aligned</em></p>
+</center>
+
+If you are using <em>d.mon</em>, you can stop the monitor it using:
```
veroandreo, line 53:
```suggestion
If you are using <em>d.mon</em>, you can stop the monitor using:
```

src/display/d.region.grid/d.region.grid.html
```diff
@@ -0,0 +1,151 @@
...
+
+<div class="code"><pre>
+d.mon stop=cairo
+</pre></div>
+
+<h3>Showing the current computational region grid</h3>
+
+Let's say you want to render grid cells of a computational region
+you plan to resample a raster map into.
+First, set the computational region to the raster map, than start
```
veroandreo, line 63:
```suggestion
First, set the computational region to the raster map, then start
```

src/display/d.region.grid/d.region.grid.html
```diff
@@ -0,0 +1,151 @@
...
+the monitor (here we use file-based rendering in the command line
+with the cairo driver), and than finally, render the raster map:
```
veroandreo, line 65:
```suggestion
with the cairo driver), and finally, render the raster map:
```

src/display/d.region.grid/d.region.grid.html
```diff
@@ -0,0 +1,151 @@
...
+
+<div class="code"><pre>
+g.region raster=elevation
+d.mon start=cairo width=600 height=400 output=new_region_grid.png
+d.rast map=elevation
+</pre></div>
+
+Than change to the desired computational region, here the new region resolution
```
veroandreo, line 73:
```suggestion
Then change to the desired computational region, here the new region resolution
```

src/display/d.region.grid/d.region.grid.html
```diff
@@ -0,0 +1,151 @@
...
+is set to 1000 meters (map units) and fitted into the current region extent
+without modifying the 1000 meters value.
+
+<div class="code"><pre>
+g.region res=1000 -a
+</pre></div>
+
+Now, you are ready to plot the grid of the current computational region:
```
veroandreo, line 81:
```suggestion
Now you are ready to plot the grid of the current computational region:
```

src/display/d.region.grid/d.region.grid.html
```diff
@@ -0,0 +1,151 @@
...
+Assuming you used <em>d.mon</em> to start rendering as in the code above,
+you can stop it using:
+
+<div class="code"><pre>
+d.mon stop=cairo
+</pre></div>
+
+<h3>Using an existing saved region</h3>
+
+Displaying a saved computation region is extremely helpful in GUI,
```
veroandreo, line 104:
```suggestion
Displaying a saved computational region is extremely helpful in GUI,
```

veroandreo (approved):
Minor grammar fixes only :)

src/display/d.region.grid/d.region.grid.html
wenzeslaus: I like the comma there. Like in "First, ...".

PR #637, opened 2021-12-03
fix addons broken due to GRASS 8 modules API

### LINKED ISSUE #616: [Bug] Some AddOns broken for GRASS 8 due to their API change
After the branch for GRASS 8 will be finally created (see #528), fix addons affected by this change. 

The PR will tag original authors of the addons as reviewers to give them a chance to state their opinion about changes.

So far those addons were discovered to need some care:

- [x] i.sentinel.coverage - caused by a change in v.db.select
- [x] v.habitat.dem - caused by a change in v.db.select
- [x] r.valley.bottom (temporarily fixed by an if statement) - going to have a PR for `grass7` and `grass8` branch - caused by a change in r.neighbors

pesekon2:
fix #616

### DIFF
--- a/src/vector/v.habitat.dem/v.habitat.dem.py
+++ b/src/vector/v.habitat.dem/v.habitat.dem.py
@@ -157,15 +157,7 @@ def main():
     PixelArea = Xres * Yres
     global tmp
 
-    ## check if r.geomorphon addon is installed
-    if not grass.find_program("r.geomorphon", "--help"):
-        grass.fatal(
-            _("The 'r.geomorphon' addon was not found, install it first:")
-            + "\n"
-            + "g.extension r.geomorphon"
-        )
-
-    ## check if r.geomorphon addon is installed
+    ## check if r.sun.hourly addon is installed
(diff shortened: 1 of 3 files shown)


src/vector/v.habitat.dem/v.habitat.dem.py
```diff
@@ -285,7 +277,7 @@ def main():
     grass.run_command(
         "v.db.select",
         map=v_habitat,
-        flags="v",
+        format="vertical",
```
hellik, line 280: Looks good

src/vector/v.habitat.dem/v.habitat.dem.py
```diff
@@ -296,7 +288,7 @@ def main():
...
```
hellik, line 291: Looks good

hellik:
Changes checked for v.habitat.dem, looks fine. Thanks for updaating.

src/vector/v.habitat.dem/v.habitat.dem.py
pesekon2: Thanks for the review.

PR #640, opened 2021-12-07
CI: Test with v8.0 in grass8 branch

wenzeslaus:
* Instead of v7.8, test with v8.0 release branch in the grass8 branch.
* Add test with Python 3.10.
* Remove test with Python 3.6.
* Keep testing only every second version to reduce jobs - add a comment about that.
* Use (explicit) strings to specify versions (not floats).


wenzeslaus:
Thanks @neteler for the help here and in the main repo!

PR #641, opened 2021-12-09
t.rast.vi: New addon to calculate vegetation indices from temporal datasets

lucadelu:
There is a problem with ParallelModuleQueue, it run the first MultiModule but after that it stop and it remain appended. Is there any suggestions?


ninsbl:
The MultiModule issue is fixed in GRASS GIS 8.0. See: 
- https://github.com/OSGeo/grass/pull/1407 and
- https://lists.osgeo.org/pipermail/grass-dev/2021-May/095186.html

It has not (yet) been backported to 7.8 though due to a small API change... If this is needed in 7.8, maybe make a backport PR (direct backporting is probably not possible)?

echoix (changes requested):
Will need a .md file now

neteler:
> Will need a .md file now

For this, see
https://github.com/OSGeo/grass/blob/main/utils/grass_html2md.sh

veroandreo (approved):
I fixed a minor thing in the manual only. Tests run and the results are fine.

petrasovaa (changes requested):
If I remember seeing @veroandreo running the tests, they take too much time, I suggest reducing the number of maps to minimum. Also random maps theoretically take more time to create, single value maps here will work as well.

Deniwn22 (commented):
Tested locally and it works. Consider extracting the retry logic into a shared utility — we have similar code in the auth module.

src/temporal/t.rast.vi/t.rast.vi.html
```diff
@@ -0,0 +1,56 @@
...
+    <a href="https://grass.osgeo.org/grass-stable/manuals/i.vi.html">i.vi</a>,
+</em>
+
+<h2>AUTHOR</h2>
+
+Luca Delucchi, Fondazione Edmund Mach
+
+<!--
+<p><i>Last changed: $Date$</i>
+-->
```
neteler:
```suggestion
```

src/temporal/t.rast.vi/t.rast.vi.html
```diff
@@ -0,0 +1,52 @@
...
+in this case the new created maps will be added to the existing space time
+raster dataset.
+
+<h2>EXAMPLE</h2>
+
+Calculate the NDVI for your Red and NIR STDRS bands
+
+```bash
+
+```
```
echoix: Empty markdown example in HTML

src/temporal/t.rast.vi/t.rast.vi.html
```diff
@@ -0,0 +1,52 @@
...
+
+<div class="code">
+    <pre>
+t.rast.vi red=red_monthly nir=nir_monthly output=ndvi_monthly \
+    prefix=ndvimonthly
+    </pre>
+</div>
+
+If you have to calculate different index that require more info as EVI or MSAVI you
+need to set the correct parameters.
```
echoix: Different text than md. Missing ", such as", and the plural of index.

src/temporal/t.rast.vi/t.rast.vi.html
```diff
@@ -0,0 +1,52 @@
...
+```bash
+
+```
+
+<div class="code">
+    <pre>
+t.rast.vi red=red_monthly nir=nir_monthly output=ndvi_monthly \
+    prefix=ndvimonthly
+    </pre>
+</div>
```
echoix, line 21: The markdown example has `viname=ndvi` after `output=ndvi_monthly`

echoix (changes requested):
Little differences in md <-> html

echoix (approved):

petrasovaa (approved):

PR #643, opened 2021-12-14
r.learn.ml: renamed jaccard_similarity_score to jaccard_score

ncinglis:
Sci-kit learn updated Jaccard score output metric to be named "jaccard_score."


stevenpawley:
Thanks for this update, I haven't been maintaining r.learn.ml because I intended it to be replaced by r.learn.ml2.

PR #649, opened 2021-12-22
Rename tools directory to utils (grass8 branch)

neteler:
Apply directory naming from grass-code also here (https://github.com/OSGeo/grass/pull/1348):

- utils is more common for helper scripts then tools which is too close to modules.


utils/addons/grass-addons-index.sh
```diff
@@ -92,8 +92,8 @@ generate () {
 <body bgcolor=\"#FFFFFF\">
 <h2>GRASS GIS ${major} Addons Manual pages</h2>
 
-<!-- Generated from: /home/martinl/src/grass_addons/tools/addons/ -->
-<!--       See also: https://github.com/OSGeo/grass-addons/blob/master/tools/addons/README.md -->
+<!-- Generated from: /home/martinl/src/grass_addons/utils/addons/ -->
+<!--       See also: https://github.com/OSGeo/grass-addons/blob/master/utils/addons/README.md -->
```
wenzeslaus, line 96: This and other URLs are now quite outdated, but perhaps they don't need to be URLs at all when they point to files in the same repo.

wenzeslaus (approved):
Looks straightforward.

utils/addons/grass-addons-index.sh
neteler: The link is actual once this PR is merged.

neteler: Done in 51b3a84a6612528b879b1619f947afc773f3e79f

PR #650, opened 2021-12-22
Rename tools directory to utils (grass7 branch)

neteler:
Apply directory naming from grass-code also here (https://github.com/OSGeo/grass/pull/1348):

- utils is more common for helper scripts then tools which is too close to modules.


wenzeslaus (approved):
I'm not sure if this is needed, but +1.

neteler:
> I'm not sure if this is needed, but +1.

Maybe not needed but it will keep backporting with cherry-picking more simple.

PR #655, opened 2021-12-29
addons scripts: define versions as variables on top

neteler:
Avoid hardcoded version numbers throughout the script and use variables defined on top.

TODO: update `utils/addons/build-xml.py`

### DIFF
--- a/utils/addons/build-xml.py
+++ b/utils/addons/build-xml.py
@@ -1,5 +1,7 @@
 #!/usr/bin/env python
 
+# TODO 2021: do not hardcode version numbers below but use GRASS_VERSION_OLD and GRASS_VERSION_STABLE
+
 # Support Python script used by compile-xml.sh
 
 import os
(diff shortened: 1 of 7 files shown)


neteler:
@tmszi may I merge this PR?

tmszi:
> @tmszi may I merge this PR?

Are these scripts running at CTU server? Wouldn't it be better to use only the ones we have in the osgeo lxd server container? I thought these were no longer necessary, if so I would need more time for a review/test.

neteler:
> > @tmszi may I merge this PR?
> 
> Are these scripts running at CTU server?

I have no idea... I thought we use them as well.

> Wouldn't it be better to use only the ones we have in the osgeo lxd server container?

Definitely!

> I thought these were no longer necessary, if so I would need more time for a review/test.

Ok, can we delete the entire directory https://github.com/OSGeo/grass-addons/blob/grass8/utils/addons/ ?

tmszi:
> Ok, can we delete the entire directory https://github.com/OSGeo/grass-addons/blob/grass8/utils/addons/ ?

We don't use them, we can delete them. Yes I agree.

PR #661, opened 2022-01-04
v.to.rast.multi: create raster maps for multiple attributes of a vector map

ninsbl:
This new module is a simple wrapper around v.to.rast to create raster maps for multiple attributes in an efficient way. Should save both time and storage.
It works with GRASS GIS 7 as well, so backporting might be appropriate....

### DIFF
--- a/src/vector/v.to.rast.multi/v.to.rast.multi.html
+++ b/src/vector/v.to.rast.multi/v.to.rast.multi.html
@@ -27,9 +27,10 @@ can be downloaded from the
 
 <div class="code"><pre>
 # Create raster maps for RINGS_OK amd TRACT attribute of the census_wake2000 map
+# Note: RINGS_OK is of type integer, so 0 digits need to be preserved
 v.to.rast.multi --o --v input=census_wake2000 type=area \
 output=vrastmulti attribute_columns="RINGS_OK,TRACT" \
-label_columns="ID,TRACTID" memory=3000 ndigits=",4" separator=","
+label_columns="ID,TRACTID" memory=3000 ndigits="0,4" separator=","
 
 
 <h2>SEE ALSO</h2>
--- a/src/vector/v.to.rast.multi/v.to.rast.multi.html
+++ b/src/vector/v.to.rast.multi/v.to.rast.multi.html
@@ -40,3 +40,8 @@ label_columns="ID,TRACTID" memory=3000 ndigits=",4" separator=","
 
 <h2>AUTHOR</h2>
 Stefan Blumentrath, Norwegian Institute for Nature Research (NINA)
+
+<!--
+<p>
+<i>Last changed: $Date$</i>
+-->
--- a/src/vector/v.to.rast.multi/v.to.rast.multi.html
+++ b/src/vector/v.to.rast.multi/v.to.rast.multi.html
@@ -1,0 +1,42 @@
+<h2>DESCRIPTION</h2>
+<em>v.to.rast.multi</em> creates a raster map for each selected numeric column
+of the attribute table of a vector map. The purpose is to provide a simple and
+efficient way to rasterize different attributes of vector maps with multiple
+numeric value columns.
+<p>
+It is a simple wrapper around <em>v.to.rast</em> and most of the options from
+that module are available. <em>v.to.rast.multi</em> works as follows: First,
+(only) the key column is rasterized with <em>v.to.rast</em>. Then reclassification
+rules are generated for all selected attribute columns and the <b>key_column</b>
+and reclass maps are created with <em>r.reclass</em>.
+<p>
+Because <em>r.relass</em> only handles integer data and rounds data with
+floating point precision, the number of significant digits to preserve during
+reclassification can be defined with the <b>ndigits</b> option. Also text labels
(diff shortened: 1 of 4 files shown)


src/vector/v.to.rast.multi/v.to.rast.multi.py
```diff
@@ -0,0 +1,356 @@
...
+                                l=module_options["layer"],
+                                col=col,
+                            )
+                        )
+                    )
+                if at == "attribute_columns":
+                    if vmap_cols[col]["type"] not in supported_numeric:
+                        gscript.fatal(
+                            _(
+                                "Type {t} of column {col} is not a numeric type supportedby this module".format(
```
veroandreo, line 193:
```suggestion
                                "Type {t} of column {col} is not a numeric type supported by this module".format(
```

veroandreo:
I tried installing with g.extension, but fails:
```
g.extension extension=v.to.rast.multi url="https://github.com/ninsbl/grass-addons/" branch="v_to_rast_multi"
Fetching <v.to.rast.multi> from
<https://github.com/ninsbl/grass-addons/archive/v_to_rast_multi.zip> (be
patient)...
Compiling...
make: *** No targets specified and no makefile found.  Stop.
ERROR: Compilation failed, sorry. Please check above error messages.
```
I seem to recall there was a different way to install a PR aside from getting your branch locally. I cannot find it.

veroandreo:
After getting the branch locally and installing the addon, I get the foolowing using the example from the manual:
```
v.to.rast.multi --o --v input=census_wake2000 type=area output=vrastmulti attribute_columns="RINGS_OK,TRACT" label_columns="ID,TRACTID" memory=3000 ndigits=",4" separator=","
  File "/home/veroandreo/.grass8/addons/scripts/v.to.rast.multi", line 356, in <module>
    sys.exit(main())
  File "/home/veroandreo/.grass8/addons/scripts/v.to.rast.multi", line 233, in main
    list(map(int, options["ndigits"].split(","))) if options["ndigits"] else None
ValueError: invalid literal for int() with base 10: ''
```
Seems there's an extra comma there... removing it, I get: 
```
v.to.rast.multi --o --v input=census_wake2000 type=area output=vrastmulti attribute_columns="RINGS_OK,TRACT" label_columns="ID,TRACTID" memory=3000 ndigits=4 separator=","
ERROR: If the <ndigits> option is used it has to have the same length
       (number of commas) as the <attribute_columns> option.
```
In this case RINGS_OK and TRACT columns are integer, so the example should then not use the `ndigits` option, IIUC, and indeed it works fine without that.

src/vector/v.to.rast.multi/v.to.rast.multi.html
```diff
@@ -0,0 +1,47 @@
...
+
+This example is based on the North Carolina sample location (nc_spm_08_grass7) which
+can be downloaded from the
+<a href="https://grass.osgeo.org/download/data/">GRASS GIS website</a>.
+
+<div class="code"><pre>
+# Create raster maps for RINGS_OK amd TRACT attribute of the census_wake2000 map
+v.to.rast.multi --o --v input=census_wake2000 type=area \
+output=vrastmulti attribute_columns="RINGS_OK,TRACT" \
+label_columns="ID,TRACTID" memory=3000 ndigits=",4" separator=","
```
veroandreo, line 32:
```suggestion
label_columns="ID,TRACTID" memory=3000 separator=","
```

src/vector/v.to.rast.multi/v.to.rast.multi.py
ninsbl: Thanks, addressed in 365732f

src/vector/v.to.rast.multi/v.to.rast.multi.html
ninsbl: This should work now. Unittests pass.

veroandreo:
With the original command in the manual page I still get the same error(s) reported above. Only works when removing ndigits option completely.

src/vector/v.to.rast.multi/v.to.rast.multi.html
veroandreo: I get the exact same error as reported before. This conversation was resolved but which change fixed the error? I only see formatting and black commits after the error report, maybe you forgot? 
Also, I find it odd to express ndigits with a comma first. The definition says only integer, and only valid for floating point columns, which as reported earlier, RINGS_OK and TRACT are not. The example using ndigits won't work then.

ninsbl:
Thanks, @veroandreo for looking at this again! You are right, I was not thinking staight...
I remembered incorrectly that I used the example in the testsuite (which passes). But in fact, there I used 0 instead of None (no number).
I updated the example and now also handle missing values (=0) in ndigits.

TRACT is according to `v.info -c census_wake2000` double precision... census_wake2000 may not be the best dataset as example here, but hopefully still good enough...

veroandreo (approved):
now yes :)

ninsbl:
Thanks, @veroandreo for taking the time!

ninsbl:
Backported in 78d7a9d

PR #646, opened 2021-12-20
r.area.createweight manual: clarify installation of dependencies

neteler:
As a first attempt to address https://lists.osgeo.org/pipermail/grass-user/2021-December/082767.html:

- r.area.createweight manual:
  - clarify installation of dependencies
  - fix addon links
- crosslink this addon in v.area.weigh manual

TODO: suggestions welcome for MacOS and Pandas dependency.

PR #662, opened 2022-01-05
r.area.createweight manual: add macOS dependency instructions

nilason:
Follow-up to #646, adding instructions for installing dependencies with macOS app bundles.

### DIFF
--- a/src/raster/r.area.createweight/r.area.createweight.html
+++ b/src/raster/r.area.createweight/r.area.createweight.html
@@ -242,7 +242,6 @@ packages with following commands in the Terminal: <span class="code">python
 -m pip install --upgrade pip</span> and <span class="code">python -m pip
 install pandas scikit-learn</span>.
 
-
 <ul>
 <li>Pandas Wheel (mainly relevant to Windows users):
 <a href="https://pypi.python.org/pypi/pandas">https://pypi.python.org/pypi/
--- a/src/raster/r.area.createweight/r.area.createweight.html
+++ b/src/raster/r.area.createweight/r.area.createweight.html
@@ -236,6 +236,11 @@ saving the "get-pip.py" python script provided
 containing the GRASS GIS Python environment
 (GRASSFOLDER/../etc/python/grass) and executing it with administrator
 rights with the following command: <em>python get-pip.py</em>
+<br>
+For users of <b>macOS</b> GRASS-X.X.app bundles, start GRASS and install
+packages with following commands in the Terminal: <em>python &#8209;m pip
+install &#8209;&#8209;upgrade pip</em> and <em>python &#8209;m pip install
+pandas scikit&#8209;learn</em>.
 
 <ul>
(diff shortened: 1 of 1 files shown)


nilason:
@neteler Is the use of `&#8209;` for non-breaking hyphen acceptable? It works for me, but...

tgrippa (approved):

neteler:
> > @neteler Is the use of `&#8209;` for non-breaking hyphen acceptable? It works for me, but...
> 
> I realised there is a css style for inline code snippets. I'll use that instead of non-breaking hyphens.

Is that anything which should go into https://trac.osgeo.org/grass/wiki/Submitting/Docs?

nilason:
> > > @neteler Is the use of `&#8209;` for non-breaking hyphen acceptable? It works for me, but...
> > 
> > 
> > I realised there is a css style for inline code snippets. I'll use that instead of non-breaking hyphens.
> 
> Is that anything which should go into https://trac.osgeo.org/grass/wiki/Submitting/Docs?

Probably the inline element ‘code’ would be better choice (than `<span class="code">`), it seems that it is also supported by g.html2man (not tested). Presumably `code` will need some css styling too.

nilason:
I settled on `<code>` and I'm ready for merge, any objections?

tgrippa:
No objection on my side but @neteler did not reviewed the change yet :)

src/raster/r.area.createweight/r.area.createweight.html
```diff
@@ -224,18 +224,23 @@ <h4>Python libraries</h4>
...
 <a href="https://bootstrap.pypa.io/get-pip.py">here</a> in the folder
 containing the GRASS GIS Python environment
 (GRASSFOLDER/../etc/python/grass) and executing it with administrator
-rights with the following command: <em>python get-pip.py</em>
+rights with the following command: <code>python get-pip.py</code>
+<br>
+For users of <b>macOS</b> GRASS-X.X.app bundles, start GRASS and install
+packages with following commands in the Terminal: <code>python
+-m pip install --upgrade pip</code> and <code>python -m pip
+install pandas scikit-learn</code>.
```
neteler, line 243:
```suggestion
packages with following commands in the Terminal:<br>
<code>python -m pip install --upgrade pip</code> and<br>
<code>python -m pip install pandas scikit-learn</code> .
```

I suggest to break the lines here for better readability.

neteler (approved):

src/raster/r.area.createweight/r.area.createweight.html
nilason: I don't think it's necessary to break for the Mac commands only, there are also Win commands.

This is what it will look like as is:



with your suggestion:



I plan to suggest in another PR (in core) css changes to the effect of:

neteler: OK, fine for me.

nilason:
Thanks!

PR #682, opened 2022-01-21
cronjob: cleanup file names, add grass8-devel

neteler:
- added cronjob to generate "grass8-devel source" snapshot (missing so far)
- cleaned up (renaming) file names to better reflect what they actually do
- updated README.md
- remove unused old script `grass-addons-fetch-xml.sh`
- updated cronjob table (needs to be deployed on server after PR merge)


tmszi (approved):
Looks good to me. With this PR we can delete this script file https://github.com/OSGeo/grass-addons/blob/grass8/utils/cronjobs_osgeo_lxd/grass-addons-fetch-xml.sh which is no needed anymore.

neteler:
> Looks good to me. With this PR we can delete this script file https://github.com/OSGeo/grass-addons/blob/grass8/utils/cronjobs_osgeo_lxd/grass-addons-fetch-xml.sh which is no needed anymore.

Yes, done in 5b7f601

neteler:
Updated cronjobs are now deployed on server.

PR #701, opened 2022-02-20
v.what.rast.label manual: minor reformulation

ecodiv:
reformulation few lines in helpfile

### DIFF
--- a/src/vector/v.what.rast.label/v.what.rast.label.html
+++ b/src/vector/v.what.rast.label/v.what.rast.label.html
@@ -10,8 +10,8 @@ which the label is taken is used.
 <p>Optionally, the user can define raster layers without labels. In that case
 only the raster values are uploaded to a column with the name of that raster. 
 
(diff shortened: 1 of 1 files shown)


src/vector/v.what.rast.label/v.what.rast.label.html
```diff
@@ -10,8 +10,8 @@ <h2>DESCRIPTION</h2>
 <p>Optionally, the user can define raster layers without labels. In that case
 only the raster values are uploaded to a column with the name of that raster. 
 
-<p>Except the columns with raster values, the attribute table of the
-output vector point map can optionally include columns with coordinates.
+<p>Another option is to add columns with the point coordinates to the attribute 
+tabel of the output vector layer.
```
veroandreo, line 14:
```suggestion
<p>Another option is to add columns with point coordinates to the attribute 
table of the output vector layer.
```

PR #709, opened 2022-03-01
v.tin.to.rast GRASS 8 compatibility

kikislater:
update GRASS 8 compatibility


neteler:
How about removing the test entirely?

For a similar check removal, see #707 (i.modis.import)

kikislater:
Yes, grass version check is useless. Not the same structure but easy to remove ... updated !

PR #713, opened 2022-03-15
i.fusion.hpf: fix "Illegal filename <.hpf>" warning

neteler:
Fixes (when `suffix` is not used) confusing warning:

`WARNING: Illegal filename <.hpf>. Cannot start with '.' or be 'NULL'.`

(In the parser `answer` no trailing dot may be used).


neteler:
Backported to grass7 branch in e0d664f.

NikosAlexandris:
Thank you @neteler!

PR #718, opened 2022-03-17
G78 cronjobs: add "new version G8" box into man pages

neteler:
This PR injects a red box into the [G78 manual pages](https://grass.osgeo.org/grass78/manuals/), pointing to the respective G80 manual page.

image

This is added to all G78 manual pages.

In addition, it adds minor version number to addon manual landing page title to reduce confusion.


veroandreo (approved):

wenzeslaus (commented):
I think the first "here" is just confusing. You searched online for a doc, got old one and are redirected to download? The download link can be the word "available" if that's even needed. See e.g. what the page for Python 3.6 does (no link to downloads, only to the new doc):

https://docs.python.org/3.6/library/datetime.html

neteler:
Thanks for the suggestion, @wenzeslaus ! Implemented in this PR as follows:

image
(https://grass.osgeo.org/grass78/manuals/v.generalize.html)

PR #719, opened 2022-03-17
cronjobs: add GRASS GIS main branch binary build job

neteler:
- adds G8.dev binary cronjob which creates the manual pages as well: 
https://grass.osgeo.org/grass-devel/manuals/index.html

Note: Once merged, it requires deploment on server with `crontab ...`.


neteler:
> Note: Once merged, it requires deploment on server with `crontab ...`.

Done on grass.osgeo.org.

PR #722, opened 2022-03-29
r.random.weight: fix string.replace bug

Giordano8:
string has no module replace, so I changed it in order to do a proper string replacement


Giordano8:
Is it possible to merge this request?

neteler:
Merging, thanks.

PR #730, opened 2022-04-20
r.viewshed.exposure manual: fix HTML

neteler:
- fix wrong tag
- HTML cosmetics

(found via inspection of https://grass.osgeo.org/addons/grass8/logs/)


ninsbl (approved):
Thanks for catching this!

PR #731, opened 2022-04-20
G78 addon cronjob: add "new version G8" box into addon man pages

neteler:
This PR injects a red box into the G78 addon manual pages, pointing to the respective G80 addon manual page.

It completes #718.

Random example:

image


veroandreo:
Ops, sorry for my delay in reviewing, I had it marked for the weekend. I had a comment/question. The message says that grass78 line is no longer supported but I believe this is not the case, right? So, we might better say that support will be soon discontinued or something on that line. What do you think?

neteler:
> So, we might better say that support will be soon discontinued or something on that line.

Yes, I'll change it accordingly in a new PR.

PR #743, opened 2022-04-30
new addon r.boxplot

ecodiv:
Addon to create boxplot of raster, for whole raster of per zone of zonal raster


wenzeslaus:
This is a great tool!

I think the name should be different. Although it deals with visualization, it is not a display command, so perhaps `r` prefix. The `d` prefix is reserved for the tools which are using the system of display drivers and monitors and can be combined with each other. This tool is separate from that. In that context, it is similar to tool such as [m.nviz.image](https://grass.osgeo.org/grass82/manuals/m.nviz.image.html) or [r.scatterplot](https://grass.osgeo.org/grass82/manuals/addons/r.scatterplot.html). On the other hand, there is no standard for naming visualization tools and thinking about this one as display (`d.`) command in general is not wrong, so I'm wondering if there is a better way of naming things here.

ecodiv:
Good point. I used this name in line with the names of d.vect.colhist and d.vect.colbp add-ons. I have no preference, but I think there should be some consistency.

ecodiv:
Having thought it over a bit more, I tend to agree that using a name `r.boxplot` is more straightforward. @wenzeslaus, what is the best way to change this? Removing this pull request and doing a new one?

@mlennert, a similar argument could be made for d.vect.colhist and d.vect.colbp? What do you think? Would renaming them to v.colhist (or v.histogram) and v.colbp (or v.boxplot) be something to consider?

wenzeslaus:
`git mv old_file new_file` is enough, no need for new branch or PR. (Alternatively, renaming the files (and directories) will make `git status` show you what are the changes and then you `git add` them as `git status` suggests.)

ninsbl:
Interesting discussion.

I wonder if it would`nt help users to find dedicated plotting functions if that would be indicated more systematically in the module name. Either as a new group of commands (e.g. starting with `p.`, like `p.rast.boxplot`) or sub-group (e.g. `r.plot.poxplot`, `v.plot.*` ...). That would also help grouping such functions (also new, future ones).

Just thinking aloud...

That said, I think the technical concept behind the `d` prefix is (even) less prominent in the heads of users than it is in the heads of developers. So, my guess is that there is a bit of a trade-off regarding technical consistency vs. UX consistency...

ecodiv:
> 

I'll think I rename it for now to the proposed naming above. If at some point, the decision is made for a specific and above all consistent naming convention, it may be anyway better to do that for all concerning modules at once.

mlennert:
> Having thought it over a bit more, I tend to agree that using a name `r.boxplot` is more straightforward. @wenzeslaus, what is the best way to change this? Removing this pull request and doing a new one?
> 
> @mlennert, a similar argument could be made for d.vect.colhist and d.vect.colbp? What do you think? Would renaming them to v.colhist (or v.histogram) and v.colbp (or v.boxplot) be something to consider?
> 
> 

I don't have a strong opinion on this. It really depends on whether you think that the command is first about displaying something or first and foremost about analyzing a type of map. I used d. as I felt the former to be more important as plotting is about displaying, but I can understand the argument that boxplots and histograms are as much, if not more, about analysis.

So feel free to rename any of these modules as you see fit.

wenzeslaus:
See the notebook ([Binder](https://mybinder.org/v2/gh/OSGeo/grass/8.2.0RC1?urlpath=lab%2Ftree%2Fdoc%2Fnotebooks%2Fbasic_example_grass_jupyter.ipynb), [GitHub](https://github.com/OSGeo/grass/blob/main/doc/notebooks/basic_example_grass_jupyter.ipynb)) or [doc](https://grass.osgeo.org/grass82/manuals/libpython/grass.jupyter.html) for grass.jupyter, an object of Map will render true display commands (one or more) into an image which you can either show in the notebook or save with a unified interface. Similarly with _d.mon_ in command line where the image goes, e.g., to `wx0` window or a PNG image. Using _d.vect.colbp_ won't have the promised effect.

The display command and driver system is more specific than visualization (regardless of how we define the terms for ourselves in other contexts). Its equivalent outside of GRASS GIS is, for example, Matplotlib, gnuplot or GMT. In GRASS GIS, it is parallel to _m.nviz.image_ (visualization in 3D) or _ps.map_ (cartography outputs in PostScript).

This should be definitively communicated in a better way. Suggestions welcome.

src/raster/r.boxplot/r.boxplot.html
```diff
@@ -0,0 +1,107 @@
+<h2>DESCRIPTION</h2>
+
+<p>
+<em>r.boxplot</em> draws boxplots of the raster values of a input raster
+layer. The user has the option to define a zonal (categorical) layer. In that
+case, a boxplot will be drawn for each zone, using the values of the input
+raster that fall within that zone.
+
+<p>
+By default, the resulting plot is displayed on screen (default).
```
veroandreo, line 10:
```suggestion
By default, the resulting plot is displayed on screen. 
```
Already said :)

src/raster/r.boxplot/r.boxplot.html
```diff
@@ -0,0 +1,107 @@
...
+be saved as a png file.
+
+<p>
+The whiskers extend to the most extreme data point, which is no more than
+<i>range</i> &#10005; the interquartile range (iqr) from the box. By default,
+a <i>range</i> of 1.5 is used, but the user can change this. Note that range
+values need to be larger than 0.
+
+<p>
+By default, outliers are not included in the plot. Set to -o flag to include
```
veroandreo, line 23:
```suggestion
By default, outliers are not included in the plot. Set the -o flag to include
```

src/raster/r.boxplot/r.boxplot.html
```diff
@@ -0,0 +1,107 @@
...
+them in the plot. To create a point layer with the locations of the outliers,
```
veroandreo, line 24: maybe replacing layer by vector maps makes it clearer that the output is a map and not something else added to the plot

src/raster/r.boxplot/r.boxplot.html
```diff
@@ -0,0 +1,107 @@
...
+
+<p>
+There are a few layout options, including the option to rotate 
+the plot and the x-axis labels, print the boxplot(s) with notches and
+sort the boxplot from low to high (ascending) or from high to low (descending)
+median.
+
+<p>
+If the user opts to have the outliers drawn, a vector layer will be created
+with the locations of the outliers.
```
veroandreo, line 35: IIUC, this is different than what is said above, where it is stated that to get the vector layer, the user needs to set the `map_outliers` option

src/raster/r.boxplot/r.boxplot.py
```diff
@@ -0,0 +1,633 @@
...
+
+try:
+    import matplotlib
+
+    matplotlib.use("WXAgg")
+    from matplotlib import pyplot as plt
+except ImportError as e:
+    raise ImportError(
+        _(
+            'r.boxplot  needs the "matplotlib" '
```
veroandreo, line 141:
```suggestion
            'r.boxplot needs the "matplotlib" '
```
This try/except might go as lazy import after calling the parser, right? It was my understanding that lazy import is good practice.

src/raster/r.boxplot/r.boxplot.html
```diff
@@ -0,0 +1,107 @@
...
+<a href="https://grass.osgeo.org/download/data/">NC sample 
+dataset</a>. Set the -o flag to include outliers. Use order=ascending 
+to order the boxplots from low to high median. Provide a name for the outlier 
+map to save the outlier locations as a point layer.
+
+<div class="code"><pre>
+r.boxplot -o order=ascending input=elevation zones=landuse map_outliers="outliers"
+</pre>
+</div>
+
```
veroandreo, line 104: Are there any `SEE ALSO` modules or links?

veroandreo (commented):
Cool module! Thanks @ecodiv! Some minor stuff here and there

src/raster/r.boxplot/r.boxplot.py
ecodiv: > This try/except might go as lazy import after calling the parser, right? It was my understanding that lazy import is good practice.

@veroandreo changed it, can you check?

PR #748, opened 2022-05-08
r.boxplot documentation bugfix

ecodiv:
Linux compilation failed (https://grass.osgeo.org/addons/grass8/logs/), not sure this empty heading has anything to do with it. Otherwise, @neteler do you have an idea why it is failing?


neteler:
There is a leftover `</em>` at

https://github.com/OSGeo/grass-addons/blob/257e8c9d633b75e7e7e3b250826529f54da62018/src/raster/r.boxplot/r.boxplot.html#L92

ecodiv:
Ah, great, thanks! That is a handy tool, will keep that available in my toolbox.

PR #750, opened 2022-05-12
d.explanation.plot: Customize raster count, operators, fonts

wenzeslaus:
Now plots can be generated to explain a + b -> c operations as well as a -> b and a + b + c -> d. Operators and fonts can be customized. Placement is improved and automated, although rectangular image is still required. New example uses Map from grass.jupyter.


wenzeslaus:
Output of the previous version of d.explanation.plot is used in [r.patch](https://grass.osgeo.org/grass80/manuals/r.patch.html) documentation to explain the rules of patching:

Screenshot from 2022-05-12 00-35-34

This PR makes it usable to a wider range of modules.

PR #751, opened 2022-05-16
t.rast.patch: allow use of virtual format

ninsbl:
Using the virtual raster format (`r.buildvrt`) should provide a reduction in both processing time and storage usage.
For some workflows a virtual raster may be good enough.
I have not yet fully tested the modifications though...

### DIFF
--- a/src/temporal/t.rast.patch/t.rast.patch.html
+++ b/src/temporal/t.rast.patch/t.rast.patch.html
@@ -1,6 +1,14 @@
 <h2>DESCRIPTION</h2>
-This module patches raster maps that have gaps in time with subsequent maps (within a space time raster dataset)
-using <em>r.patch</em>. Hence it is a wrapper for <em>r.patch</em> in the temporal domain.
+This module patches raster maps that have gaps in time with subsequent maps (within a space time raster dataset) using <em>r.patch</em> or <b>r.buildvrt</b>. Hence it is a wrapper for those two modules in the temporal domain.
+<p>
+By default <em>r.patch</em> is used to create a patched raster map.
+Especially for temporary data, using <b>r.buildvrt</b> for patching
+can be advantageous with regards to processing time and storage space.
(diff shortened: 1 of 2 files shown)


ninsbl:
This is a minor change in terms of code logic. But I like to have at least one other person looking at it before I push the change to an existing module...

veroandreo:
This is what I get when running the manual page example (no errors, all good, but display of data looks odd):

image

There are data everywhere, just the color palette seems weird... is this related to vrt? If I manually set it to `grey` then it looks fine:

image

ninsbl:
> 
> Would it be possible to generate binder links to test? I'm also willing to learn how to prepare such links, they make testing so much easier and amenable

Thanks for testing!

A binder link has to use the core repo, I think, like this one:

https://mybinder.org/v2/gh/OSGeo/grass/main?urlpath=lab%2Ftree%2Fdoc%2Fnotebooks%2Fjupyter_example.ipynb

Unfortunately, there is no IPython Notebook in the PR that could be used directly. Maybe it could be generated from a testsuite if available. But I am not sure how to provide testing with binder in the most convenient way. Maybe @wenzeslaus has some suggestions?

ninsbl:
> There are data everywhere, just the color palette seems weird... is this related to vrt? If I manually set it to `grey` then it looks fine:

Yes, good catch! I added a note in the manual. @metzm , is it correct that with r.buildvrt, the color map is taken from the first input raster map, or did I get that wrong?

ninsbl:
I am not too familiar with C, but if I read the code here correctly:
https://github.com/OSGeo/grass/blob/82fb7b8e93a8b87f7de827c0b1f98992759ce7a1/raster/r.buildvrt/link.c#L136
my assumption seems validated that the color map is taken from the first map.

Therefore, I take the liberty to merge this PR, pleas file a bug report if anything is missing og picked up incorrectly...

PR #767, opened 2022-07-22
r.area.createweight: fix lazy import py packages

tmszi:
Due compilation error on OS MS Windows (missing some py packages). Fixes https://github.com/OSGeo/grass/issues/2492.

### DIFF
--- a/src/raster/r.area.createweight/r.area.createweight.py
+++ b/src/raster/r.area.createweight/r.area.createweight.py
@@ -202,23 +202,6 @@ import matplotlib.pyplot as plt
 from ast import literal_eval
 from copy import deepcopy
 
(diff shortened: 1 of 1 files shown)


neteler:
Tested on Linux like this:

```
cd ~/software/grass-addons/
# get patch from gitHub and apply locally
wget https://github.com/OSGeo/grass-addons/pull/767.diff
patch -p1 < 767.diff

# test installation of patched version
grass78 ~/grassdata/nc_spm_08_grass7/user1 --exec g.extension r.area.createweight url=src/raster/r.area.createweight/
...
Installation of <r.area.createweight> successfully finished

# revert local path
patch -p1 --reverse < 767.diff
```

Thanks for the quick fix.

neteler (approved):

PR #769, opened 2022-07-28
i.sentinel.import: allow importing for lat/lon locations

griembauer:
Using `i.sentinel.import` in a lat/lon location (or any location that does not have meters as projection units) fails as the `resolution_value` parameter for `r.import` is read from the band metadata (which is in metres, e.g. `10`, `20`, `60`), so a final reprojected map will be empty. 
This PR adapts `i.sentinel.import` such that it only applies the `resolution_value` parameter to `r.import` if the location units are `meters`


metzm (approved):
Looks good to me, thanks for the contribution!

griembauer:
Thanks for your review, I will merge now.

PR #774, opened 2022-08-14
r.learn.ml2: fix r.category call in r.learn.predict

neteler:
Avoid crash when category labels contain a comma as e.g. in US NLCD Land cover product

image

... by changing the separator from comma to pipe character.

(see https://grasswiki.osgeo.org/wiki/NLCD_Land_Cover#Legend)


neteler:
I take liberty to merge this.

wenzeslaus:
+1 for the merge. The ultimate fix is something like JSON or real CSV output from r.category (and a related real parser of that format), but that goes for all these cases.

PR #786, opened 2022-08-30
i.sentinel.parallel.download: allow download from GCS with scenename specifications and without credentials

griembauer:
This PR allows `i.sentinel.parallel.download` to directly download from GCS if the sentinel-2 scene names are specified via the `-s `flag and if **no settings file is provided**. This specific case is already possible in `i.sentinel.download` but so far the `settings` parameter was required in `i.sentinel.parallel.download`

### DIFF
--- a/src/imagery/i.sentinel/i.sentinel.parallel.download/i.sentinel.parallel.download.py
+++ b/src/imagery/i.sentinel/i.sentinel.parallel.download/i.sentinel.parallel.download.py
@@ -166,13 +166,6 @@ def scenename_split(scenename, datasource, esa_name_for_usgs=False):
         start_day = start_day_dt.strftime("%Y-%m-%d")
         end_day = end_day_dt.strftime("%Y-%m-%d")
         query_string = f"identifier={scenename.replace('.SAFE', '')}"
-        # if datasource == "GCS":
-        #     query_string = f"identifier={scenename.replace('.SAFE', '')}"
-        # else:
-        #     # get query string
-        #     if not scenename.endswith(".SAFE"):
-        #         scenename = scenename + ".SAFE"
-        #     query_string = "filename={}".format(scenename)
     else:
         # when usgs downloads via identifier, start/end are ignored
         producttype = "S2MSI1C"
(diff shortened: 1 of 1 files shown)


src/imagery/i.sentinel/i.sentinel.parallel.download/i.sentinel.parallel.download.py
```diff
@@ -164,10 +165,13 @@ def scenename_split(scenename, datasource, esa_name_for_usgs=False):
         end_day_dt = dt_obj + timedelta(days=1)
         start_day = start_day_dt.strftime("%Y-%m-%d")
         end_day = end_day_dt.strftime("%Y-%m-%d")
-        # get query string
-        if not scenename.endswith(".SAFE"):
-            scenename = scenename + ".SAFE"
-        query_string = "filename={}".format(scenename)
+        if datasource == "GCS":
+            query_string = f"identifier={scenename.replace('.SAFE', '')}"
```
metzm, line 169:
```suggestion
            query_string = f"identifier={scenename.removesuffix('.SAFE')}"
```
Here you remove the suffix `.SAFE` ...

griembauer: `removesuffix` was introduced in Python 3.9 - not sure if we can expect this?

src/imagery/i.sentinel/i.sentinel.parallel.download/i.sentinel.parallel.download.py
```diff
@@ -164,10 +165,13 @@ def scenename_split(scenename, datasource, esa_name_for_usgs=False):
...
+        else:
+            # get query string
+            if not scenename.endswith(".SAFE"):
+                scenename = scenename + ".SAFE"
```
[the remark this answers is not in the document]
metzm: Ah ok, then I misunderstood the manual:
"Querying and filtering data on Google Cloud Storage requires the fee-based usage of [BigQuery](https://cloud.google.com/bigquery/external-data-cloud-storage?hl=en), which is currently not supported by i.sentinel.download. Instead, when using the datasource=GCS option, querying is performed using the [Copernicus Open Access Hub](https://scihub.copernicus.eu/) interface, while only the download step makes use of Google Cloud Storage."

src/imagery/i.sentinel/i.sentinel.parallel.download/i.sentinel.parallel.download.py
metzm: Indeed, I missed this. I would not expect/require Python 3.9, forget about the suggestion.

metzm (approved):
All fine with me, thanks!

PR #804, opened 2022-10-04
r.in.usgs: Use parallel for patching, cache tolerance

wenzeslaus:
* Use nprocs and memory for r.patch with >=8.2.
* Make the cache tolerance configurable and increase the default value.


petrasovaa (approved):
Please add parallel keyword.

wenzeslaus:
Parallel keyword added in #809 which is more related.

PR #805, opened 2022-10-04
i.sentinel.parallel.download: Fix CPU count and options access

wenzeslaus:
CPU count function and data source option were not referenced correctly.


wenzeslaus:
@griembauer Can you please review this or recommend someone else?

griembauer (approved):

griembauer:
Thanks for catching these!

PR #806, opened 2022-10-04
.github/workflows: fix `MODULE_TOPDIR` ENV var path value

### LINKED ISSUE #803: [Bug] Addon manuals don`t build in CI
**all addons**
During github workflows, it seems that manuals are not build:
https://github.com/OSGeo/grass-addons/actions/runs/3172752971/jobs/5167576298#step:13:41

**Describe the bug**
During build of addons in CI the following error message pops up:
`FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/install/grass*'`

**To Reproduce**
See: https://github.com/OSGeo/grass-addons/actions/runs/3172752971/jobs/5167576298#step:13:41

**Expected behavior**
Addon manuals should build in CI

- Operating System: Ubuntu Linux
- GRASS GIS version 8.3.dev

tmszi:
Python code doesn't interpret wildcard `*` inside path. Fixes #803.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,7 +87,7 @@ jobs:
     - name: Build addons
       run: |
         cd grass-addons/src
-        GRASS_INSTALL=$HOME/install/grass*
+        GRASS_INSTALL=$($HOME/install/bin/grass --config | sed -n '4{p;q}')
         make MODULE_TOPDIR=$GRASS_INSTALL
 
     - name: Get extra Python dependencies


.github/workflows/ci.yml
```diff
@@ -87,7 +87,7 @@ jobs:
     - name: Build addons
       run: |
         cd grass-addons/src
-        GRASS_INSTALL=$HOME/install/grass*
+        GRASS_INSTALL=$HOME/install/grass
```
ninsbl: For the current development version the installation base directory is grass83, thus CI is still failing.

The approach below should catch that (works in my fork)...

```suggestion
        GRASS_INSTALL=$($HOME/install/bin/grass --config | sed -n '4{p;q}')
```

ninsbl (changes requested):
See my comment in the code

.github/workflows/ci.yml
ninsbl: Instead of hard-coding GRASS versions, we could take it from grass --config (see above)...

tmszi: Yes you are right. Thanks for hint!

ninsbl (approved):
Seems to work now.

PR #822, opened 2022-10-19
r.in.nasadem: catch http error

### LINKED ISSUE #821: [Bug] r.in.nasadem runs with wrong password
**Name of the addon**
r.in.nasadem

**Describe the bug**
r.in.nasadem starts running with wrong password but gives an error for other reasons so that it is not apparent that the error occured because of the wrong password.

**To Reproduce**
1. Run r.in.nasadem user=USER password=PASSWORD ...  but insert a wrong password.
2. See error

**Expected behavior**
r.in.nasasdem does not run with wrong password input / produces error because of wrong password 

**System description:**
- Operating System: Linux
- GRASS GIS version 8.2.1

griembauer:
This PR fixes #821:

- Any connection error for the download of nasadem tiles is now properly caught and reported using the generic [urllib.error.URLError](https://docs.python.org/3/library/urllib.error.html)
- A bug in the cleanup function is fixed: The region is saved in the original location/mapset, but at the beginning of the cleanup function, we may still be in the temp location. This resulted in errors of `g.remove type=region...`

### DIFF
--- a/src/raster/r.in.nasadem/r.in.nasadem.py
+++ b/src/raster/r.in.nasadem/r.in.nasadem.py
@@ -301,9 +301,10 @@ def download_tile(tile, url, pid, version, username, password):
         time.sleep(0.5)
     except urllib.error.URLError as err:
         grass.fatal(
-            _("Download of tile {0} from URL {1} was not successful:\n{2}").format(
-                local_tile, remote_tile, err
-            )
+            _(
+                "Download of tile {local_tile} from URL {remote_tile} was not"
+                "successful:\n{err}"
+            ).format(local_tile=local_tile, remote_tile=remote_tile, err=err)
         )
 
     return goturl
--- a/src/raster/r.in.nasadem/r.in.nasadem.py
+++ b/src/raster/r.in.nasadem/r.in.nasadem.py
@@ -301,9 +301,8 @@ def download_tile(tile, url, pid, version, username, password):
         time.sleep(0.5)
     except urllib.error.URLError as err:
         grass.fatal(
-            _(
-                f"Download of tile {local_tile} from URL {remote_tile} was not "
-                f"successful:\n{err}"
+            _("Download of tile {0} from URL {1} was not successful:\n{2}").format(
+                local_tile, remote_tile, err
             )
         )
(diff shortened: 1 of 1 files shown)


neteler (approved):
Tested successfully, thanks!

src/raster/r.in.nasadem/r.in.nasadem.py
```diff
@@ -294,14 +295,13 @@ def download_tile(tile, url, pid, version, username, password):
...
-
         fo = open(local_tile, "w+b")
         fo.write(response.read())
         fo.close
         time.sleep(0.5)
-    except:
-        goturl = 0
-        pass
+    except urllib.error.URLError as err:
+        grass.fatal(_(f"Download of tile {local_tile} from URL {remote_tile} was not "
```
tmszi: Use `.format()` method instead of f-string (In particular, Python [f-strings](https://docs.python.org/3/reference/lexical_analysis.html#f-strings) are not yet supported by `xgettext`.) for translation string, please.

griembauer: Thanks, I updated it accordingly

src/raster/r.in.nasadem/r.in.nasadem.py
```diff
@@ -301,9 +301,8 @@ def download_tile(tile, url, pid, version, username, password):
         time.sleep(0.5)
     except urllib.error.URLError as err:
         grass.fatal(
-            _(
-                f"Download of tile {local_tile} from URL {remote_tile} was not "
-                f"successful:\n{err}"
+            _("Download of tile {0} from URL {1} was not successful:\n{2}").format(
```
tmszi, line 304: Cosmetic change: Ordering string interpolation is no needed

`"Download of tile {} from URL {} was not successful:\n{}"` work as expected

but better if number of string interpolation > 1 (your case is 3), use variable substitutions by name inside `{}`

```
_("Download of tile {local_tile} from URL {remote_tile} was not successful:\n{err}").format(
    local_tile=local_tile, remote_tile=remote_tile, err=err,
)
```

PR #824, opened 2022-10-22
i.ann.maskrcnn: remove deprecated distutils

ninsbl:
See also: https://github.com/OSGeo/grass/issues/2601

### DIFF
--- a/src/imagery/i.ann.maskrcnn/maskrcnnlib/model.py
+++ b/src/imagery/i.ann.maskrcnn/maskrcnnlib/model.py
@@ -29,10 +29,15 @@ import keras.models as KM
 import utils
 
 # Requires TensorFlow 1.3+ and Keras 2.0.8+.
-from distutils.version import LooseVersion
-
-assert LooseVersion(tf.__version__) >= LooseVersion("1.3")
-assert LooseVersion(keras.__version__) >= LooseVersion("2.0.8")
+version_pattern = "([0-9]{1,3}).([0-9]{1,4})"
+assert list(map(int, re.match(version_pattern, tf.__version__).groups())) >= [1, 3]
+version_pattern = "([0-9]{1,3}).([0-9]{1,4}).([0-9]{1,4})"
+assert list(
+    map(
+        int,
+        re.match(version_pattern, keras.__version__).groups(),
+    )
+) >= [2, 0, 8]
 
 
 ############################################################
--- a/src/imagery/i.ann.maskrcnn/maskrcnnlib/utils.py
+++ b/src/imagery/i.ann.maskrcnn/maskrcnnlib/utils.py
@@ -24,7 +24,6 @@ import skimage.transform
 import urllib.request
 import glob
 import warnings
-from distutils.version import LooseVersion
 
 # URL from which to download the latest COCO trained weights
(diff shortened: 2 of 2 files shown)


src/imagery/i.ann.maskrcnn/maskrcnnlib/model.py
```diff
@@ -29,10 +29,15 @@
 import utils
 
 # Requires TensorFlow 1.3+ and Keras 2.0.8+.
-from distutils.version import LooseVersion
-
-assert LooseVersion(tf.__version__) >= LooseVersion("1.3")
-assert LooseVersion(keras.__version__) >= LooseVersion("2.0.8")
+version_pattern = "([0-9]{1,3}).([0-9]{1,4})"
```
pesekon2, line 32:
```suggestion
version_pattern = "([0-9]{1}).([0-9]{1,2})"
```

src/imagery/i.ann.maskrcnn/maskrcnnlib/model.py
```diff
@@ -29,10 +29,15 @@
...
+assert list(map(int, re.match(version_pattern, tf.__version__).groups())) >= [1, 3]
+version_pattern = "([0-9]{1,3}).([0-9]{1,4}).([0-9]{1,4})"
```
pesekon2, line 34:
```suggestion
version_pattern = "([0-9]{1}).([0-9]{1,2}).([0-9]{1,2})"
```

src/imagery/i.ann.maskrcnn/maskrcnnlib/utils.py
```diff
@@ -979,7 +978,9 @@ def resize(
     of skimage. This solves the problem by using different parameters per
     version. And it provides a central place to control resizing defaults.
     """
-    if LooseVersion(skimage.__version__) >= LooseVersion("0.14"):
+    if list(
+        map(int, re.match("([0-9]{1,3}).([0-9]{1,4})", skimage.__version__).groups())
```
pesekon2, line 982:
```suggestion
        map(int, re.match("([0-9]{1}).([0-9]{1,2})", skimage.__version__).groups())
```

src/imagery/i.ann.maskrcnn/maskrcnnlib/utils.py
```diff
@@ -24,7 +24,6 @@
 import urllib.request
 import glob
 import warnings
```
pesekon2, line 26:
```suggestion
import re
import warnings
```

pesekon2 (changes requested):
* add missing `import re`
* remove unnecessarily big sizes of regex strings (first number just one int, then two ints)

src/imagery/i.ann.maskrcnn/maskrcnnlib/model.py
ninsbl: I agree that the likelyhood that broad version ranges are needed is low. However, it should not hurt either and with wider ranges you are less vulnerable. But I have nothing against reducing the range if you prefer that...

pesekon2: Okay. In the end, I have no strong opinion on that. So let's keep the broader ranges. Approving the PR.

pesekon2 (approved):

ninsbl:
Thanks for reviewing, @pesekon2

PR #819, opened 2022-10-17
g.citation: fix module output parameter

tmszi:
Fix module `output` parameter.

**Current behavior**

```bash
GRASS nc_spm_08_grass7/landsat:~ > g.citation v.surf.rst output=/tmp/result

GRASS nc_spm_08_grass7/landsat:~ > file /tmp/result
/tmp/result: cannot open `/tmp/result' (No such file or directory)
```

**Expected behavior:**

```bash
GRASS nc_spm_08_grass7/landsat:~ > g.citation v.surf.rst output=/tmp/result 

GRASS nc_spm_08_grass7/landsat:~ > cat /tmp/result 
@misc{v_surf_rst,
  title = {{GRASS GIS: v.surf.rst module}},
  author = {Lubos Mitas and Helena Mitasova and Irina Kosinovsky and Darrel McCauley and Jaroslav Hofierka and Stanislav Zubal and Michal Lacko},
  howpublished = {https://github.com/OSGeo/grass/tree/main/vector/v.surf.rst},
  year = {2022}
  note = {Accessed: 2022-10-16T07:02:16},
}

```

PR #825, opened 2022-10-23
.github/workflows: remove build old stable GRASS GIS 8.0 version from the CI

tmszi:
See https://github.com/OSGeo/grass-addons/pull/819#issuecomment-1287415722.


ninsbl (approved):
Looks good, as removing the test is "PSC approved". See:

https://github.com/OSGeo/grass-addons/pull/819#issuecomment-1287415722

PR #828, opened 2022-10-24
SEO GRASS GIS 7 manual: properly inject canonical

neteler:
- Since the canonical was overwritten during addon compilation, a second (modified) injection in the cronjob
- shift timings of cronjobs to avoid overlaps


wenzeslaus:
Overlapping jobs: Just to be clear, the issue is that they actually not independent (e.g., use the same directories) or is it sever performance?

neteler:
The latter: we must not overload the virtual machine.

PR #871, opened 2023-02-24
CI: do not configure with PDAL

nilason:
New settings on GRASS main with pkg-config means `--with-pdal` is default.
This sets grass-addons CI to the previous state, with explicit `--without-pdal`.

### DIFF
--- a/.github/workflows/build_grass.sh
+++ b/.github/workflows/build_grass.sh
@@ -36,6 +36,7 @@ export INSTALL_PREFIX=$1
     --with-geos \
     --with-sqlite \
     --with-fftw \
+    --without-pdal \
     --with-netcdf
 
 make


neteler:
@nilason Any objections to bring PDAL support back to addons CI?

Its absence breaks `./raster/r.in.pdal/testsuite/test_base_resolution.sh`, see e.g. https://github.com/neteler/grass-addons/actions/runs/12647834206/job/35240965201 (at bottom of the failed tests).

nilason:
This didn’t change anything regarding using or not of PDAL, only adjusted to different behaviour of configure. The problem is perhaps related, but this in itself is not the cause of current issue.

marisn:
Is there anything of value provided by r.in.pdal add-on over the r.in.pdal from the core?

neteler:
The core module should be used.

Issue #872, opened 2023-02-27
[Bug] r.learn.train: depricated numpy.object


anikaweinmann:
**Name of the addon**
r.learn.ml2/r.learn.train

**Describe the bug**
With numpy > 1.24.0 the numpy object is depricated:
```
r.learn.train group=lsat7_2000 training_points=training_points field=class model_name=RandomForestClassifier n_estimators=500 save_model=rf_model.gz --o
Extracting training data
Removing samples with NaN values in the raster feature variables...
Traceback (most recent call last):
  File "/home/aweinmann/repos/grass/grass/dist.x86_64-pc-linux-gnu/scripts/r.learn.train", line 892, in <module>
    main()
  File "/home/aweinmann/repos/grass/grass/dist.x86_64-pc-linux-gnu/scripts/r.learn.train", line 599, in main
    if y.dtype in (np.object_, np.object):
  File "/home/aweinmann/.local/lib/python3.8/site-packages/numpy/__init__.py", line 284, in __getattr__
    raise AttributeError("module {!r} has no attribute "
AttributeError: module 'numpy' has no attribute 'object'
```
See https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations and https://numpy.org/devdocs/release/1.24.0-notes.html#expired-deprecations.

**To Reproduce**

1. Use data from 'nc_spm_08/landsat' mapset:
2. Compute the following
```
g.region raster=landclass96 -p
r.random input=landclass96 npoints=1000 raster=training_pixels -s
r.to.vect in=training_pixels out=training_points type=point column=class
r.learn.train group=lsat7_2000 training_points=training_points field=class model_name=RandomForestClassifier n_estimators=500 save_model=rf_model.gz
```
3. See error

**Expected behavior**
No error with numpy >  1.24.0 

**System description (please complete the following information):**

- Operating System: Linux Mint 20.1
- GRASS GIS version :
  ```
  g.version -reg
  version=8.3.dev
  date=2023
  revision=b8d7eaf52c
  build_date=2023-02-10
  build_platform=x86_64-pc-linux-gnu
  build_off_t_size=8
  libgis_revision=d22ee4748e
  libgis_date=2023-02-02T07:48:36+00:00
  proj=6.3.1
  gdal=3.0.4
  geos=3.8.0
  sqlite=3.31.1
  ```
- numpy: 1.24.1

ecodiv:
This issue should be solved with #878

PR #878, opened 2023-03-18
r.learn.train: np.object to object

ecodiv:
Following up on bug report #872 - with numpy > 1.24.0 the numpy object is depricated. Changing `np.object` to `object` should solve the problem.


anikaweinmann:
Thanks a lot for this fix :)

anikaweinmann (approved):

PR #873, opened 2023-03-08
New feature -  fixed encoding issue with stdin

isaacullah:
Fixed encoding issue with stdin for r.colors

PR #881, opened 2023-03-22
r.landscape.evol: fix iterative numbering scheme that was broken by #873

isaacullah:
When used in iterative mode, module failed on second iteration due to a bug introduced in #873. Added an additional elif statement to ensure module can be used in single mode (no iteration number appended to output map names) and in iterative mode (iteration numbers added).


neteler:
(I have taken liberty to rename the PR title to a more standardized style:

`module: description`)

isaacullah:
Thanks Markus. This is a critical bugfix involving only a few lines of code. I suggest to merge immediately.

PR #886, opened 2023-04-11
r.bioclim: End with success when precipitation is missing

wenzeslaus:
The precipitation parameter is optional, so the process return code should be 0, not 1, as no error occured.

A return statement is more appropriate here than a direct exit call because this is a main function. Using return is consistent with the other branch which ends with the end of the function (an implicit return).

### DIFF
--- a/src/raster/r.bioclim/r.bioclim.py
+++ b/src/raster/r.bioclim/r.bioclim.py
@@ -412,7 +412,7 @@ def main():
         grass.run_command(
             "g.remove", flags="f", type="raster", pattern=tmp_pattern, quiet=True
         )
-        sys.exit(1)
+        return
 
     precl = prec.split(",")
 
--- a/src/raster/r.bioclim/r.bioclim.py
+++ b/src/raster/r.bioclim/r.bioclim.py
@@ -78,7 +78,6 @@
 # % description: Number of parallel processes to launch
 # %End
 
-import sys
 import os
 import string
 import grass.script as grass


src/raster/r.bioclim/r.bioclim.py
```diff
@@ -412,7 +411,7 @@ def main():
         grass.run_command(
             "g.remove", flags="f", type="raster", pattern=tmp_pattern, quiet=True
         )
-        sys.exit(1)
+        return
```
metzm, line 414:
```suggestion
        sys.exit(0)
```
This is a process, not a function. See https://docs.python.org/3/library/sys.html#sys.exit. `sys.exit()` is also needed to call `cleanup()` registered with `atexit.register(cleanup)`

wenzeslaus: Thanks for looking at this. As far as I understand, calling `main()` in the `if __name__ == "__main__":` block works sort of like a main function in C. This works as expected with return:

```python
import atexit

def cleanup():
    print("Doing cleanup")

def main():
    if True:
        return
    print("Not doing this")

if __name__ == "__main__":
    atexit.register(cleanup)
    main()
```

```bash
$ python test.py 
Doing cleanup
$ echo $?
0
```

metzm: Sure it works with `return`. Using `sys.exit(0)` is IMHO more in line with the purpose of the code. I regard it as good practice to call `sys.exit()` (or `exit()` in C code) in `main` functions, otherwise use `return` unless there is good reason to tear down the whole process with `(sys.)exit()`.

wenzeslaus: I think the idea in Python is to be able the function in a different context, [this post](https://www.artima.com/weblogs/viewpost.jsp?thread=4829) by Guido van Rossum (2003) suggests return and summarizes:

>  For Python programmers, I've got some suggestions on how to write a main() function that's easy to invoke in other contexts, e.g. from the interactive Python prompt when you feel like experimenting.

We are not 100% there, because you need to do `sys.argv = ...` instead of simply supplying the arguments, but we are close. However, my main point is that it is the best practice. After all, the main function is, in the first place, a function and in Python, its special meaning is only given by the way it is called.

wenzeslaus:
As mentioned in the initial description, using return makes this function exit-point consistent with the implicit return used at the end of the function in the current code.

PR #887, opened 2023-04-15
Dynamic subdirs in Makefile

ninsbl:
With sparse checkout of addons in `g.extension` implemented in https://github.com/OSGeo/grass/pull/2895, not all subdirs of the repository exist at all time.

Runing make in a directory below module level is thus likely to fail. This PR should make sure that _make_ is run in all subdirs that contain a _Makefile_. And instead of hardcoded directory names, relevant subdirs are identified with the `find` command line tool.

When `g.extension` is improved to support also the installation of multiple AddOns (which now should be relatively straight forward on UNIX like systems) then _make_ should be run at the root of the addon repository to install all downloaded addons.

Are find, basename, dirname and xargs available on all relevant UNIX like platforms*?
It seems Mac OS has those (but I have only Fedora and Ubuntu to test).

*On MS Windows addons are not compiled anyway.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -1,8 +1,7 @@
 
 MODULE_TOPDIR = ..
 
-# display gui misc
-SUBDIRS := $(shell find ./ -maxdepth 2 -mindepth 2 -name "Makefile" -exec dirname {} \; | xargs -I {} basename {} )
+SUBDIRS := ${sort ${dir ${wildcard */Makefile}}}
 
 include $(MODULE_TOPDIR)/include/Make/Dir.make
 
--- a/src/Makefile
+++ b/src/Makefile
@@ -2,15 +2,7 @@
 MODULE_TOPDIR = ..
 
 # display gui misc
-SUBDIRS = \
-	db \
-	display \
-	general \
-	imagery \
-	raster \
-	raster3d \
-	vector \
-	temporal
+SUBDIRS := $(shell find ./ -maxdepth 2 -mindepth 2 -name "Makefile" -exec dirname {} \; | xargs -I {} basename {} )
 
 include $(MODULE_TOPDIR)/include/Make/Dir.make
 
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,8 @@
+
+MODULE_TOPDIR = ..
+
+SUBDIRS = src
+
+include $(MODULE_TOPDIR)/include/Make/Dir.make
+
+default: parsubdirs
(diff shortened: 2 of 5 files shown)


src/Makefile
```diff
@@ -2,15 +2,7 @@
...
-SUBDIRS = \
-	db \
-	display \
-	general \
-	imagery \
-	raster \
-	raster3d \
-	vector \
-	temporal
+SUBDIRS := $(shell find ./ -maxdepth 2 -mindepth 2 -name "Makefile" -exec dirname {} \; | xargs -I {} basename {} )
```
tmszi: Find command omit [gui ](https://github.com/OSGeo/grass-addons/tree/grass8/src/gui/wxpython) and [hadoop ](https://github.com/OSGeo/grass-addons/tree/grass8/src/hadoop/hd) subdirectory, which contains addons too. These directories are missing from original list of subdirs.

https://github.com/OSGeo/grass-addons/blob/fcbb27b383d61aab40977060f1d2a6ca72184a94/src/Makefile#L5-L13

tmszi: Looks good to me now.

tmszi:
@ninsbl with this change we introduce this issue:

```
GRASS nc_basic_spm_grass7/PERMANENT:~ > g.extension db.join
WARNING: Extension <db.join> already installed. Re-installing...
Fetching <db.join> from <https://github.com/OSGeo/grass-addons> (be
patient)...
  File "/usr/lib64/grass83/scripts/g.extension", line 2877, in <module>
    sys.exit(main())
  File "/usr/lib64/grass83/scripts/g.extension", line 2857, in main
    install_extension(source=source, url=url, xmlurl=xmlurl, branch=branch)
  File "/usr/lib64/grass83/scripts/g.extension", line 1240, in install_extension
    ) = install_extension_std_platforms(
  File "/usr/lib64/grass83/scripts/g.extension", line 2001, in install_extension_std_platforms
    srcdir = download_source_code(
  File "/usr/lib64/grass83/scripts/g.extension", line 1913, in download_source_code
    directory = download_source_code_official_github(
  File "/usr/lib64/grass83/scripts/g.extension", line 1779, in download_source_code_official_github
    ga = GitAdapter(
  File "/usr/lib64/grass83/scripts/g.extension", line 252, in __init__
    self.addons = self._get_addons_list()
  File "/usr/lib64/grass83/scripts/g.extension", line 388, in _get_addons_list
    addons_dict[file_path.split("/")[3]] = "/".join(
IndexError: list index out of range

```

PR #900, opened 2023-05-26
i.sentinel.download: Run sys.exit() after message "No product found"

nifg414:


src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
```diff
@@ -667,7 +667,8 @@ def filter(
         products_df = self._api.to_dataframe(products)
         if len(products_df) < 1:
             gs.message(_("No product found"))
-            return
+            sys.exit()
+
```
ninsbl:
```suggestion
            sys.exit()
```

ninsbl (commented):
Required change to satisfy black

nifg414:
i.sentinel.download tested locally and works as expected.

ninsbl (approved):

wenzeslaus:
I'm late, but that's what `gs.fatal(...)` is for. User then sees `ERROR:` (signaling that further operation is not possible) and receives non-zero return code (useful for scripting).

ninsbl:
Right, sorry. I should have kept the PR open a bit longer for discussion.
The reason `sys.exit()` appeared right to me is that I would not consider it an error if querying the Copernicus database returns an an empty result. The previous code implied the same line of thinking (message + return)...

PR #911, opened 2023-06-03
r.mblend: basic tests

ldesousa:
* Adds first test suit: basic module functioning and simple control parameter effect
* Corrects small issue in the module with temporary map name

**NOTE** depends on: #910.

### DIFF
--- a/src/raster/r.mblend/testsuite/test_rmblend.py
+++ b/src/raster/r.mblend/testsuite/test_rmblend.py
@@ -61,7 +61,7 @@ class TestRMblend(TestCase):
             inter_points=100,
         )
         self.assertRasterExists(name=self.result_1)
-        values = "min=1\nmax=2\nmean=1.187\nstddev=0.316"
+        values = {"min": 1, "max": 2, "mean": 1.187, "stddev": 0.316}
         self.assertRasterFitsUnivar(
             raster=self.result_1, reference=values, precision=1e-3
         )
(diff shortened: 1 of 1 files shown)


src/raster/r.mblend/testsuite/test_rmblend.py
```diff
@@ -0,0 +1,84 @@
...
+
+        self.assertModule(
+            "r.mblend",
+            high=self.high_res_1,
+            low=self.low_res,
+            output=self.result_1,
+            inter_points=100,
+        )
+        self.assertRasterExists(name=self.result_1)
+        values = "min=1\nmax=2\nmean=1.187\nstddev=0.316"
```
wenzeslaus: This can be a dict.

src/raster/r.mblend/testsuite/test_rmblend.py
```diff
@@ -0,0 +1,84 @@
...
+
+    low_res = "low_res"
+    high_res_1 = "high_res_1"
+    high_res_2 = "high_res_2"
+    result_1 = "result_1"
+    result_2 = "result_2"
+
+    @classmethod
+    def setUpClass(cls):
+
```
wenzeslaus, line 18: The empty line is not needed here, but if you run Pylint it will make you add a comment in this place.

src/raster/r.mblend/testsuite/test_rmblend.py
ldesousa: Modified in e178a3aeea7010c2315e34e075adfdfa2e88f1d9

src/raster/r.mblend/testsuite/test_rmblend.py
ldesousa: I added a doc string to the class but left the methods without it, like in other unit tests. Since these tests are so simple the doc string won't be adding much.

The space between class/method declarations and body is a matter of style. Makes it easier for me read. Pylint is fine with it.

ldesousa:
@wenzeslaus A reminder to merge this PR once #910 is merged.

PR #924, opened 2023-06-14
wx.metadata: lazy import owslib so pre-compiling for MS Windows succeeds

ninsbl:
Pre-compilation of ` wx.metadata` for MS Windows fails because `owslib` is not consequently lazy-imported.

See:
https://github.com/OSGeo/grass/issues/2517

### DIFF
--- a/src/gui/wxpython/wx.metadata/mdlib/mdeditorfactory.py
+++ b/src/gui/wxpython/wx.metadata/mdlib/mdeditorfactory.py
@@ -41,7 +41,6 @@ import wx
 import wx.lib.scrolledpanel as scrolled
 
 from . import globalvar
-from . import mdutil
 from .mdjinjaparser import JinjaTemplateParser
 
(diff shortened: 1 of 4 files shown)


tmszi (commented):
My implementation of lazy module import (mdutil module) not working as expected. I'm going to refactoring it all.

tmszi:
Successfully tested with:

```
test@gentoo-gnu-linux:~$ python -c "import owslib"
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'owslib'

GRASS epsg_4326/PERMANENT:~ > g.extension wx.metadata url=$HOME/src/grass-addons/src/gui/wxpython/wx.metadata/
Compiling...
Installing...
Updating extensions metadata file...
Updating extension modules metadata file...
Installation of <wx.metadata> successfully finished
```

ninsbl:
Thanks @tmszi !

Issue #930, opened 2023-07-10
[Bug] r.fuzzy.set rounds shape/converts parameter decimal value to 0


ecodiv:
**Name of the addon**
r.fuzzy.set

**Describe the bug**
Decimal values for the shape parameter results in the same result as filling in 0. It seems decimal values are rounded to 0.

**To Reproduce**
Steps to reproduce the behaviour:

1. Open r.fuzzy.set and run it using any input raster layer of type float or double. Set the Shape parameter to 0, the boundary type to ' S-shaped', and the fuzzy range to 'Left'
2. Repeat the same, but this time, set the Shape parameter to 0.5
3. Compare the two outcomes: they are the same. Using the 

**Expected behavior**
Function 

**Screenshots**
The screenshot shows the command, with the shape=0.1 and shape=0.5. The result  is the same. As the bivariate plots show, the results are what one would be expected for shape=0.

![screenshotexample](https://github.com/OSGeo/grass-addons/assets/7816789/4d565ee9-2a92-4a33-bf19-dae64199006d)

**System description (please complete the following information):**

- Operating System: [Windows 10, Ubuntu 22.04]
- GRASS GIS version [e.g. 8.3]

YannChemin:
Using code from August 22, rebuilt grass + grass-addons from main, using the blue band of BMNG [https://visibleearth.nasa.gov/images/57752/blue-marble-land-surface-shallow-water-and-shaded-topography](url).

Make it Float
-------------------
r.mapcalc expression="bmngfloat=bmng.blue/100.0"

Map 1
---------
r.fuzzy.set --overwrite input=bmngfloat@PERMANENT output=out1 points=0.5,0.6 side=left shape=0.5

![image](https://github.com/OSGeo/grass-addons/assets/5877104/3637c5e7-907a-48b1-a745-7c02a7f446b2)


Map 2
---------
r.fuzzy.set --overwrite input=bmngfloat@PERMANENT output=out points=0.5,0.6 side=left shape=0.1

![image](https://github.com/OSGeo/grass-addons/assets/5877104/3028f4d8-05ab-41b3-89b5-56e60032f1ef)

ecodiv:
Thanks, I wonder if I introduced an error in https://github.com/OSGeo/grass-addons/pull/931. I'll do some further testing.

ecodiv:
Hi @YannChemin, below a reproducible example. 

[...]

PR #931, opened 2023-07-10
r.fuzzy.set: correction of sshape formula in fuzzy.c

ecodiv:
The calculation of the Sshape for when the shape parameter < 0 seems to be incorrect. 

I haven't checked (yet) the other shape formulas

Note; there is another issue I am not sure how to solve. See https://github.com/OSGeo/grass-addons/issues/930

### DIFF
--- a/src/raster/r.fuzzy.set/fuzzy.c
+++ b/src/raster/r.fuzzy.set/fuzzy.c
@@ -1,5 +1,6 @@
 #include <grass/glocale.h>
 #include "local_proto.h"
+#include <math.h>
 
 float fuzzy(FCELL cell)
 {
@@ -42,7 +43,7 @@ float fuzzy(FCELL cell)
     }
 
     if (type == SSHAPE) {
-        m = pow(2, exp(2 * abs(shape)));
+        m = pow(2, exp(2 * fabs(shape)));
         return (shape < 0) ? 1 - pow(cos(x * PI2), m) : pow(sin(x * PI2), m);
     }
 
--- a/src/raster/r.fuzzy.set/fuzzy.c
+++ b/src/raster/r.fuzzy.set/fuzzy.c
@@ -43,7 +43,7 @@ float fuzzy(FCELL cell)
 
     if (type == SSHAPE) {
         m = pow(2, exp(2 * abs(shape)));
-        return (shape < 0) ? pow(1 - cos(x * PI2), m) : pow(sin(x * PI2), m);
+        return (shape < 0) ? 1 - pow(cos(x * PI2), m) : pow(sin(x * PI2), m);
     }
 
     return -1; /* error */
--- a/src/raster/r.fuzzy.set/r.fuzzy.set.html
+++ b/src/raster/r.fuzzy.set/r.fuzzy.set.html
@@ -75,7 +75,7 @@ GIS operations.
 Depending on type of the boundary different equation are used to determine its
 shape:
 <p>
-<b>Linear:</b> the membership is calculated according following equation:<br>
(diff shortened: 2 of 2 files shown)


ecodiv:
The issue reported in https://github.com/OSGeo/grass-addons/issues/930 should be solved after replacing the `abs` function (works with integer values) with the `fabs` function (works with floating and double values). Thanks to @GISjoerd for the pointer.

wenzeslaus (commented):
Given that shape is float, fabs is correct. I can't comment on the other change. Will you link an equation?

ecodiv:
> Given that shape is float, fabs is correct. I can't comment on the other change. Will you link an equation?

According to the manual page, the equation for the s-shape is:

```
sin(x * Pi/2)^m (for positive shape parameter)
1-cos(x * Pi/2)^m (for negative shape parameter)
```

The formula in the original code is:

```
return (shape < 0) ? pow(1 - cos(x * PI2), m) : pow(sin(x * PI2), m);
```

The question is which one is correct. I can't find the equation in one of the references provided in the manual page. Testing both shows that the formula used in the original code does not yield the expected result (shape of the curve). Using the formula from the manual does.

Using the original formula, varying the s-parameter between -1 and 1 gives the following curves (created using the R/grass code in [test_fuzzy_set.zip](https://github.com/OSGeo/grass-addons/files/12681053/test_fuzzy_set.zip) with as input the landsat layer lsat5_1987_10@landsat from the North Carolina full dataset (https://grass.osgeo.org/download/data/).

original_code

The results shows that the results for negative shape parameters are not correct.

The equation in the PR is:

```
return (shape < 0) ? 1 - pow(cos(x * PI2), m) : pow(sin(x * PI2), m);
```
Using this equation gives the following results:

adapted_formula

These results show (more or less) the expected s-shaped curves shown in the manual page:

general shape of the s-shaped curve for different shape values

wenzeslaus (approved):
This is great. Thank you for demonstrating why the new code is correct. I'm for merging this PR now.

I have two notes which don't prevent this PR from being merged:

1. It would be a good idea to write tests for the tool now when you consider the results to be correct. (You write the test using the current values of the tool assuming they are correct. Any change which would change these values fails the test allowing us to evaluate if the change should or should not modify the results.)
2. The bottom curves in the images look more angular, less smooth. I think I see it in both images, so maybe it is just sampling, not an issue with the code. I did not check the R code.

ecodiv:
> This is great. Thank you for demonstrating why the new code is correct. I'm for merging this PR now.

Thanks. I'll merge it.

> 
> I have two notes which don't prevent this PR from being merged:
> 
>     1. It would be a good idea to write tests for the tool now when you consider the results to be correct. (You write the test using the current values of the tool assuming they are correct. Any change which would change these values fails the test allowing us to evaluate if the change should or should not modify the results.)

For now, I have some code to test locally, and perhaps it is relatively easy to convert that into an automated test, but I have no experience of doing that whatsoever. So will look into that when time allows.

> 
>     2. The bottom curves in the images look more angular, less smooth. I think I see it in both images, so maybe it is just sampling, not an issue with the code. I did not check the R code.

Yes, I sampled a real RS layer, which probably has much less values in the extremes of the value range.

wenzeslaus:
> ...I have some code to test locally, and perhaps it is relatively easy to convert that into an automated test, but I have no experience of doing that whatsoever. So will look into that when time allows.

The grass.gunittest testing framework allows tests in Python unittest syntax as well as shell scripts. In any case, some automated check of results is needed (as opposed to human inspecting the results). pytest-based tests are not yet checked in addons, but are actively checked in core, so I think it's okay to start adding them to addons, too.

PR #935, opened 2023-07-27
v.stream.inbasin: Modernize interface, fix for v8

wenzeslaus:
- Changes x_outlet and y_outlet to single coordinates standard option which is recognized by GUI and user can click to get coordinates there. Call in Python has less parameters and same readability (x,y).
- Makes coordinates optional which seems to be the original intention.
- Uses rules to decide about valid options instead of custom if statements.
- Updates interfacing with numpy to Python 3.
- Adds more recent reference.
- Describes stream input more.
- Adds tests.


awickert (approved):
Thanks to @wenzeslaus, especially for knowing how to appropriately interface this to the various GRASS GIS interface options. Streamlining the code is also much appreciated!

PR #936, opened 2023-07-27
v.stream.network: Fix tostream name, fix for v8

wenzeslaus:
- Fixes use of custom tostream which was ignored.
- Updates interfacing with numpy for Python 3.
- Updates v.to.db use for v8 where columns are created automatically (saves >10 lines here).
- Simplifies code and removes old code and comments.
- Renames variables and comply more with Flake8 and Pylint.
- Adds tests.
- Explains input more.
- Adds reference.


wenzeslaus:
@awickert If you can review this or at least give general +1, that would be great. Same for #935. Any reviews are welcome of course.

awickert (approved):
Thank you @wenzeslaus for modernizing this module!

awickert:
> @awickert If you can review this or at least give general +1, that would be great. Same for #935. Any reviews are welcome of course.

And here: Thanks for bearing with the few-days delay -- I know that writing good code and then having it sit can be annoying. I just got back to Potsdam.

Many thanks as always, for all your work on GRASS!

wenzeslaus:
You are welcome and thanks for the reviews and merges. Much appreciated!

PR #940, opened 2023-09-06
r.pi manual: completion of submodule list and HTML fixes

neteler:
- update scientific citation
- fix HTML header
- add missing keywords


neteler:
Hi @wegmann - I brushed up the `r.pi` manual pages, please take a look if possible.

PR #942, opened 2023-09-16
r.vif.py: add error message

ecodiv:
The error message given when one of the input layers has no values (within the computational region) does not help to understand the problem. Added an error message to suggest the possible cause of the error. On the command line, the error message is written in red.

### DIFF
--- a/src/raster/r.vif/r.vif.py
+++ b/src/raster/r.vif/r.vif.py
@@ -123,10 +123,12 @@ import grass.script as gs
 
 
 # Functions
-def prRed(skk): print("\033[91m {}\033[00m" .format(skk))
+def prRed(skk):
+    print("\033[91m {}\033[00m".format(skk))
 
 
-def prGreen(skk): print("\033[92m {}\033[00m" .format(skk))
+def prGreen(skk):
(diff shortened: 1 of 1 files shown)


wenzeslaus:
Here is some feedback about this: We try to make the behavior across the 800+ different tools in GRASS GIS standardized. This is adding two colors for messages which are not present in other tools. I'm all for colorful messages with message for an error and a hint/tip, but we need a more standardized approach which would be reusable in other tools. Having a PR adding your two functions to the library would be a good start.

ecodiv:
Hi @wenzeslaus , good point, didn't think about that. Wouldn't mind making such PR, but wouldn't know where to do that?

wenzeslaus:
The closest existing thing is _grass.script.message_ and _grass.script.fatal_. They live in [python/grass/script/core.py](https://github.com/OSGeo/grass/blob/main/python/grass/script/core.py#L655) in the core repo. You can see from there that the Python side mostly relies on the _g.message_ tool [in C](https://github.com/OSGeo/grass/blob/main/general/g.message/main.c), but let's have that discussion in the PR.

PR #944, opened 2023-09-20
r.learn.ml2: change max_features=auto to sqrt

metzm:
From the [scikit docs](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.RandomForestClassifier.html#sklearn.ensemble.RandomForestClassifier):

```
max_features{“sqrt”, “log2”, None}, int or float, default=”sqrt”
[...]
Changed in version 1.1: The default of max_features changed from "auto" to "sqrt".
```
The "auto" option is no longer existing and was an alias for "sqrt", this PR adapts `r.learn.train` accordingly without changing the behaviour and staying backwards compatible with older scikit versions.


mmacata (approved):
thanks!

PR #957, opened 2023-10-17
r.flowaccumulation: Handle null values properly

HuidaeCho:


src/raster/r.flowaccumulation/global.h
```diff
@@ -32,4 +32,12 @@ long long timeval_diff(struct timeval *, struct timeval *, struct timeval *);
 /* accumulate.c */
 void accumulate(struct raster_map *, struct raster_map *);
 
+#ifdef _MAIN_C_
+#define GLOBAL
+#else
+#define GLOBAL extern
+#endif
+
+GLOBAL CELL cell_null;
```
wenzeslaus, line 41: Using Rast_is_c_null_value would avoid the need for an award global variable. Would Rast_is_c_null_value work here?

HuidaeCho: Right, Rast_is_c_null_value should also work. That global was my attempt for micro-optimization. Just did a benchmark test against Rast_is_c_null_value. Well.. not much difference... Maybe, GCC can optimize out this function call as inline. Yeah.. not worth it.

wenzeslaus: Benchmark! Nice! The function is in fact [a macro](https://grass.osgeo.org/programming8/defs_2raster_8h_source.html#l00410), so it should be always inlined.

HuidaeCho: Ah! I missed `#ifndef Rast_is_c_null_value`!
```c
#ifndef Rast_is_c_null_value
int Rast_is_c_null_value(const CELL *cellVal)
{
    /* Check if the CELL value matches the null pattern */
    return *cellVal == (CELL)0x80000000;
}
#endif
```

PR #972, opened 2023-10-28
v.in.pygbif: fix identifiers

ninsbl:
Format / data type of identifiers was changed in _pygbif_. This PR adjusts _v.in.pygbif_ accordingly...


hellik:
thanks for updating

PR #978, opened 2023-11-07
r.flowaccumulation: Remove a global variable

HuidaeCho:
This PR removes a global variable and uses `Rast_is_c_null_value()` instead. There wasn't much performance difference between inline and function calls.


wenzeslaus (approved):

wenzeslaus:
Thanks. The original code would require some more explanation anyway, while this new code is clear.

HuidaeCho:
> Thanks. The original code would require some more explanation anyway, while this new code is clear.

Yep. And both are inlined anyway.

PR #979, opened 2023-11-07
m.prism.download: Add a new addon for downloading PRISM climate data

HuidaeCho:
This PR adds a new addon for downloading climate data from the [PRISM Climate Group](https://prism.oregonstate.edu/).


neteler:
Now online:

https://grass.osgeo.org/grass8/manuals/addons/m.prism.download.html

PR #980, opened 2023-11-08
v.rast.move: Add missing keywords

### LINKED ISSUE #950: v.rast.move: empty keywords break addon overview script
The cronjob to generate addons struggles with

https://github.com/OSGeo/grass-addons/blob/7a3dfb74e3827a3960a761060449f2ddaf3503e8/src/vector/v.rast.move/v.rast.move.py#L22

`Empty keyword from file v.rast.move.html line: <a href="../vector.html">vector</a>, <a href="../topic_.html"></a>, <a href="../topic_.html"></a>, <a href="../topic_.html"></a>
`

At least three keywords are needed:
- module family
- topic
- specific keyword(s)
-

wenzeslaus:
Fixes #950

### DIFF
--- a/src/vector/v.rast.move/v.rast.move.py
+++ b/src/vector/v.rast.move/v.rast.move.py
@@ -20,6 +20,7 @@
 # % description: Move vertices by distance specified in a raster
 # % keyword: vector
 # % keyword: transformation
+# % keyword: geometry
 # % keyword: raster
 # % keyword: lines
 # %end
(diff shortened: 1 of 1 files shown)


src/vector/v.rast.move/v.rast.move.py
```diff
@@ -19,9 +19,9 @@
 # %module
 # % description: Move vertices by distance specified in a raster
 # % keyword: vector
-# % keyword:
-# % keyword:
-# % keyword:
+# % keyword: transformation
+# % keyword: raster
+# % keyword: lines
```
neteler, line 25:
```suggestion
# % keyword: lines
# % keyword: geometry
```

This additional keyword might be a good fit for
https://grass.osgeo.org/grass-stable/manuals/keywords.html#geometry

wenzeslaus: Thanks. Added as a third keyword.

PR #1001, opened 2023-12-27
r.edm.eval: port shell/R script to Python

ecodiv:
The addon is rewritten in Python. This includes some changes in parameter names and functionality. I am planning to add some functionality, like comparing prediction layers and creating reclass layers, but that will be for a later stage. I would like to change the name of the addon to better reflect the functionalities, but haven't been able to come up with something yet.


lucadelu (commented):
I'm not able to judge but if it works fine for me.

PR #1002, opened 2023-12-28
i.sentinel.mask: refactor and add sun_position and output option

ninsbl:
This PR is a refactoring of i.sentinel.mask. The main function is split into several sub-functions, and cloud shadow detections is switched to raster processing (thus replacing #138). On my local machine, _i.sentinel.mask_ now finishes masking a scene with a relative high amount of clouds (~30%) within 2.5 minutes (> 10x faster). Furthermore, processing times should be relatively stable if no vector output is requested.

In addition a sun_position and an output option are added. In the "output" raster map both, cloud and cloud shadow raster are combined. Hence, "output" can be the only result from _i.sentinel.mask_. cloud_raster and shadow_raster output is kept for backwards compatibility. With the _sun_position_ option the metadata file is no longer needed as input. 

There is a slight difference compared to the original results for the cloud shadow_raster output if users request size filtered cloud shadows (now the resulting raster is size filtered too)...

Tests with real world data as well as all othe feedback would be most welcome!

### DIFF
--- a/src/imagery/i.sentinel/i.sentinel.mask/i.sentinel.mask.py
+++ b/src/imagery/i.sentinel/i.sentinel.mask/i.sentinel.mask.py
@@ -184,7 +184,7 @@ COPYRIGHT:    (C) 2018-2023 by Roberta Fagandini, and the GRASS Development Team
 # %rules
 # % collective: blue,green,red,nir,nir8a,swir11,swir12
 # % excludes: mtd_file,metadata,sun_position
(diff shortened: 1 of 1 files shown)


ninsbl:
Anyone able to review?
If there are no objections either, I would like to merge this rather soon...

ninsbl:
If there are still no objections tonight, I will merge this. So, if the module is critical for you and yu see issues here, please speak up now...

PR #1018, opened 2024-01-21
v.what.rast.multi.py: deal with @ in column names

ecodiv:
**Problem:**
When no column names are provided, the names of the input raster layers are used as column names. If these include the mapset suffix, an error is thrown (SQLite does not accept @).

**Provided solution:**
The mapset suffix is removed. Alternatively, if the mapset names should be in the column names, the -m flag can be set. In that case, the @ is replaced with an underscore


pierreroudier:
Thanks @ecodiv :)

PR #1028, opened 2024-02-21
v.delaunay3D: remove -lCGAL

lbartoletti:
CGAL is header only since v5 and no need to link against a lib


neteler (approved):
Locally tested, compiles fine.

PR #1030, opened 2024-02-28
i.sentinel.download: address pandas iloc FutureWarning

ninsbl:
Currently, `i.sentinel.download` throws a _FutureWarning_ from the use of pandas:

E.g.:
```
.grass8/addons/scripts/i.sentinel.download:704: FutureWarning: Series.__getitem__ treating keys as positions is deprecated. In a future version, integer keys will always be treated as labels (consistent with DataFrame behavior). To access a value by position, use `ser.iloc[pos]`
  self._products_df_sorted[id_kw[kw_idx]][idx],
```

This PR addresses the FutureWarning....

### DIFF
--- a/src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
+++ b/src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
@@ -707,12 +707,16 @@ class SentinelDownloader(object):
             if kw_idx == 1:
                 time_string = self._products_df_sorted[time_kw[kw_idx]].iloc[idx]
             else:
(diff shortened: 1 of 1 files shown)


griembauer (approved):
Thanks, looks good to me! Have you used i.sentinel.download recently? I would expect the only way it works at the moment given the change to CDSE is to directly search for specific scenes using `identifier` or `uuid` and downloading from GCS.

ninsbl:
Yes, it is still relevant and useful for the Norwegian Ground Segment (see: https://github.com/sentinelsat/sentinelsat/pull/612) until end of 2023...
So, we are still using it...

PR #1034, opened 2024-03-04
cronjobs: fix new version red boxes in older manuals

neteler:
- inject "current manual page" red boxes at a later stage in the building process
- use `NEW_CURRENT` variable to define current version only at top of script (avoid versions hardcoded in URLs)

Needed for
- https://grass.osgeo.org/grass78/manuals/ and https://grass.osgeo.org/grass78/manuals/addons/
- https://grass.osgeo.org/grass82/manuals/ and https://grass.osgeo.org/grass82/manuals/addons/


echoix (approved):
I'm fine with the concept, and reading through I didn't find anything that caught my eye. I can't pronounce myself on the bash/script itself, but I assume it's more of your specialty. Since it is only for supporting the infrastructure, and you maintain it anyways, there's no reason to block it. If it's problematic it most likely will be you that will make the changes

neteler:
> Since it is only for supporting the infrastructure, and you maintain it anyways, there's no reason to block it. If it's problematic it most likely will be you that will make the changes

Yes, confirmed.

PR #1062, opened 2024-04-18
v.histogram: update content to match new name

veroandreo:
Meanwhile, I also fixed html tags and some leftovers, as well as keywords in v.scatterplot.


petrasovaa (approved):
Thanks!

Issue #1071, opened 2024-04-25
Addons for plotting: Standardize parameters and flags


veroandreo:
## Context

While creating a tutorial on the use of GRASS GIS addons to create plots, I found out that parameters doing the same task were sometimes named differently. Also, there were some addons providing a lot of control over the plot, and others much less so. I ended up creating a big double entry table comparing flags and parameters names and functionalities across GRASS addons for creating plots. 

This issue attempts to point out which addons would need flag/parameter name standardization and which would benefit from the addition of further options already present in some but not all.    

### Parameter name standardization

- [ ] r.boxplot: input should be changed to map
- [x] v.boxplot, v.scatterplot and v.histogram: plot_output and file_name should be changed to output
    - Implemented in #1075 , but wondering, do we need to keep the old parameter for backward compatibility (not my preference, but if needed, can somebody tell me the best way to do that?)
- [x] t.rast.boxplot and t.rast.line: font_size should be changed to fontsize consistently with d.* modules in core done in #1075
- [ ] r.boxplot and v.boxplot: standardize parameter to sort boxplots based on median values
- [ ] r.boxplot, r.series.boxplot and t.rast.boxplot: standardize option for boxplot fill color
- [ ] r.boxplot, r.series.boxplot and t.rast.boxplot: standardize option for boxplot width
- [ ] v.boxplot and v.scatterplot: standardize option to group by
- [x] v.scatterplot and t.rast.line: standardize option for number of SD to plot. 
    - Change sd_n to n in v.scatterplot because in t.rast.line the n is for are sd or se. (done in #1085)
- [ ] v.scatterplot, t.rast.boxplot and t.rast.line: standardize option to set y axis range values
    - what about axis_limits, y_axis_limits and x_axis_limits (which to use depend on options for the particular plot type)?
    - Done for v.scatterplot following suggestion above bulletpoint.

### Change flag default behavior

- [x] v.boxplot: -o flag in v.boxplot hides outliers, while in the other boxplot addons it includes outliers. 

### Flags into parameters

- [ ] v.boxplot: convert flag -r in into rotate_labels

### Add other control parameters or flags (parameter name: addons that would benefit from the added parameter / addon name: parameters)

- [ ] plot_dimensions: v.boxplot, v.histogram
- [ ] fontsize: v.boxplot, v.histogram
- [ ] title: r.boxplot, r.series.boxplot, v.boxplot, v.histogram, t.rast.boxplot, t.rast.line
[...]

PR #1075, opened 2024-05-09
Standardize parameters addons for plotting (issue #1071 )

ecodiv:
Implementing some of the suggestions in issue #1071

Question is how to deal with backward compatibility.


veroandreo:
> Implementing some of the suggestions in issue #1071
> 
> Question is how to deal with backward compatibility.

Well, we don't really assure backward compatibility in addons though it might be nice to have. What's the usual practice @petrasovaa ?

PR #1083, opened 2024-05-20
r.suitability.regions: change/add parameter

ecodiv:
Add parameter percentage, to be used instead of percentile_threshold (the latter is still available for back-ward compatibility.


neteler:
Ok, too late... 


While at it, would you mind to rename the figure files?

```
# (on ubuntu: rename.ul)
rename r_suitability_regions_ r_suitability_regions_option r_suitability_regions_*
sed -i 's+"option+"r_suitability_regions_option+g' r.suitability.regions.html
```

Then commit...

PR #1084, opened 2024-05-21
r.suitability.regions renames image files

ecodiv:
Renamed images following guidelines. @neteler, is this you requested (the rename.ul function doesn't exist anymore on Ubuntu 22.04 it seems, so I changed it manually).
Added reference to dependency on r.area (for one of the options).


src/raster/r.suitability.regions/r.suitability.regions.html
```diff
@@ -200,8 +202,14 @@ <h2>See also</h2>
 <li><a href="https://grass.osgeo.org/grass-stable/manuals/r.reclass.area.html">r.reclass.area</a></li>
 <li><a href="https://grass.osgeo.org/grass-stable/manuals/r.clump.html">r.clump</a></li>
 <li><a href="https://grass.osgeo.org/grass-stable/manuals/v.to.db.html">v.to.db</a></li>
+<li><a href="r.area">r.area</a></li>
```
neteler:
```suggestion
<li><a href="r.area.html">r.area</a></li>
```

neteler (approved):
Thanks (just accept the small fix).

PR #1091, opened 2024-05-31
r.connectivity.distance bugfix

ecodiv:
- Bugfix resulting in error when using the -p flag
- Correction in example on manual page
- Some small code cleanup

_question: can `from osgeo import gdal, osr, ogr`  be replaced by `from osgeo import ogr`? The gdal and osr do not seem to be used?_

### DIFF
--- a/src/raster/r.connectivity/r.connectivity.distance/r.connectivity.distance.html
+++ b/src/raster/r.connectivity/r.connectivity.distance/r.connectivity.distance.html
@@ -1,32 +1,43 @@
 <h2>DESCRIPTION:</h2>
-<em>r.connectivity.distance</em> computes cost-distance between all
+
+<em>r.connectivity.distance</em> computes cost-distances between all
 areas (patches) of an input vector map within a user defined Euclidean
 distance threshold.
 
-<p>Recently, graph-theory has been characterised as an efficient and
-useful tool for conservation planning (e.g. Bunn et al. 2000,
-Calabrese &amp; Fagan 2004, Minor &amp; Urban 2008, Zetterberg et. al. 2010).</p>
+<p>
+Recently, graph-theory has been characterised as an efficient and 
+useful tool for conservation planning (e.g. Bunn et al. 2000, Calabrese 
+&amp; Fagan 2004, Minor &amp; Urban 2008, Zetterberg et. al. 2010).
 
-<p>As a part of the r.connectivity.* tool-chain, <em>r.connectivity.distance</em> is
-intended to make graph-theory more easily available to conservation
-planning.</p>
+<p>
+As a part of the r.connectivity.* tool-chain, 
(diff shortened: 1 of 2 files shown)


ninsbl:
Hi, @ecodiv and sorry for the inconvenience. I do have a major update of the module that includes parallel processing. However, when testing my update again, I discovered a bug in `v.db.join` in 8.4. I plan to post a fix for `v.db.join` tonight and then create a PR for the new `r.connectivity.distance`

ecodiv:
I'm having trouble getting rpy2 to work on my computer so I was just looking into how difficult it would be to replace the R part in the second module with Python. Is that something which you are (still) considering to do as well?

ninsbl:
Yes, The R implementation is buggy and not as performant as I wish it would be. I do have a partial replacement of R with Python for `r.connectivity.network`. Will have a look into it and at least provide you with what I have so far (as I am currently not working on ecology topics...).

ecodiv:
Thanks, that would be great.

src/raster/r.connectivity/r.connectivity.distance/r.connectivity.distance.html
```diff
@@ -1,32 +1,43 @@
...
 polygons within a user-defined Euclidean distance threshold.</p>
 
-<p>It produces two vector maps that hold the network:</p>
+<p>
+It produces two vector maps that hold the network:
+
+<p>
 <ul>
 <li>an edge-map (connections between patches) and a</li>
 <li>vertex-map (centroid representations of the patches).</li>
```
veroandreo, line 33:
```suggestion
<li>an edge-map (connections between patches) and,</li>
<li>a vertex-map (centroid representations of the patches).</li>
```

src/raster/r.connectivity/r.connectivity.distance/r.connectivity.distance.html
```diff
@@ -47,52 +58,62 @@ <h2>DESCRIPTION:</h2>
...
+with start and stop points.
 
-<p>In addition, <em>r.connectivity.distance</em> outputs a cost distance
-raster map for every input area which later on are used in
-<em>r.connectivity.corridors</em> (together with output from
-<em>r.connectivity.network</em>) for corridor identification.</p>
+<p>
+In addition, <em>r.connectivity.distance</em> outputs a cost distance 
+raster map for every input area which later on are used in 
+<em>r.connectivity.corridors</em> (together with output from
```
veroandreo, line 91:
```suggestion
In addition, <em>r.connectivity.distance</em> outputs a cost distance 
raster map for every input area. These are used later on in  
<em>r.connectivity.corridors</em> (together with the output of 
```

ecodiv:
> I plan to post a fix for `v.db.join` tonight and then create a PR for the new `r.connectivity.distance` 

@ninsbl how far are you with this? If you need somebody to test things, let me know.

PR #1093, opened 2024-06-03
v.nnstat: fix bool compilation error

neteler:
Attempt to fix compilation error:

```
make
...
                 from mbb.c:1:
mbb.h:9:32: error: expected ‘;’, identifier or ‘(’ before ‘_Bool’
    9 | typedef enum { BFALSE, BTRUE } bool;
      |                                ^~~~
compilation terminated due to -Wfatal-errors.
```

### DIFF
--- a/src/vector/v.nnstat/getval.c
+++ b/src/vector/v.nnstat/getval.c
@@ -83,7 +83,7 @@ void read_points(struct Map_info *map, int field, struct nna_par *xD,
 
     /* Get 3rd coordinate of 2D points from attribute column -> 3D interpolation
      */
-    if (xD->v3 == FALSE &&
+    if (xD->v3 == false &&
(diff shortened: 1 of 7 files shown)


nilason (changes requested):
You might as well remove the lines 8-10 all together.

Not necessary, but in my opinion additional improvement, would be to add `#include <stdbool.h>` to `local_proto.h` and change all `FALSE` and `TRUE` to the standard lowercase versions (`false` and `true`) in the C files in the v.nnstat addon directory.
The reason the uppercase macros work is that they are defined in grass/gis.h and kept with https://github.com/OSGeo/grass/commit/1e0c11f526f64755f877957ce6c14b2159c3823f.

nilason (approved):
I have not tested, but it looks good to me :)

PR #1104, opened 2024-06-17
i.eodag: remove pandas dependency and save results in json/geojson files

HamedElgizery:
This PR will:

- [x] Remove Pandas dependency
- [x] Add filtering with eodag crunch [1]
- [x] Add sorting
- [x] Add DOCSTRINGS  
- [x] Add JSON support
- [x] Save search results as a geojson file, which can later be read by i.eodag.

Note: Reading geojson containing search results is not yet supported, and it will be addressed in a future PR.  

Added filtering options:
* area_relation
* minimum_overlap
* filtering by dates (useful when reading products from a file)

Sorting parameters:
* cloudcover
* ingestiondate 

Both can be set to ascending or descending.

[1] https://eodag.readthedocs.io/en/stable/api_reference/searchresult.html#crunch


src/imagery/i.eodag/i.eodag.py
```diff
@@ -300,82 +396,200 @@ def no_fallback_search(search_parameters, provider):
...
+
+    if area_relation or minimum_overlap:
+        # Product's geometry intersects with AOI
+        if area_relation == "Intersects":
+            search_result = search_result.filter_overlap(
+                geometry=geometry, intersects=True
+            )
+        # Product's geometry contains the AOI
+        elif area_relation == "Contains":
+            print("HERE")
```
```suggestion
```

src/imagery/i.eodag/i.eodag.py
```diff
@@ -82,6 +82,23 @@
...
+# % description: Spatial relation of footprint to AOI
+# % options: Intersects,Contains,IsWithin
+# % required: no
+# % guisection: Region
+# %end
+
+# %option
+# % key: minimum_overlap
+# % type: integer
+# % description: Spatial percentage of AOI covered
```
ninsbl: Mayb something like this:
```suggestion
# % description: Minimal overlap with the AOI in percent of the scene covered
```

src/imagery/i.eodag/i.eodag.py
```diff
@@ -255,37 +330,58 @@ def setup_environment_variables(env, **kwargs):
...
+    :param datetime_str: Datetime in ISO format
+    :type datetime_str: str
+
+    :return: Datetime converted to 'YYYY-MM-DDTHH:MM:SS'
+    :rtype: str
+    """
     normalized_datetime = datetime.fromisoformat(datetime_str)
+    # Remove microseconds
     normalized_datetime = normalized_datetime.replace(microsecond=0)
+    # Remove timezone
```
ninsbl: Maybe converting to UTC is more appropriate than removing a potentially assigned time zone?

src/imagery/i.eodag/i.eodag.py
```diff
@@ -300,82 +396,199 @@ def no_fallback_search(search_parameters, provider):
...
+    """Lists products on the Standard Output (Console) stream.
+
+    :param products: EO poducts to be listed
+    :type products: class:'eodag.api.search_result.SearchResult'
+    """
+    columns = ["id", "startTimeFromAscendingNode", "cloudCover", "productType"]
+    columns_NA = ["id_NA", "time_NA", "cloudCover_NA", "productType_NA"]
+    for product in products:
+        product_line = ""
+        for i, column in enumerate(columns):
```
ninsbl, line 438: The EOProduct class has an _as_dict_ method (https://eodag.readthedocs.io/en/latest/api_reference/eoproduct.html#eodag.api.product._product.EOProduct.as_dict) that could be used for printing (e.g. in JSON format, or converted to CSV like output (`",".join(product.as_dict().values())` ).

You could then also iterate over the dict items or use the get method to retrieve specific attributes....

src/imagery/i.eodag/i.eodag.py
```diff
@@ -300,82 +396,199 @@ def no_fallback_search(search_parameters, provider):
...
-                    "startTimeFromAscendingNode" in product.properties
-                    and product.properties["startTimeFromAscendingNode"] is not None
-                ):
+def list_products(products):
+    """Lists products on the Standard Output (Console) stream.
+
+    :param products: EO poducts to be listed
+    :type products: class:'eodag.api.search_result.SearchResult'
+    """
+    columns = ["id", "startTimeFromAscendingNode", "cloudCover", "productType"]
```
ninsbl, line 434: I am not sure about this one. The way it is implemented now seems too limiting to me. I would rather print all attributes of a product if the user does not deliberately wants to limit what is printed.

columns should therefore probably be a kwarg with None as default. Personally, I would avoid hardcoding column names in the function.... Yet, unless you add a columns option to the module, users will not be able to control the behavior. This is not necessarily needed, but I would prefer a less restrictive solution...

src/imagery/i.eodag/i.eodag.py
```diff
@@ -300,82 +396,199 @@ def no_fallback_search(search_parameters, provider):
...
+            if i != 0:
+                product_line += " "
+            product_line += product_attribute_value
+        print(product_line)
 
-    df = pd.DataFrame().from_dict(result_dict)
-    return df
 
+def remove_duplicates(search_result):
+    """Removes duplicated products, in case a provider returns a product multiple time."""
```
ninsbl: Does that happen in praxis? Or do you get duplicates from a search across different providers? I can imagine that UUIDs differ but scenes may be not (same scene file name, but different IDs in the different backends)... Is the latter handeled here?
```suggestion
    """Removes duplicated products, in case a provider returns a product multiple times."""
```

src/imagery/i.eodag/i.eodag.py
```diff
@@ -300,82 +396,199 @@ def no_fallback_search(search_parameters, provider):
...
+
+    sort_keys = options["sort"].split(",")
+    sort_order = options["order"].split(",")
+    sort_order.extend(["asc"] * max(0, len(sort_keys) - len(sort_order)))
+
+    # Sort keys and sort orders are matched respectively
+    def products_compare(first, second):
+        for idx, sort_key in enumerate(sort_keys):
+            if sort_key == "ingestiondate":
+                first_value = first.properties["startTimeFromAscendingNode"]
```
ninsbl, line 608: Does "startTimeFromAscendingNode" translate to "ingestiondate" (when the product was uploaded to the repository)? I thought it was rather start of sensing time (when data take started).

src/imagery/i.eodag/i.eodag.py
```diff
@@ -300,82 +396,199 @@ def no_fallback_search(search_parameters, provider):
...
+                second_value = second.properties["startTimeFromAscendingNode"]
+            elif sort_key == "cloudcover":
+                first_value = first.properties["cloudCover"]
+                second_value = second.properties["cloudCover"]
+            if first_value < second_value:
+                return 1 if sort_order[idx] == "desc" else -1
+            elif first_value > second_value:
+                return -1 if sort_order[idx] == "desc" else 1
```
ninsbl: What is supposed to happen if `first_value > second_value`?

After a return elif is unnecessary. Could be removed
```suggestion
            return -1 if sort_order[idx] == "desc" else 1
```
unless you want to handle first_value > second_value specifically...

src/imagery/i.eodag/i.eodag.py
```diff
@@ -300,82 +396,199 @@ def no_fallback_search(search_parameters, provider):
...
+    :param search_result: EO products to be sorted
+    :type search_result: class'eodag.api.search_result.SearchResult'
+
+    :return: Sorted EO products
+    :rtype: class:'eodag.api.search_result.SearchResult'
+    """
+    gs.verbose(_("Sorting..."))
+
+    sort_keys = options["sort"].split(",")
+    sort_order = options["order"].split(",")
```
ninsbl: If those two options do not have input with equal length, the code below will fail... You should check that early on.

Alternatively, you could merge the two and allow add "asc" or "desc" after sort property in the sort input... But then you would have to check that if "order" is given, that is written either _asc_ or _desc_ and you could not use a drop-down option in the GUI...

src/imagery/i.eodag/i.eodag.py
```diff
@@ -300,82 +396,199 @@ def no_fallback_search(search_parameters, provider):
...
+    gs.verbose(_("Filtering results..."))
+
+    if area_relation or minimum_overlap:
+        # Product's geometry intersects with AOI
+        if area_relation == "Intersects":
+            search_result = search_result.filter_overlap(
+                geometry=geometry, intersects=True
+            )
+        # Product's geometry contains the AOI
+        elif area_relation == "Contains":
```
ninsbl, line 556: `minimum_overlap` probably does not make a lot of sense with other `area_relations` than _Intersects_ as _IsWithin_ or _Contains_ means 100% overlap anyway...
That means aso the structure of the _elif_ statements should probably be reconsidered...

ninsbl (approved):
Nice! Lots of useful improvements!
I added a couple of suggestions for your consideration. Feel free to reject or incorporate as you feel for it. I leave an approving review anyways...

src/imagery/i.eodag/i.eodag.py
```diff
@@ -220,22 +280,37 @@ def get_aoi(vector=None):
 
 
 def search_by_ids(products_ids):
-    gs.message("Searching for products...")
+    """Searchs for products based on their ids.
```
lucadelu:
```suggestion
    """Search for products based on their ids.
```

src/imagery/i.eodag/i.eodag.py
HamedElgizery: Yes, especially cop_dataspace, not sure if there is other providers that has the same issue. But, cop_dataspace sometimes returns the same scene twice (same scene name and same uuid). I have been trying to figure why it is happening, but no luck yet...

src/imagery/i.eodag/i.eodag.py
HamedElgizery: If the values are equal it should skip the current sorting criteria, e.g. cloudcover, and try to differentiate between them with the next sorting criteria, e.g. ingestiondate. If it couldn't differentiate it returns 0 at the end indicating equality.

src/imagery/i.eodag/i.eodag.py
HamedElgizery: I made the minimum_overlap and area_relations options exclusive, that is why it works. But I agree, it would make more sense if I move it to another separate condition.

src/imagery/i.eodag/i.eodag.py
lucadelu: Yeah I support this, I suggest to print as json or as CSV. Look at the other GSOC about the JSON output to use a similar approach.

src/imagery/i.eodag/i.eodag.py
HamedElgizery: Maybe we can limit the `order` option to one value, either "asc" or "desc", that is how it is implemented in i.sentinel.download and i.landsat.download, and also because the GUI, AFAIK, can't have multiple values and preset limiting options at the same time, and it just allows the user to choose only one value from the drop-down once...

HamedElgizery: Did that here ebc988487906cb2346efba999c12c71fc3770a9c.

src/imagery/i.eodag/i.eodag.py
HamedElgizery: Maybe we can have output to be in JSON format by default. We can then have two ways of printing, maybe 'l' flag and 'L' flag, the former to just print the main product attributes, and the latter to print all the metadata of a product?

I don't think it would be practical to add a `column` option, because the attributes varies with different products.

HamedElgizery: I am mainly concerned that the products metadata are quite a lot to present on the console.

See for example:
<details>
  <summary>
    One product JSON metadata
  </summary>

```
{
    "type": "Feature",
    "geometry": {
        "type": "Polygon",
        "coordinates": [
            [
                [
                    1.727450341675,
                    45.146747187364
                ],
                [
                    3.124166397187,
                    45.15376975552
                ],
                [
                    3.122074490936,
                    44.16530127443
                ],
                [
                    1.748882946646,
                    44.158515309138
                ],
                [
                    1.727450341675,
                    45.146747187364
                ]
            ]
        ]
    },
    "id": "LANDSAT8-OLITIRS-XS_20201120-103609-534_L2A_T31TDK_D",
    "assets": {},
    "properties": {
        "eodag_product_type": "L8_REFLECTANCE",
        "eodag_provider": "theia",
        "eodag_search_intersection": {
            "type": "Polygon",
            "coordinates": [
                [
                    [
                        1.9,
                        44.1592620913938
                    ],
                    [
                        1.9,
                        45.0
                    ],
                    [
                        2.0,
                        45.0
                    ],
                    [
                        2.0,
                        44.1597562661009
                    ],
                    [
                        1.9,
                        44.1592620913938
                    ]
                ]
            ]
        },
        "abstract": "Landsat 8 L2A data distributed by Theia since 2017-03-20 using operational version of MUSCATE, UTM projection,\nand tiled using Sentinel-2 tiles.\n",
        "instrument": "OLITIRS",
        "platform": "LANDSAT8",
        "platformSerialIdentifier": "LANDSAT8",
        "processingLevel": "LEVEL2A",
        "keywords": "OLI,TIRS,LANDSAT,LANDSAT8,L8,L2,L2A,MUSCATE",
        "sensorType": "OPTICAL",
        "license": "proprietary",
        "missionStartDate": "2013-02-11T00:00:00Z",
        "title": "LANDSAT8-OLITIRS-XS_20201120-103609-534_L2A_T31TDK_D",
        "productType": "REFLECTANCE",
        "uid": "37e9cc4b-b3c5-5929-bc3a-0e332e5e43ac",
        "keyword": [
            {
                "id": "bce097de57d1ab7",
                "name": "Europe",
                "normalized": "europe",
                "type": "continent",
                "href": "https://theia.cnes.fr/atdistrib/resto2/api/collections/LANDSAT/search.json?&lang=en&q=Europe"
            },
            {
                "id": "936bcbcf5c43fbd",
                "name": "France",
                "normalized": "france",
                "type": "country",
                "parentHash": "bce097de57d1ab7",
                "value": 100,
                "gcover": 1.89,
                "href": "https://theia.cnes.fr/atdistrib/resto2/api/collections/LANDSAT/search.json?&lang=en&q=France"
            },
            {
                "id": "fa018f4f7b20706",
                "name": "Midi-Pyr\u00e9n\u00e9es",
                "normalized": "midi-pyrenees",
                "type": "region",
                "parentHash": "936bcbcf5c43fbd",
                "value": 46.94,
                "gcover": 64.22,
                "href": "https://theia.cnes.fr/atdistrib/resto2/api/collections/LANDSAT/search.json?&lang=en&q=Midi-Pyr%C3%A9n%C3%A9es"
            },
            {
                "id": "a507bd46cbf5398",
                "name": "Aveyron",
                "normalized": "aveyron",
                "type": "state",
                "parentHash": "fa018f4f7b20706",
                "value": 46.94,
                "gcover": 64.22,
                "href": "https://theia.cnes.fr/atdistrib/resto2/api/collections/LANDSAT/search.json?&lang=en&q=Aveyron"
            },
            {
                "id": "c958f1cb7878712",
                "name": "Auvergne",
                "normalized": "auvergne",
                "type": "region",
                "parentHash": "936bcbcf5c43fbd",
                "value": 28.83,
                "gcover": 60.04,
                "href": "https://theia.cnes.fr/atdistrib/resto2/api/collections/LANDSAT/search.json?&lang=en&q=Auvergne"
            },
            {
                "id": "36b380ec75fcb73",
                "name": "Cantal",
                "normalized": "cantal",
                "type": "state",
                "parentHash": "c958f1cb7878712",
                "value": 28.83,
                "gcover": 60.04,
                "href": "https://theia.cnes.fr/atdistrib/resto2/api/collections/LANDSAT/search.json?&lang=en&q=Cantal"
            },
            {
                "id": "595a2a0433fc5be",
                "name": "Limousin",
                "normalized": "limousin",
                "type": "region",
                "parentHash": "936bcbcf5c43fbd",
                "value": 5.61,
                "gcover": 11.45,
                "href": "https://theia.cnes.fr/atdistrib/resto2/api/collections/LANDSAT/search.json?&lang=en&q=Limousin"
            },
            {
                "id": "2b6fe1d663541d5",
                "name": "Corr\u00e8ze",
                "normalized": "correze",
                "type": "state",
                "parentHash": "595a2a0433fc5be",
                "value": 5.61,
                "gcover": 11.45,
                "href": "https://theia.cnes.fr/atdistrib/resto2/api/collections/LANDSAT/search.json?&lang=en&q=Corr%C3%A8ze"
            },
            {
                "id": "455ae3d75f982e8",
                "name": "Languedoc-Roussillon",
                "normalized": "languedoc-roussillon",
                "type": "region",
                "parentHash": "936bcbcf5c43fbd",
                "value": 2.33,
                "gcover": 5.44,
                "href": "https://theia.cnes.fr/atdistrib/resto2/api/collections/LANDSAT/search.json?&lang=en&q=Languedoc-Roussillon"
            },
            {
                "id": "c34df802ddfe137",
                "name": "Loz\u00e8re",
                "normalized": "lozere",
                "type": "state",
                "parentHash": "455ae3d75f982e8",
                "value": 2.33,
                "gcover": 5.44,
                "href": "https://theia.cnes.fr/atdistrib/resto2/api/collections/LANDSAT/search.json?&lang=en&q=Loz%C3%A8re"
            },
            {
                "id": "a0e4a10aa754ab4",
                "name": "Northern",
                "type": "location",
                "href": "https://theia.cnes.fr/atdistrib/resto2/api/collections/LANDSAT/search.json?&lang=en&q=Northern"
            },
            {
                "id": "f3221bcc6ca49ca",
                "name": "Autumn",
                "type": "season",
                "href": "https://theia.cnes.fr/atdistrib/resto2/api/collections/LANDSAT/search.json?&lang=en&q=Autumn"
            },
            {
                "id": "c16401fc77fb38d",
                "name": "LANDSAT",
                "type": "collection",
                "href": "https://theia.cnes.fr/atdistrib/resto2/api/collections/LANDSAT/search.json?&lang=en&q=LANDSAT"
            },
            {
                "id": "988188952c531b3",
                "name": "REFLECTANCE",
                "type": "productType",
                "href": "https://theia.cnes.fr/atdistrib/resto2/api/collections/LANDSAT/search.json?&lang=en&q=REFLECTANCE"
            },
            {
                "id": "9ff2d85eb9bf912",
                "name": "LEVEL2A",
                "type": "processingLevel",
                "href": "https://theia.cnes.fr/atdistrib/resto2/api/collections/LANDSAT/search.json?&lang=en&q=LEVEL2A"
            },
            {
                "id": "9df06a863dca8a5",
                "name": "landsat8",
                "type": "platform",
                "href": "https://theia.cnes.fr/atdistrib/resto2/api/collections/LANDSAT/search.json?&lang=en&q=landsat8"
            },
            {
                "id": "3eac071fb9c3418",
                "name": "OLITIRS",
                "type": "instrument",
                "parentHash": "9df06a863dca8a5",
                "href": "https://theia.cnes.fr/atdistrib/resto2/api/collections/LANDSAT/search.json?&lang=en&q=OLITIRS"
            },
            {
                "id": "e4d819a79fcf300",
                "name": "XS",
                "type": "sensorMode",
                "parentHash": "3eac071fb9c3418",
                "href": "https://theia.cnes.fr/atdistrib/resto2/api/collections/LANDSAT/search.json?&lang=en&q=XS"
            },
            {
                "id": "6414a4147122d1b",
                "name": "2020",
                "type": "year",
                "href": "https://theia.cnes.fr/atdistrib/resto2/api/collections/LANDSAT/search.json?&lang=en&q=2020"
            },
            {
                "id": "d212fdc821d78a2",
                "name": "November",
                "type": "month",
                "parentHash": "6414a4147122d1b",
                "href": "https://theia.cnes.fr/atdistrib/resto2/api/collections/LANDSAT/search.json?&lang=en&q=November"
            },
            {
                "id": "5cfcce1f5fb3c9e",
                "name": "20",
                "type": "day",
                "parentHash": "d212fdc821d78a2",
                "href": "https://theia.cnes.fr/atdistrib/resto2/api/collections/LANDSAT/search.json?&lang=en&q=20"
            }
        ],
        "resolution": 30,
        "organisationName": "THEIA",
        "publicationDate": "2021-03-29T21:04:40Z",
        "accessConstraint": {
            "licenseId": "unlicensed",
            "hasToBeSigned": "never",
            "grantedCountries": null,
            "grantedOrganizationCountries": null,
            "grantedFlags": null,
            "viewService": "public",
            "signatureQuota": -1,
            "description": {
                "shortName": "No license"
            }
        },
        "parentIdentifier": "urn:ogc:def:EOP:CNES::LANDSAT:",
        "orbitNumber": 198,
        "cloudCover": 87,
        "snowCover": 0,
        "productVersion": "2-2",
        "creationDate": "2021-03-29T19:14:33Z",
        "sensorMode": "XS",
        "startTimeFromAscendingNode": "2020-11-20T10:36:09Z",
        "completionTimeFromAscendingNode": "2020-11-20T10:36:09Z",
        "quicklook": "https://theia.cnes.fr/data//LANDSAT/2020/11/20/LANDSAT8-OLITIRS-XS_20201120-103609-534_L2A_T31TDK_D_V2-2/LANDSAT8-OLITIRS-XS_20201120-103609-534_L2A_T31TDK_D_V2-2_QKL_ALL.png",
        "downloadLink": "https://theia.cnes.fr/atdistrib/resto2/collections/LANDSAT/37e9cc4b-b3c5-5929-bc3a-0e332e5e43ac/download",
        "tileIdentifier": "31TDK",
        "thumbnail": "https://theia.cnes.fr/data//LANDSAT/2020/11/20/LANDSAT8-OLITIRS-XS_20201120-103609-534_L2A_T31TDK_D_V2-2/LANDSAT8-OLITIRS-XS_20201120-103609-534_L2A_T31TDK_D_V2-2_THB_ALL.png",
        "storageStatus": "ONLINE",
        "relativeOrbitNumber": 198,
        "updated": "2021-03-29T21:04:40Z",
        "waterCover": 0,
        "centroid": {
            "type": "Point",
            "coordinates": [
                2.425808369431,
                44.656142532329
            ]
        },
        "percentSaturatedPixelsMax": null,
        "percentNoDataPixelsMax": 0,
        "nbColInterpolationErrorMax": null,
        "percentGroundUsefulPixels": null,
        "percentUsefulPixelsMin": null,
        "bands": "B1;B2;B3;B4;B5;B6;B7",
        "thermBands": "",
        "nb_cols": 3660,
        "nb_rows": 3660,
        "tileId": null,
        "oragnisationName": "CNES",
        "services": {
            "browse": {
                "title": "Display full resolution product on map",
                "layer": {
                    "type": "WMS",
                    "url": "https://theia.cnes.fr/atdistrib/resto2/collections/LANDSAT/37e9cc4b-b3c5-5929-bc3a-0e332e5e43ac/wms?map=/nfs/muscate_data//muscate.map&idProduit=LANDSAT8-OLITIRS-XS_20201120-103609-534_L2A_T31TDK_D&versionProduit=2-2&idProjet=LANDSAT&anneeProduit=ANNEEPRODUIT&moisProduit=MOISPRODUIT&jourProduit=JOURPRODUIT&LAYERS=muscate&FORMAT=image%2Fpng&TRANSITIONEFFECT=resize&TRANSPARENT=true&VERSION=1.1.1&SERVICE=WMS&REQUEST=GetMap&STYLES=&SRS=EPSG%3A3857&BBOX={:bbox3857:}&WIDTH=256&HEIGHT=256",
                    "layers": ""
                }
            },
            "download": {
                "url": "https://theia.cnes.fr/atdistrib/resto2/collections/LANDSAT/37e9cc4b-b3c5-5929-bc3a-0e332e5e43ac/download",
                "mimeType": "application/zip",
                "checksum": "2bffda59f38ac6e71bfbbd29acbd149d"
            }
        },
        "links": [
            {
                "rel": "self",
                "type": "application/json",
                "title": "GeoJSON link for 37e9cc4b-b3c5-5929-bc3a-0e332e5e43ac",
                "href": "https://theia.cnes.fr/atdistrib/resto2/collections/LANDSAT/37e9cc4b-b3c5-5929-bc3a-0e332e5e43ac.json?&lang=en"
            },
            {
                "rel": "via",
                "type": "application/unknown",
                "title": "Metadata link for 37e9cc4b-b3c5-5929-bc3a-0e332e5e43ac",
                "href": "https://theia.cnes.fr/data//LANDSAT/ANNEEPRODUIT/MOISPRODUIT/JOURPRODUIT/LANDSAT8-OLITIRS-XS_20201120-103609-534_L2A_T31TDK_D_V2-2/LANDSAT8-OLITIRS-XS_20201120-103609-534_L2A_T31TDK_D_V2-2_MTD_ALL.xml"
            }
        ],
        "urlwms": "https://mapserver.theia.cnes.fr/cgi-bin/mapserv?map=/nfs/muscate_data//muscate.map&idProduit=LANDSAT8-OLITIRS-XS_20201120-103609-534_L2A_T31TDK_D&versionProduit=2-2&idProjet=LANDSAT&anneeProduit=2020&moisProduit=11&jourProduit=20&LAYERS=muscate&FORMAT=image%2Fpng&TRANSITIONEFFECT=resize&TRANSPARENT=true&VERSION=1.1.1&SERVICE=WMS&REQUEST=GetMap&STYLES=&SRS=EPSG%3A3857&BBOX=192298.89237916,5490005.6506699,347780.6124399,5645761.9581448&WIDTH=256&HEIGHT=256&"
    }
}
```
</details>

So having all of that on the console for multiple products, won't be practical. Shall we just keep the normal listing to the console with these four main attributes (ID, Ingestion Date, Cloud Cover, Product Type)... and if the user needs the full metadata of the products, we can add an option `save` and then save the results in either a JSON or a CSV file.

EODAG has this serializable methods for writing/reading scenes in/from geojson files BTW.
https://eodag.readthedocs.io/en/latest/api_reference/core.html#eodag.api.core.EODataAccessGateway.serialize
https://eodag.readthedocs.io/en/latest/api_reference/core.html#eodag.api.core.EODataAccessGateway.deserialize_and_register

src/imagery/i.eodag/i.eodag.html
```diff
@@ -63,24 +63,56 @@ <h2>EXAMPLES</h2>
...
+v.extract input=urbanarea where="NAME = 'Durham'" output=durham
+
+i.eodag -l start=2021-05-25 end=2022-06-01 \
+    dataset=S2_MSI_L2A provider=cop_dataspace \
+    clouds=50 map=durham minimum_overlap=70
+<pre></div>
+
+Sort results by <b>cloudcover</b>, and then by <b>ingestiondate</b> descendingly
+(latest to earliest):<br><em> Note that sorting with <b>cloudcover</b> use
+unrounded values, while they are rounded to the nearest intger when listing.</em>
```
```suggestion
(latest to earliest). Note that sorting with <b>cloudcover</b> use
unrounded values, while they are rounded to the nearest integer when listing.
```

src/imagery/i.eodag/i.eodag.html
```diff
@@ -63,24 +63,56 @@ <h2>EXAMPLES</h2>
...
+
+<div class="code"><pre>
+i.eodag -l start=2021-05-25 end=2022-06-01 \
+    dataset=S2_MSI_L2A provider=cop_dataspace \
+    sort=cloudcover,ingestiondate order=asc,desc
```
veroandreo: the explanation above says descending only... but the command uses both asc and desc. Does it sort asc by cloud cover and desc by ingestion date? Please, clarify the text or fix the example

src/imagery/i.eodag/i.eodag.html
```diff
@@ -63,24 +63,56 @@ <h2>EXAMPLES</h2>
...
+<pre></div>
+
+Search for scenes with a list of IDs text file, and filter the results with the
+provided parameters:
+
+<div class="code"><pre>
+i.eodag -l file=ids_list.txt \
+    start=2022-05-25 \
+    area_relation=Contains clouds=3
+<pre></div>
```
veroandreo, line 97: This example or functionality does not make sense to me. If users provides a file with IDs is because they already did the search and filtering and they want to download the scenes found (i.e. IDs in the file). I cannot imagine to provide a text file with IDs and eodag doing a reverse search and filter... Is that what it does?

src/imagery/i.eodag/i.eodag.html
```diff
@@ -63,24 +63,56 @@ <h2>EXAMPLES</h2>
...
+
+Download all available scenes in the tmp directory, with Cloud Coverage not
+exceeding 50%:
```
```suggestion
Download all available scenes with cloud coverage not exceeding 50% 
in the tmp directory:
```

src/imagery/i.eodag/i.eodag.py
```diff
@@ -82,6 +82,23 @@
...
+# % description: Spatial relation of footprint to AOI
+# % options: Intersects,Contains,IsWithin
+# % required: no
+# % guisection: Region
+# %end
+
+# %option
+# % key: minimum_overlap
+# % type: integer
+# % description: Minimal overlap with the AOI in percent of the scene covered
```
veroandreo: is it percent of the scene covered by the AOI or percent of the AOI covered by the scene? It is not clear from the definition

src/imagery/i.eodag/i.eodag.html
```diff
@@ -63,24 +63,56 @@ <h2>EXAMPLES</h2>
...
+    sort=cloudcover,ingestiondate order=asc,desc
 <pre></div>
 
-Download all available scenes in the tmp directory, with Cloud Coverage not exceeding 50%:
+Search and list the available Sentinel 2 scenes in the Copernicus Data Space
+Ecosystem, with at least 70% of the AOI covered:
+<div class="code"><pre>
+v.extract input=urbanarea where="NAME = 'Durham'" output=durham
+
+i.eodag -l start=2021-05-25 end=2022-06-01 \
```
lucadelu:
```suggestion
i.eodag -l start=2022-05-25 end=2022-06-01 \
```

src/imagery/i.eodag/i.eodag.html
```diff
@@ -63,24 +63,56 @@ <h2>EXAMPLES</h2>
...
+    dataset=S2_MSI_L2A provider=cop_dataspace \
+    clouds=50 map=durham minimum_overlap=70
+<pre></div>
+
+Sort results by <b>cloudcover</b>, and then by <b>ingestiondate</b> descendingly
+(latest to earliest):<br><em> Note that sorting with <b>cloudcover</b> use
+unrounded values, while they are rounded to the nearest intger when listing.</em>
+
+<div class="code"><pre>
+i.eodag -l start=2021-05-25 end=2022-06-01 \
```
lucadelu:
```suggestion
i.eodag -l start=2022-05-25 end=2022-06-01 \
```

src/imagery/i.eodag/i.eodag.py
```diff
@@ -101,11 +118,29 @@
...
-# % description: The provider to search within. Providers available by default: https://eodag.readthedocs.io/en/stable/getting_started_guide/providers.html
+# % label: The provider to search within.
+# % description: Providers available by default: https://eodag.readthedocs.io/en/stable/getting_started_guide/providers.html
 # % required: no
 # % guisection: Filter
 # %end
 
+# %option
+# % key: sort
+# % description: Sort by values in given order
```
```suggestion
# % description: Field to sort values by
```

src/imagery/i.eodag/i.eodag.py
```diff
@@ -128,23 +163,37 @@
...
+    """Creates directory."""
     try:
         Path(directory).mkdir(parents=True, exist_ok=True)
     except:
-        gs.fatal(_("Could not create directory {}").format(dir))
+        gs.fatal(_("Could not create directory {}").format(directory))
 
 
 def get_bb(proj):
+    """Gets the bouding box of a given projection information.
```
```suggestion
    """Gets the bounding box of the current computational 
    region in geographic coordinates.
```

src/imagery/i.eodag/i.eodag.py
```diff
@@ -220,22 +280,37 @@ def get_aoi(vector=None):
...
         else:
-            gs.message(_("Found."))
+            gs.verbose(_("Found."))
             search_result.append(product[0])
-    return search_result
+    return SearchResult(search_result)
 
 
 def setup_environment_variables(env, **kwargs):
+    """Sets up the eodag envirionmnets variables based on the provided options/flags.
```
```suggestion
    """Sets the eodag environment variables based on the provided options/flags.
```

src/imagery/i.eodag/i.eodag.py
```diff
@@ -255,37 +330,62 @@ def setup_environment_variables(env, **kwargs):
         if extract:
             gs.warning(
                 _(
-                    """Ignoring 'e' flag...
-                    'extract' option in the config file will be used.
-                    If you wish to use the 'e' flag, please specify a provider."""
+                    "Ignoring 'e' flag... \
+                    'extract' option in the config file will be used. \
```
veroandreo: Are these `\` supposed to be new lines? Shouldn't that be `\n`?

src/imagery/i.eodag/i.eodag.py
```diff
@@ -255,37 +330,62 @@ def setup_environment_variables(env, **kwargs):
...
                 )
             )
         if delete_archive:
             gs.warning(
                 _(
-                    """Ignoring 'd' flag...
-                    'delete_archive' option in the config file will be used.
-                    If you wish to use the 'd' flag, please specify a provider."""
+                    "Ignoring 'd' flag... \
+                    'delete_archive' option in the config file will be used. \
```
veroandreo: Same as above

src/imagery/i.eodag/i.eodag.py
```diff
@@ -255,37 +330,62 @@ def setup_environment_variables(env, **kwargs):
...
                 )
             )
         if output:
             gs.warning(
                 _(
-                    """Ignoring 'output' option...
-                    'output' option in the config file will be used.
-                    If you wish to use the 'output' option, please specify a provider."""
+                    "Ignoring 'output' option... \
+                    'output' option in the config file will be used. \
```
veroandreo: same as above

src/imagery/i.eodag/i.eodag.py
```diff
@@ -255,37 +330,62 @@ def setup_environment_variables(env, **kwargs):
...
     normalized_datetime = normalized_datetime.replace(microsecond=0)
+    # Convert time to UTC
+    normalized_datetime = normalized_datetime.astimezone(pytz.utc)
+    # Remove timezone info
     normalized_datetime = normalized_datetime.replace(tzinfo=None)
     return normalized_datetime.isoformat()
 
 
 def no_fallback_search(search_parameters, provider):
+    """Searches in a one and only one provider (fallback is disabled).
```
```suggestion
    """Search in only one provider (fallback is disabled).
```

src/imagery/i.eodag/i.eodag.py
```diff
@@ -300,82 +400,199 @@ def no_fallback_search(search_parameters, provider):
...
-                    "startTimeFromAscendingNode" in product.properties
-                    and product.properties["startTimeFromAscendingNode"] is not None
-                ):
+def list_products(products):
+    """Lists products on the Standard Output (Console) stream.
+
+    :param products: EO poducts to be listed
+    :type products: class:'eodag.api.search_result.SearchResult'
+    """
+    columns = ["id", "publicationDate", "cloudCover", "productType"]
```
veroandreo: @HamedElgizery do you know what publication date actually means? I had a look around and didn't find the definition. To me it sounds like the date when the product was put online, I think ingestion date is more relevant, i.e. when the sensor recorded the data. But I didn't find ingestionDate, I only found creationDate... whatever that means.

src/imagery/i.eodag/i.eodag.py
```diff
@@ -300,82 +400,199 @@ def no_fallback_search(search_parameters, provider):
...
+    :rtype: class:'eodag.api.search_result.SearchResult'
+    """
+    prefilter_count = len(search_result)
+    area_relation = kwargs["area_relation"]
+    minimum_overlap = kwargs["minimum_overlap"]
+    cloud_cover = kwargs["clouds"]
+    start_date = kwargs["start"]
+    end_date = kwargs["end"]
+
+    # If neither a geometry is provided as a paraemter
```
```suggestion
    # If neither a geometry is provided as a parameter
```

src/imagery/i.eodag/i.eodag.py
```diff
@@ -300,82 +400,199 @@ def no_fallback_search(search_parameters, provider):
...
+        elif area_relation == "Contains":
+            search_result = search_result.filter_overlap(
+                geometry=geometry, contains=True
+            )
+        # Product's geometry is within the AOI
+        elif area_relation == "IsWithin":
+            search_result = search_result.filter_overlap(geometry=geometry, within=True)
+
+    if minimum_overlap:
+        # Percentage of area covered from the AOI by the product's geometry
```
```suggestion
        # Percentage of the AOI area covered by the product's geometry
```

src/imagery/i.eodag/i.eodag.py
```diff
@@ -300,82 +400,199 @@ def no_fallback_search(search_parameters, provider):
...
+
+    sort_keys = options["sort"].split(",")
+    sort_order = options["order"]
+
+    # Sort keys and sort orders are matched respectively
+    def products_compare(first, second):
+        for sort_key in sort_keys:
+            if sort_key == "ingestiondate":
+                first_value = first.properties["publicationDate"]
+                second_value = second.properties["publicationDate"]
```
veroandreo: Same concern as above, let's try to understand if publication date is really equivalent to ingestion date. Users want the date sensor took the data, not the date the image was published, this is important information.

src/imagery/i.eodag/i.eodag.html
HamedElgizery: Yes, it takes the text file with a list of IDs, search for the scenes that is associated with each ID, and then it filters these founded scenes to make sure all of them satisfies the rest of the parameters. 

I think a use case would be, if for example, a user did a general search (i.e. without setting the cloud coverage percentage), and then they want to pick scenes from this list of products with a certain cloud coverage percentage...

src/imagery/i.eodag/i.eodag.html
HamedElgizery: Should be fixed now.

src/imagery/i.eodag/i.eodag.py
HamedElgizery: Should be fixed now.

src/imagery/i.eodag/i.eodag.py
HamedElgizery: I was leaving the part of print new lines to `gs.warning`, and these backslashes `\` was to be able to write a string on multiple lines (just in the code). 
But yes you are right, I added `\n` as well now.

src/imagery/i.eodag/i.eodag.py
HamedElgizery: Yeah right, just checked...

Apparently:
* `publicationDate` is the date they was put online.
* `startTimeFromAscendingNode` is the time when the scene was recorded.

Fixed here [6d99e7f](https://github.com/OSGeo/grass-addons/pull/1104/commits/6d99e7fc3f5b43adab54a8ae432b29e1ec623719).

src/imagery/i.eodag/i.eodag.html
HamedElgizery: Anyways IMO, I can imagine that in most use cases products found by searching can be saved as a geojson file, and that geojson file can later be parsed by the module again. What is good about these geojson files, is that you don't need to do the searching again, as you already have the products ready to be downloaded after parsing the geojson file.
For reference: https://eodag.readthedocs.io/en/stable/notebooks/api_user_guide/5_serialize_deserialize.html?highlight=serialize 

And for the text files it would just be a collection of IDs that is collected apart of the i.eodag module, and instead of typing them manually by hand in the shell they are just written in a text file... That is how I think it could be optimally used.

HamedElgizery:
Thanks for the feedback! @ninsbl @veroandreo @lucadelu 

I will be adding `footprint` option in the next PR, because I think this one already contains off-topic additions, right?
Also I will probably be adding more qureyables for searching in the next PR, or the one after.

src/imagery/i.eodag/i.eodag.html
lucadelu: @HamedElgizery ok your explanation could make sense for me, Your idea it is not a common behaviour, people usually look for the data and later they download them. @veroandreo what do you think? 

This seems the last open point before accept the pull request

veroandreo: I'm at a eodag workshop in FOSS4G, and now I see the potential utility of this. @HamedElgizery is this similar to what they do with the crunch functionality, that it filters a search object?

In any case, I think this can be merged.

PR #1136, opened 2024-07-01
i.eodag: add query, print, and footprints options

HamedElgizery:
- [x] Add `query` option, from which one can set extra querayables for searching.

- [x] Add `print` option to allow printing of the following in JSON format:
  - Current EODAG configuration.
  - Available products.
  - Available products for a given provider.
  - Available providers. 
  - Available providers offering a specific product.
  - Queryables for a given provider and/or product. 
  
**Note**: Queryables from the above list can then be used in the query option.

- [x] Add `foorprints` option to allow saving the scenes in a single combined footprint vector map.

### DIFF
--- a/src/imagery/i.eodag/i.eodag.html
+++ b/src/imagery/i.eodag/i.eodag.html
@@ -62,9 +62,16 @@ Search and list the available Sentinel 2 scenes in the Copernicus Data Space Eco
 <div class="code"><pre>
 v.extract input=urbanarea where="NAME = 'Durham'" output=durham
 
-i.eodag -l start=2022-05-25 end=2022-06-01 \
-    map=durham producttype=S2_MSI_L2A provider=cop_dataspace \
-    sort=cloudcover,ingestiondate order=asc,desc
+i.eodag -l start=2022-05-01 end=2022-06-01 \
+    map=durham producttype=S2_MSI_L2A provider=cop_dataspace
+
+6 scenes(s) found.
+S2B_MSIL2A_20220518T155819_N0400_R097_T17SPV_20220518T204903 2022-05-18T15:58:19 24% S2MSI2A
+S2A_MSIL2A_20220513T155821_N0400_R097_T17SPV_20220514T002317 2022-05-13T15:58:21 99% S2MSI2A
+S2B_MSIL2A_20220508T155819_N0400_R097_T17SPV_20220508T205707 2022-05-08T15:58:19 100% S2MSI2A
+S2A_MSIL2A_20220523T155831_N0400_R097_T17SPV_20220524T002713 2022-05-23T15:58:31 95% S2MSI2A
+S2A_MSIL2A_20220503T155831_N0400_R097_T17SPV_20220503T233818 2022-05-03T15:58:31 86% S2MSI2A
+S2B_MSIL2A_20220528T155819_N0400_R097_T17SPV_20220528T205325 2022-05-28T15:58:19  2% S2MSI2A
 <pre></div>
 
 Search and list the available Sentinel 2 scenes in the Copernicus Data Space
@@ -72,7 +79,7 @@ Ecosystem, with at least 70% of the AOI covered:
 <div class="code"><pre>
 v.extract input=urbanarea where="NAME = 'Durham'" output=durham
 
-i.eodag -l start=2022-05-25 end=2022-06-01 \
+i.eodag -l start=2022-05-01 end=2022-06-01 \
     producttype=S2_MSI_L2A provider=cop_dataspace \
     clouds=50 map=durham minimum_overlap=70
 <pre></div>
@@ -129,6 +136,41 @@ i.eodag -e provider=cop_dataspace \
     S2B_MSIL2A_20240529T081609_N0510_R121_T37SED_20240529T124818" \
     config=full/path/to/eodag/config.yaml \
     output=download_here
+<pre></div>
+
+List recognized eodag providers:
+<div class="code"><pre>
+i.eodag print=providers
+<pre></div>
+
+List recognized eodag providers, offering Sentinel 2 scenes:
+<div class="code"><pre>
+i.eodag print=providers producttype=S2_MSI_L2A
+<pre></div>
+
+List recognized eodag products:
+<div class="code"><pre>
+i.eodag print=products
+<pre></div>
+
+List recognized eodag products, offered by Copernicus Data Space Ecosystem:
+<div class="code"><pre>
+i.eodag print=products provider=cop_dataspace
+<pre></div>
+
+List queryables for Sentinel 2 scenes, that is offered by Copernicus Data Space Ecosystem:
+<div class="code"><pre>
+i.eodag print=queryables provider=cop_dataspace producttype=S2_MSI_L2A
+<pre></div>
+
+List current EODAG configuration:
+<div class="code"><pre>
+i.eodag print=config
+<pre></div>
+
+List current EODAG configuration for Copernicus Data Space Ecosystem:
+<div class="code"><pre>
+i.eodag print=config provider=cop_dataspace
 <pre></div>
 
 <h2>REQUIREMENTS</h2>
--- a/src/imagery/i.eodag/i.eodag.html
+++ b/src/imagery/i.eodag/i.eodag.html
@@ -100,7 +100,7 @@ Search and list the available Sentinel 2 scenes in the Copernicus Data Space
 Ecosystem, with relative orbit number 54:
 
 <div class="code"><pre>
-./i.eodag.py -l producttype=S2_MSI_L2A \
+i.eodag -l producttype=S2_MSI_L2A \
     provider=cop_dataspace save=search_result.geojson \
     query="relativeOrbitNumber=54"
 <pre></div>
--- a/src/imagery/i.eodag/i.eodag.html
+++ b/src/imagery/i.eodag/i.eodag.html
@@ -63,24 +63,56 @@ Search and list the available Sentinel 2 scenes in the Copernicus Data Space Eco
 v.extract input=urbanarea where="NAME = 'Durham'" output=durham
 
 i.eodag -l start=2022-05-25 end=2022-06-01 \
-    map=durham dataset=S2_MSI_L2A provider=cop_dataspace
+    map=durham dataset=S2_MSI_L2A provider=cop_dataspace \
+    sort=cloudcover,ingestiondate order=asc,desc
 <pre></div>
 
-Download all available scenes in the tmp directory, with Cloud Coverage not exceeding 50%:
+Search and list the available Sentinel 2 scenes in the Copernicus Data Space
+Ecosystem, with at least 70% of the AOI covered:
+<div class="code"><pre>
+v.extract input=urbanarea where="NAME = 'Durham'" output=durham
+
+i.eodag -l start=2021-05-25 end=2022-06-01 \
+    dataset=S2_MSI_L2A provider=cop_dataspace \
+    clouds=50 map=durham minimum_overlap=70
+<pre></div>
+
+Sort results by <b>cloudcover</b>, and then by <b>ingestiondate</b> descendingly
+(latest to earliest):<br><em> Note that sorting with <b>cloudcover</b> use
+unrounded values, while they are rounded to the nearest intger when listing.</em>
+
+<div class="code"><pre>
+i.eodag -l start=2021-05-25 end=2022-06-01 \
+    dataset=S2_MSI_L2A provider=cop_dataspace \
+    sort=cloudcover,ingestiondate order=asc,desc
+<pre></div>
+
+Search for scenes with a list of IDs text file, and filter the results with the
+provided parameters:
+
+<div class="code"><pre>
+i.eodag -l file=ids_list.txt \
+  area_relation=Contains \
+  clouds=3
+<pre></div>
+
+Download all available scenes in the tmp directory, with Cloud Coverage not
+exceeding 50%:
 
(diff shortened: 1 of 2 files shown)


src/imagery/i.eodag/i.eodag.html
```diff
@@ -96,12 +96,21 @@ <h2>EXAMPLES</h2>
     area_relation=Contains clouds=3
 <pre></div>
 
-Download all available scenes with cloud coverage not exceeding 50% 
+Search and list the available Sentinel 2 scenes in the Copernicus Data Space
+Ecosystem, with relative orbit number 54:
+
+<div class="code"><pre>
+./i.eodag.py -l producttype=S2_MSI_L2A \
```
```suggestion
i.eodag -l producttype=S2_MSI_L2A \
```

src/imagery/i.eodag/i.eodag.py
```diff
@@ -41,21 +43,21 @@
 
 # %flag
 # % key: e
-# % description: Extract the downloaded the datasets, not considered unless provider is set
+# % description: Extract the downloaded the scenes, not considered unless provider is set
```
```suggestion
# % description: Extract the downloaded scenes, not considered unless provider is set
```

src/imagery/i.eodag/i.eodag.py
```diff
@@ -41,21 +43,21 @@
...
 # %end
 
+
 # OPTIONS
 # %option
-# % key: dataset
+# % key: producttype
 # % type: string
-# % description: Imagery dataset to search for
+# % description: Imagery producttype to search for
```
```suggestion
# % description: Imagery product type to search for
```

src/imagery/i.eodag/i.eodag.py
```diff
@@ -146,6 +155,13 @@
 # % guisection: Sort
 # %end
 
+# %option
+# % key: query
+# % label: Extra searching parameters to use in query
```
```suggestion
# % label: Extra searching parameters to use in the search
```

src/imagery/i.eodag/i.eodag.py
```diff
@@ -146,6 +155,13 @@
...
+# % description: Note: Make sure to use provided options when possible, otherwise the values mignt not be recognized
```
veroandreo: What is the message here? I do not understand

src/imagery/i.eodag/i.eodag.py
```diff
@@ -165,12 +181,23 @@
 # % type: string
 # % description: File name to save in (the format will be adjusted according to the file extension)
 # % label: Supported files extensions [geojson: Rreadable by i.eodag | json: Beautified]
```
```suggestion
# % label: Supported file extensions [geojson: Readable by i.eodag | json: Beautified]
```

src/imagery/i.eodag/i.eodag.py
```diff
@@ -165,12 +181,23 @@
 # % type: string
 # % description: File name to save in (the format will be adjusted according to the file extension)
```
```suggestion
# % description: File name to save the search results in (the format will be adjusted according to the file extension)
```

src/imagery/i.eodag/i.eodag.py
```diff
@@ -620,6 +654,91 @@ def products_compare(first, second):
     return search_result
 
 
+def save_footprints(search_result, map_name):
+    """Save products footprints as a vector map in the current mapset.
+
+    Reprojection is done on the fly.
+
+    :param search_result: EO products to be sorted
```
veroandreo: Isn't this the search result per se, i.e., whatever is obtained from eodag after performing a search?

src/imagery/i.eodag/i.eodag.py
```diff
@@ -620,6 +654,91 @@ def products_compare(first, second):
     return search_result
 
 
+def save_footprints(search_result, map_name):
```
veroandreo, line 640: AFAIU from the eodag workshop, the search results can be saved as geojson. For the footprints vector map, isn't it much easier to get that geojson with all attributes and directly import it into grass?? I believe that is more straightforward and we offer all attr for each footprint, i.e. if the user wants the footprint within grass, just use the save_search_result function below to get a geojson and then v.import to import into GRASS. Also, users might want all attr (instead of a selection of them) to carry out a filtering with (v).db.* tools within grass. 

@lucadelu @ninsbl what do you think?

src/imagery/i.eodag/i.eodag.py
veroandreo: What does it mean extract in this context and why is it dependent on setting a provider?

src/imagery/i.eodag/i.eodag.py
```diff
@@ -41,21 +43,21 @@
 
 # %flag
 # % key: e
-# % description: Extract the downloaded the datasets, not considered unless provider is set
+# % description: Extract the downloaded the scenes, not considered unless provider is set
 # %end
 
 # %flag
 # % key: d
 # % description: Delete the product archive after downloading, not considered unless provider is set
```
veroandreo: What product will be deleted? and why is it dependent on setting a provider? i.eodag only downloads products, it won't be importing them, so why would the user use this flag to delete the downloaded files? Such flag makes more sense when a tool unzips and imports data into grass, but IMO, that's out of scope for i.eodag

veroandreo (commented):
I left some comments and questions regarding flags and footprint option. Also, several docstrings have the "EO products to be sorted" as parameter description, it doesn't seem right in all cases, it is maybe a left over of copy-pasting, please have a look.

Also some more use cases and examples would be desirable in the html

src/imagery/i.eodag/i.eodag.py
HamedElgizery: Scenes are downloaded as ZIP files, and **extract** here just means whether to unzip these files or not.

It is dependent on the provider because each provider has its own extract configuration that has to be set to either True or False.

Nevertheless, I have found a way to make the `extract`, `delete_archive`, and `output` options independent of the provider. 
Here https://github.com/OSGeo/grass-addons/pull/1136/commits/8870f158b62b9c934f6621b08332af9d8b8183f8

src/imagery/i.eodag/i.eodag.py
HamedElgizery: Setting `delete_archive` to `Yes` will delete the ZIP files  if they are extracted. (ZIP files aka. archive files in the eodag context)

So for example you could say `extract=Yes` `delete_archive=No` that will unzip the zip files without deleting them.
Another case would be let's say, `extract=Yes` `delete_archive=Yes` here it will unzip the files and delete the archive.

I added these `extract` and `delete_archive` options because it is an integral part of the eodag config file, so if the user messes something up in the `edoag.yml`, we could use these options to guarantee the intended behavior.

src/imagery/i.eodag/i.eodag.py
HamedElgizery: What I am trying to convey is that if you are able to use one of the provided i.eodag options e.g. `start`, `end`, `producttype`, `provider`, etc... Then you should use them instead of passing them through the query option, because they might cause unexpected problems.

Shall I just type this, if it is more clear? or is it too wordy?

src/imagery/i.eodag/i.eodag.py
HamedElgizery: Yup, copy-paste error. 😃 
Fixed.

src/imagery/i.eodag/i.eodag.py
ninsbl: Sounds reasonable to me, so I tend to agree...

src/imagery/i.eodag/i.eodag.py
```diff
@@ -143,34 +154,79 @@
...
 # % answer: asc
+# % required: no
 # % guisection: Sort
 # %end
 
+# %option
+# % key: query
+# % multiple: yes
+# % label: Extra searching parameters to use in the search
+# % description: Note: Make sure to use provided options when possible, otherwise the values mignt not be recognized
```
```suggestion
# % description: Note: Make sure to use provided options when possible, otherwise the values might not be recognized
```

src/imagery/i.eodag/i.eodag.py
```diff
@@ -143,34 +154,79 @@
...
+# % required: no
+# % guisection: Filter
+# %end
+
+# %option
+# % key: extract
+# % type: string
+# % description: Whether to extract downloaded scenes zip files or not (will override values set in the config file)
+# % options: Yes,No
```
ninsbl: Why do you implant yes/no with an option, when it simply can be (and was) a flag?

src/imagery/i.eodag/i.eodag.py
```diff
@@ -143,34 +154,79 @@
...
+# % guisection: Config
+# %end
+
+# %option
+# % key: deletearchive
+# % type: string
+# % label: Whether to delete downloaded scenes zip files or not (will override values set in the config file)
+# % options: Yes,No
```
ninsbl: Same as with `extract`

src/imagery/i.eodag/i.eodag.py
```diff
@@ -213,11 +270,12 @@ def get_bb(proj):
...
-            "lonmin": info["nw_long"],
-            "latmin": info["sw_lat"],
-            "lonmax": info["ne_long"],
-            "latmax": info["nw_lat"],
+            "lonmin": float(info["nw_long"]),
+            "latmin": float(info["sw_lat"]),
+            "lonmax": float(info["ne_long"]),
+            "latmax": float(info["nw_lat"]),
         }
+    print("HERE")
```
```suggestion
```

src/imagery/i.eodag/i.eodag.py
```diff
@@ -649,6 +765,139 @@ def save_search_result(search_result, file_name):
...
+        gs.message(_("Recognized products offered by {}".format(provider)))
+    else:
+        gs.message(_("Recongnizaed providers"))
+    print(json.dumps({"products": dag.list_product_types(provider or None)}, indent=4))
+
+
+def print_eodag_queryables(**kwargs):
+    """Print queryables info for given provider and/or product type in JSON format.
+
+    :param kwargs: Presit parameters values.
```
ninsbl: Presit?

src/imagery/i.eodag/i.eodag.py
```diff
@@ -660,21 +909,45 @@ def main():
 
     dates_to_iso_format()
 
+    if options["print"]:
```
ninsbl, line 849: Instead of many if-else you could use a dictionary where the print functions are the "values"...

ninsbl (changes requested):
Currently, I have to review on a mobile phone, which is somewhat limiting. So, i may have overlooked something. Anyway, please see my comments in the code...

src/imagery/i.eodag/i.eodag.py
HamedElgizery: Because I want the module to use the config file for the extract and the delete archive options by default... and if we want to enforce extract or delete archive to be True or False we can use the grass parser option. But if we use flags here we will only be able to enforce the extract/delete_archive to be True.

src/imagery/i.eodag/i.eodag.py
HamedElgizery: I changed it to -> "options/flags from gs.parser, with the crietria that will be used for filtering."

src/imagery/i.eodag/i.eodag.py
ecodiv: If I understand right, this option is to use searching parameters not provided for by the other filter parameters? Perhaps turning it around:

"Query using extra searching (filter?) parameters not provided for by the provided filter options"

Would it be possible to check if the user uses query parameters that are already available as parameter option, and issue a warning if the user uses these parameters?

veroandreo (approved):
I've just tested different use cases. In generally, it works as expected, except for the query option. I tried with OrbitDirection and start, but nothing gets filtered really with those... so, I'm not sure it actually works or how it is supposed to work (the queryables for S2 will return mostly the same values, because they are like organization, instrument and so on...). Perhaps if multiple products are passed to producttype, it makes more sense??

This is what I tried with queryables:
```
i.eodag -l start=2022-05-01 end=2022-06-01 clouds=50 producttype=S2_MSI_L2A provider=cop_dataspace query="start=2022-05-20"

i.eodag.py -l start=2022-05-01 end=2022-06-01 clouds=50 producttype=S2_MSI_L2A provider=cop_dataspace query="orbitDirection=DESCENDING"
```
Try with and without the query option in NC project with this region setting:
```
n=228510
s=214980
w=630000
e=645000
```
In general, I'm in to merge this one as it got too long already (maybe smaller PRs are easier to review and test), and if there's something else to fix, create a new PR.

src/imagery/i.eodag/i.eodag.py
ninsbl: This dhould be made clear in the manual...

ninsbl:
I think the additional queryables need to be used in the crunch function on the returned initial surchresults, not during initial surch.

As for implementing the queriables, please consider the different operators and possible use of multiple values. Please double check also my earlier suggestions.

ninsbl (approved):
I agree with @veroandreo that we should merge and then improve from the current state.

I still think that however, that extraction and deletion of archives is more appropriately implemented with flags. In any case, if a settings file gets priority over options/flags should be made clear in the manual...

A next step should be to add test cases that do not require login credentials, like for printing, filtering and so on..

HamedElgizery:
> I've just tested different use cases. In generally, it works as expected, except for the query option. I tried with OrbitDirection and start, but nothing gets filtered really with those... so, I'm not sure it actually works or how it is supposed to work (the queryables for S2 will return mostly the same values, because they are like organization, instrument and so on...).

Thanks for the testing! I am currently working on fixing the issue.

> In general, I'm in to merge this one as it got too long already (maybe smaller PRs are easier to review and test), and if there's something else to fix, create a new PR.

Makes sense, I will be pushing the fix on the next PR if we agree on that, which will also extend on the current query functionality.

PR #1140, opened 2024-07-08
r.series.diversity: add options to set nprocs and memory

ecodiv:
Add option to set `nprocs` for parallel computing (used when running r.series) and memory + cleanup code and help page.

### DIFF
--- a/src/raster/r.series.diversity/r.series.diversity.py
+++ b/src/raster/r.series.diversity/r.series.diversity.py
@@ -7,7 +7,7 @@
 # AUTHOR(S):    Paulo van Breugel <paulo ecodiv earth>
 # PURPOSE:      Compute biodiversity indices over input layers
 #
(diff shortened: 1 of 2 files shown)


src/raster/r.series.diversity/r.series.diversity.py
```diff
@@ -115,86 +115,83 @@
...
+
+# %option
+# % key: memory
+# % type: integer
+# % required: no
+# % multiple: no
+# % label: Maximum memory to be used (in MB)
+# % description: Maximum memory to be used (in MB)
+# % answer: 300
+# %end
```
neteler: Please only define changes here with regard to the already defined options:

https://github.com/OSGeo/grass/blob/ff2da11b41e6943ff3d83f4801df203228b81ecd/lib/gis/parser_standard_options.c#L757

ecodiv: Hi @neteler, I completely overlooked the standard `G_OPT_MEMORYMB` option. Thanks. I'll change that (I assume that is what you were referring too?).

PR #1144, opened 2024-07-17
i.landsat.download: change API from landsatxplore to EODAG using i.eodag

HamedElgizery:
This PR will change the API used in i.landsat.download from **landsatxplore** to **EODAG** using the i.**eodag** addon.

### DIFF
--- a/src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
+++ b/src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
@@ -130,11 +130,31 @@
 # %end
 
 import os
+import pytz
 import json
 from datetime import *
 import grass.script as gs
 from grass.pygrass.modules import Module
 
+def normalize_time(datetime_str: str):
+    """Unifies the different ISO formats into 'YYYY-MM-DDTHH:MM:SS'
+
+    :param datetime_str: Datetime in ISO format
+    :type datetime_str: str
+
+    :return: Datetime converted to 'YYYY-MM-DDTHH:MM:SS'
+    :rtype: str
+    """
+    normalized_datetime = datetime.fromisoformat(datetime_str)
+    if normalized_datetime.tzinfo is None:
+        normalized_datetime = normalized_datetime.replace(tzinfo=timezone.utc)
+    # Remove microseconds
+    normalized_datetime = normalized_datetime.replace(microsecond=0)
+    # Convert time to UTC
+    normalized_datetime = normalized_datetime.astimezone(pytz.utc)
+    # Remove timezone info
+    normalized_datetime = normalized_datetime.replace(tzinfo=None)
+    return normalized_datetime.isoformat()
 
 def main():
     start_date = options["start"]
@@ -176,9 +196,22 @@ def main():
                 limit=options["limit"],
             )
         )
-
+        if flags["l"]:
+            for scene in scenes["features"]:
+                product_line = ""
+                product_line = scene["properties"]["landsat:scene_id"]
+                product_line += " " + scene["id"]
+                # Special formatting for datetime
+                try:
+                    acquisition_time = normalize_time(scene["properties"]["startTimeFromAscendingNode"])
+                except:
+                    acquisition_time = scene["properties"]["startTimeFromAscendingNode"]
+                product_line += " " + acquisition_time
+                cloud_cover = scene["properties"]["cloudCover"]
+                product_line += f" {cloud_cover:2.0f}%"
+                print(product_line)
         # TODO: Do extra landsat specifc filtering
-        # TODO: Add list option
+
 
 
 if __name__ == "__main__":
--- a/src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
+++ b/src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
@@ -9,7 +9,7 @@
 # PURPOSE:     Downloads Landsat TM, ETM and OLI data from EarthExplorer
 #              and Planetary Computer using EODAG Python library.
 #
-# COPYRIGHT:   (C) 2020-2025 by Veronica Andreo, and the GRASS development team
+# COPYRIGHT:   (C) 2020-2024 by Veronica Andreo, and the GRASS development team
 #
 #              This program is free software under the GNU General Public
 #              License (>=v2). Read the file COPYING that comes with GRASS
--- a/src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
+++ b/src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
@@ -322,4 +322,16 @@ if __name__ == "__main__":
     except:
         gs.fatal(_("Cannot import eodag. Please intall the library first."))
 
+    # Check if eodag config file is created
+    file_path = os.path.join(os.path.expanduser("~"), ".config/eodag/eodag.yml")
+    if not os.path.isfile(file_path):
+        dag = EODataAccessGateway()
+        gs.info(
+            _(
+                "EODAG Config file is created, you can rerun the module after filling the necessary credentials in {}".format(
+                    file_path
+                )
+            )
+        )
+        sys.exit(0)
     sys.exit(main())
--- a/src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
+++ b/src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
@@ -39,6 +39,7 @@
 # %option
 # % key: clouds
 # % type: integer
+# % answer: 100
 # % description: Maximum cloud cover percentage for Landsat scene
 # % required: no
 # % guisection: Filter
@@ -187,50 +188,53 @@ def main():
 
     # Download by ID
     if options["id"]:
-        gs.run_command("i.eodag", id=options["id"], output=outdir)
+        # Should use other provider other than USGS,
+        # as there was a bug in USGS API, fixed here
+        # https://github.com/CS-SI/eodag/issues/1252
+        # TODO: set provider to USGS when the above changes goes into production
+        gs.run_command(
+            "i.eodag", id=options["id"], output=outdir, provider="planetary_computer"
+        )
     else:
         # TODO: Map dataset to eodag productType
         eodag_producttype = "LANDSAT_C2L2"
-        eodga_sort = ""
+        eodag_sort = ""
         for sort_var in options["sort"].split(","):
             if sort_var == "cloud_cover":
-                eodga_sort += "cloudcover"
+                eodag_sort += "cloudcover,"
             if sort_var == "acquisition_date":
-                eodga_sort += "ingestiondate"
+                eodag_sort += "ingestiondate,"
+        eodag_sort = eodag_sort[:-1]
 
         scenes = json.loads(
             gs.read_command(
                 "i.eodag",
                 flags="j",
                 producttype=eodag_producttype,
-                map=options["map"],
+                map=options["map"] if options["map"] else None,
                 start=start_date,
                 end=end_date,
-                clouds=options["clouds"],
+                clouds=options["clouds"] if options["clouds"] else None,
                 limit=options["limit"],
                 order=options["order"],
-                sort=eodga_sort,
+                sort=eodag_sort,
+                provider="planetary_computer",
+                query=(
+                    f"landsat:collection_category={options['tier']}"
+                    if options["tier"]
+                    else None
+                ),
                 # Each provider provides data in differet format,
-                # so using USGS for consistency
-                provider="usgs",
+                # so using Microsoft Planetary Computer for consistency
                 quiet=True,
             )
         )
-        if options["tier"]:
-            scenes["features"] = list(
-                filter(lambda scene: options["tier"] in scene["id"], scenes["features"])
-            )
-        scenes["features"] = scenes["features"][
-            : min(len(scenes["features"]), int(options["limit"]))
-        ]
 
         # Output number of scenes found
         gs.message(_("{} scenes found.".format(len(scenes["features"]))))
 
-        # print(json.dumps(scenes, indent=4))
         if flags["l"]:
             for scene in scenes["features"]:
-                product_line = ""
                 product_line = scene["properties"]["landsat:scene_id"]
                 product_line += " " + scene["id"]
                 # Special formatting for datetime
@@ -259,7 +263,10 @@ def main():
             with open(geojson_temp_file, "w") as file:
                 file.write(json.dumps(scenes))
             gs.run_command(
-                "i.eodag", file=geojson_temp_file, provider="usgs", output=outdir
+                "i.eodag",
+                file=geojson_temp_file,
+                provider="planetary_computer",
+                output=outdir,
             )
 
 
--- a/src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
+++ b/src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
@@ -197,7 +197,7 @@ def main():
         # https://github.com/CS-SI/eodag/issues/1252
         # TODO: set provider to USGS when the above changes goes into production
         if options["datasource"] == "usgs" and not (
-            eodag.__version__ == "3.0.0b3" or eodag.__version__ >= "3.0.0"
+            int(eodag.__version__.split(".")[0]) >= 3
         ):
             gs.fatal(
                 _(
--- a/src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
+++ b/src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
@@ -227,38 +227,23 @@ def main():
             gs.fatal(_("Dataset was not recognized"))
 
         eodag_sort = ""
-        eodag_query = ""  # Used with Planetary Computer
-        eodag_pattern = ""  # Used with USGS
+        eodag_pattern = ""
         for sort_var in options["sort"].split(","):
             if sort_var == "cloud_cover":
                 eodag_sort += "cloudcover,"
             if sort_var == "acquisition_date":
                 eodag_sort += "ingestiondate,"
 
-        if options["datasource"] == "planetary_computer":
-            if options["tier"]:
-                eodag_query += f"landsat:collection_category={options['tier']},"
-            if "tm" in options["dataset"]:
-                eodag_query += "platformSerialIdentifier=landsat-5"
-            if "etm" in options["dataset"]:
-                eodag_query += "platformSerialIdentifier=landsat-7"
-            if "8_ot" in options["dataset"]:
-                eodag_query += "platformSerialIdentifier=landsat-8"
-            if "9_ot" in options["dataset"]:
-                eodag_query += "platformSerialIdentifier=landsat-9"
-
-        elif options["datasource"] == "usgs":
-            eodag_pattern = ""
-            if "tm" in options["dataset"]:
-                eodag_pattern += "LM05.+"
-            if "etm" in options["dataset"]:
-                eodag_pattern += "LE07.+"
-            if "8_ot" in options["dataset"]:
-                eodag_pattern += "LC08.+"
-            if "9_ot" in options["dataset"]:
-                eodag_pattern += "LC09.+"
-            if options["tier"]:
-                eodag_pattern += options["tier"]
+        if "tm" in options["dataset"]:
+            eodag_pattern += "LM05.+"
+        if "etm" in options["dataset"]:
+            eodag_pattern += "LE07.+"
+        if "8_ot" in options["dataset"]:
+            eodag_pattern += "LC08.+"
+        if "9_ot" in options["dataset"]:
+            eodag_pattern += "LC09.+"
+        if options["tier"]:
+            eodag_pattern += options["tier"]
 
         try:
             scenes = json.loads(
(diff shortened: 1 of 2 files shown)


src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
```diff
@@ -55,8 +50,8 @@
 # % type: string
 # % description: Landsat dataset to search for
 # % required: no
-# % options: landsat_tm_c1, landsat_etm_c1, landsat_8_c1, landsat_tm_c2_l1, landsat_tm_c2_l2, landsat_etm_c2_l1, landsat_etm_c2_l2, landsat_ot_c2_l1, landsat_ot_c2_l2
-# % answer: landsat_8_c1
+# % options: landsat_tm_c2_l1, landsat_tm_c2_l2, landsat_etm_c2_l1, landsat_etm_c2_l2, landsat_8_ot_c2_l1, landsat_8_ot_c2_l2, landsat_9_ot_c2_l1, landsat_9_ot_c2_l2
+# % answer: landsat_9_ot_c2_l1
```
veroandreo:
```suggestion
# % answer: landsat_9_ot_c2_l2
```

Better have the L2 data as default which is surface reflectance already

src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
```diff
@@ -280,34 +270,21 @@ def main():
...
+                file=geojson_temp_file,
+                provider="planetary_computer",
+                output=outdir,
+            )
 
 
 if __name__ == "__main__":
     options, flags = gs.parser()
 
-    # lazy import
```
veroandreo, line 304: Shouldn't we do lazy import of eodag here and tell users that both eodag and i.eodag need to be installed for i.landsat.download to work?

HamedElgizery:
In order to test this branch, you need to update your i.eodag module with the following PR https://github.com/OSGeo/grass-addons/pull/1145, because there is an update in i.eodag (adding regex pattern option), on which searching/downloading with USGS depends.

veroandreo (commented):
Listing products seems to work fine. 

```
i.landsat.download -l dataset=landsat_8_ot_c2_l2 clouds=5 start="2022-12-23" end="2023-06-23"
         guaranteed to be reliable
14 scenes found.
LC81930292023010LGN00 LC08_L2SP_193029_20230110_02_T1 2023-01-10T10:05:19  0%
LC81930302023042LGN00 LC08_L2SP_193030_20230211_02_T1 2023-02-11T10:05:39  0%
LC81930302023010LGN00 LC08_L2SP_193030_20230110_02_T1 2023-01-10T10:05:43  1%
LC81910292023076LGN00 LC08_L2SP_191029_20230317_02_T1 2023-03-17T09:52:31  1%
LC81940292023033LGN00 LC08_L2SP_194029_20230202_02_T1 2023-02-02T10:11:30  2%
LC81960282023047LGN00 LC08_L2SP_196028_20230216_02_T1 2023-02-16T10:23:20  2%
LC81950272023040LGN00 LC08_L2SP_195027_20230209_02_T1 2023-02-09T10:16:51  2%
LC81910292023044LGN00 LC08_L2SP_191029_20230213_02_T1 2023-02-13T09:52:52  2%
LC81910302023076LGN00 LC08_L2SP_191030_20230317_02_T1 2023-03-17T09:52:55  2%
LC81930302023074LGN00 LC08_L2SP_193030_20230315_02_T1 2023-03-15T10:05:18  2%
LC81910282023044LGN00 LC08_L2SP_191028_20230213_02_T1 2023-02-13T09:52:28  3%
LC81930272022359LGN00 LC08_L2SP_193027_20221225_02_T1 2022-12-25T10:04:32  3%
LC81940272023081LGN00 LC08_L2SP_194027_20230322_02_T1 2023-03-22T10:10:11  4%
LC81960282023031LGN00 LC08_L2SP_196028_20230131_02_T1 2023-01-31T10:23:28  4%
To download all scenes found, re-run the previous command without -l flag.
Note that if no output option is provided, files will be downloaded in /tmp
```

However, download with id yields the following in all my attempts (also which is the id to use? maybe when printing we could indicate that?):

```
i.landsat.download id=LC81930292023010LGN00 output=/tmp
  File "/home/vandreo/software/grass-addons/src/imagery/i.landsat/i.landsat.download/i.landsat.download.py", line 337, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/vandreo/software/grass-
addons/src/imagery/i.landsat/i.landsat.download/i.landsat.download.py", line 283, in main
    "i.eodag", id=options["id"], output=outdir, provider=options["provider"]
                                                         ~~~~~~~^^^^^^^^^^^^
KeyError: 'provider'

i.landsat.download id=LC08_L2SP_193029_20230110_02_T1 output=/tmp
  File "/home/vandreo/software/grass-addons/src/imagery/i.landsat/i.landsat.download/i.landsat.download.py", line 337, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/vandreo/software/grass-addons/src/imagery/i.landsat/i.landsat.download/i.landsat.download.py", line 283, in main
    "i.eodag", id=options["id"], output=outdir, provider=options["provider"]
                                                         ~~~~~~~^^^^^^^^^^^^
KeyError: 'provider'

i.landsat.download id=LC81910302023076LGN00 output=/tmp datasource=planetary_computer
  File "/home/vandreo/software/grass-addons/src/imagery/i.landsat/i.landsat.download/i.landsat.download.py", line 337, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/vandreo/software/grass-addons/src/imagery/i.landsat/i.landsat.download/i.landsat.download.py", line 283, in main
    "i.eodag", id=options["id"], output=outdir, provider=options["provider"]
                                                         ~~~~~~~^^^^^^^^^^^^
KeyError: 'provider'
```

HamedElgizery:
> However, download with id yields the following in all my attempts (also which is the id to use? maybe when printing we could indicate that?):

Bug fixed in the last commit [82096e8](https://github.com/OSGeo/grass-addons/pull/1144/commits/82096e80c734ae6e309c1979f8c35895ab995037)

You need to use the expanded ID (EODAG calls it the title), so this one: LC08_L2SP_193029_20230110_02_T1

veroandreo:
> > However, download with id yields the following in all my attempts (also which is the id to use? maybe when printing we could indicate that?):
> 
> Bug fixed in the last commit [82096e8](https://github.com/OSGeo/grass-addons/pull/1144/commits/82096e80c734ae6e309c1979f8c35895ab995037)
> 
> You need to use the expanded ID (EODAG calls it the title), so this one: LC08_L2SP_193029_20230110_02_T1

Different error now:
```
i.landsat.download id=LC08_L2SP_193029_20230110_02_T1
         guaranteed to be reliable
Found 1 distinct ID(s).
LC08_L2SP_193029_20230110_02_T1
  File "/home/vandreo/.grass8/addons/scripts/i.eodag", line 1152, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/vandreo/.grass8/addons/scripts/i.eodag", line 1055, in main
    search_result = search_by_ids(ids_set)
                    ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vandreo/.grass8/addons/scripts/i.eodag", line 359, in search_by_ids
    product, count = dag.search(id=query_id, provider=options["provider"] or None)
    ^^^^^^^^^^^^^^
ValueError: not enough values to unpack (expected 2, got 1)
  File "/home/vandreo/software/grass-addons/src/imagery/i.landsat/i.landsat.download/i.landsat.download.py", line 415, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/vandreo/software/grass-addons/src/imagery/i.landsat/i.landsat.download/i.landsat.download.py", line 361, in main
    gs.run_command(
  File "/home/vandreo/software/grass-dev/dist.x86_64-pc-linux-gnu/etc/python/grass/script/core.py", line 485, in run_command
    return handle_errors(returncode, result=None, args=args, kwargs=kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vandreo/software/grass-dev/dist.x86_64-pc-linux-gnu/etc/python/grass/script/core.py", line 364, in handle_errors
    raise CalledModuleError(module=module, code=code, returncode=returncode)
grass.exceptions.CalledModuleError: Module run `i.eodag id=LC08_L2SP_193029_20230110_02_T1 output=/tmp provider=planetary_computer` ended with an error.
The subprocess ended with a non-zero return code: 1. See errors above the traceback or in the error output.
```

HamedElgizery:
> Different error now:

It was a problem in i.eodag, because of EODAG beta version 3.0.0b1... They added a new parameter `count` to their search method.
The fix is in this PR https://github.com/OSGeo/grass-addons/pull/1145.

src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
```diff
@@ -136,55 +147,331 @@
...
+    """Unifies the different ISO formats into 'YYYY-MM-DDTHH:MM:SS'
+
+    :param datetime_str: Datetime in ISO format
+    :type datetime_str: str
+
+    :return: Datetime converted to 'YYYY-MM-DDTHH:MM:SS'
+    :rtype: str
+    """
+    normalized_datetime = datetime.fromisoformat(datetime_str)
+    if normalized_datetime.tzinfo is None:
```
ninsbl, line 167: I guess the whole function can be simplified to:

```
normalized_datetime = datetime.fromisoformat(datetime_str)
return normalized_datetime.astimezone(timezone.utc).strftime("%Y-%m-%dT%H:%M:%S")
```
Or if you want to treat missing tzinfo as UTC instead of local time use:
```
normalized_datetime = datetime.fromisoformat(datetime_str)
if normalized_datetime.tzinfo is None:
     return normalized_datetime.strftime("%Y-%m-%dT%H:%M:%S")
return normalized_datetime.astimezone(timezone.utc).strftime("%Y-%m-%dT%H:%M:%S")
```

src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
```diff
@@ -136,55 +147,331 @@
...
+        normalized_datetime = normalized_datetime.replace(tzinfo=timezone.utc)
+    # Remove microseconds
+    normalized_datetime = normalized_datetime.replace(microsecond=0)
+    # Convert time to UTC
+    normalized_datetime = normalized_datetime.astimezone(pytz.utc)
```
ninsbl: Why using pytz here instead of timezone

src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
```diff
@@ -136,55 +147,331 @@
 # %end
 
 import os
+import pytz
```
ninsbl: I guess you can nowadays use builtin libraries instead of pytz, see e.g. first note here:
https://pypi.org/project/pytz/
```suggestion
```
In the datetime import you should replace the * import with explicit imports (datetime, timedelta(?), timmezone)

src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
```diff
@@ -136,55 +147,331 @@
...
+    def list(self):
+        """Should list scenes saved in self.search_result"""
+        raise NotImplementedError
+
+    def download(self, output_directory):
+        """Should download scenes saved in self.search_result"""
+        raise NotImplementedError
+
+
+class PlanetaryComputerLandsat(LandsatDownloader):
```
ninsbl: This basically looks like code duplication to me.
Do you really have to differenciate the providers like this, with different classes?
Isn`t it enough to call _i.eodag_ with the right options and _flags_

src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
```diff
@@ -206,108 +492,74 @@ def main():
...
+    if not os.path.isfile(file_path):
+        dag = EODataAccessGateway()
+        gs.info(
+            _(
+                "EODAG Config file is created, you can rerun the module after filling the necessary credentials in {}".format(
+                    file_path
+                )
+            )
+        )
+        sys.exit(0)
```
ninsbl: Is it necessary to exit here?
This would be a show stopper if it is used in e.g. a cloud setting where fressh pod are started to perform downloading...
We should not make the module(s) overly dependent on the configfile. When I experimented with EODAG I never touched the config file, so it does not seem strictly necessary. And my guess is that for the average user modifying YAML files is a significant hurdle...

src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
```diff
@@ -136,55 +147,331 @@
 # %end
 
 import os
+import pytz
+import json
+import sys
 from datetime import *
 import grass.script as gs
+from grass.pygrass.modules import Module
```
ninsbl: This does not seem to be used, yet.

You can replace both `gs.read_command()` and `gs.run_command()` with `Module()`

src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
```diff
@@ -206,108 +492,74 @@ def main():
...
-    except ImportError:
-
-        gs.fatal(_("Cannot import landsatxplore. Please install the library first."))
-
-    main()
+        import eodag
+        from eodag import EODataAccessGateway
+    except:
+        gs.fatal(_("Cannot import eodag. Please intall the library first."))
+
```
ninsbl, line 342: You should check if _i.eodag_ is installed with e.g.: [find_program](https://grass.osgeo.org/grass83/manuals/libpython/grass.script.html?highlight=find_program#grass.script.core.find_program)

```suggestion
try:
    gs.find_program("i.eodag", "--help")
except ImportError:
    gs.fatal(_("Addon i.eodag not found. Please intall it with g.extension."))
```
Alternatively, if you use `Module()` in _main()_, you could do something along the lines of:

```
try:
    i_eodag = Module("i.eodag", run_=False)
except ImportError:
    gs.fatal(_("Addon i.eodag not found. Please intall it with g.extension."))
```

src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
```diff
@@ -206,108 +492,74 @@ def main():
...
+        eodag_order=options["order"],
+    )
+
+    if flags["l"]:
+        downloader.list()
+        gs.message(
+            _(
+                "To download all scenes found, re-run the previous "
+                "command without -l flag. Note that if no output "
+                "option is provided, files will be downloaded in /tmp"
```
ninsbl: As Markus mentioned in another comment downloading to `/tmp` by default is at least unexpected.
Usually rather the current working directory is used as a default.
That said, I suggest to to remove the message. It does not provide any specific information that is not already in the manual...

ninsbl (changes requested):
My understanding is that _i.landsat.download_ is supposed to be a wapper module around _i.eodag_.

I therefore suggest to significantly simplify the code by just calling `Module("i.eodag", start=...)` with the appropriate options and flags. I do not see the need for e.g. downloader classes...
You could probably create a dict that translates unified landsat options or old _i.landsat.download_ options to the suitable options for the different providers...

ninsbl:
BTW: Do you use _pre-commit_ hooks? If not, I recommend installing them with `python3 -m pip install pre-commit`. That way your code is quality checked on each commit...

src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
lucadelu: Yeah I also was thinking that i.landsat.download should just call i.eodag with the right parameters...

HamedElgizery: @ninsbl The json output differs depending on the provider, and also the way we query in i.eodag... That is why they need to be separated... though we can simplify that and make them just two separate functions (one for each provider) which was the case before the last commit. 
Like in this commit: [82096e8](https://github.com/OSGeo/grass-addons/pull/1144/commits/82096e80c734ae6e309c1979f8c35895ab995037)

HamedElgizery:
> BTW: Do you use _pre-commit_ hooks? If not, I recommend installing them with `python3 -m pip install pre-commit`. That way your code is quality checked on each commit...

Thanks noted.
I switched machines and forgot to use the pre-commit on the new one.

HamedElgizery:
> My understanding is that _i.landsat.download_ is supposed to be a wapper module around _i.eodag_.
> 
> I therefore suggest to significantly simplify the code by just calling `Module("i.eodag", start=...)` with the appropriate options and flags. I do not see the need for e.g. downloader classes... You could probably create a dict that translates unified landsat options or old _i.landsat.download_ options to the suitable options for the different providers...

Thanks for the review @ninsbl.
The main problem is that each provider will have a different way to query their products... 
That is why we need to separate them in functions or classes. I did separate them in functions at first, but I figured it wouldn't be convenient to add more providers later on. The classes should, supposedly, provide a more clear path of what needs to be implemented for the code to work with the new provider, yet I agree there might be some code duplication, and maybe it is not worth it.
Anyway, there is also an implementation that can have everything handled in one function in a sort of a clean manner.
Shall we give that a try, or maybe at least have them in a separate functions?

src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
HamedElgizery: Nice, I think we shall go with the later because typically all the scenes use UTC.

src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
HamedElgizery: I think we can stick with `gs.read_command()` & `gs.run_command()` for simplicity, if `Module()` doesn't have a significant advantage.

src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
HamedElgizery: Generally the config file is necessary especially when using USGS.
For Planetary Computer, you can go fine most of the time without credentials.

We can remove that exit part, but it will give an error that the user needs to fill in the USGS credentials (if they were attempting to use USGS).


> This would be a show stopper if it is used in e.g. a cloud setting where fressh pod are started to perform downloading...

I see, you are right that would be annoying if it could just go fine without the credentials, and we force the user to rerun again in a different pod. I will remove it for now... Also, I will put the `settings` option back to i.landsat.download once I add it in i.eodag...

src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
```diff
@@ -1,17 +1,21 @@
...
-# AUTHOR(S):   Veronica Andreo
+# AUTHOR(S):   Veronica Andreo (original contributor)
+#              Hamed Elgizery <hamedashraf2004 gmail.com>
+#
 # PURPOSE:     Downloads Landsat TM, ETM and OLI data from EarthExplorer
-#              using landsatxplore Python library.
-# COPYRIGHT:   (C) 2020-2021 by Veronica Andreo, and the GRASS development team
+#              and Planetary Computer using EODAG Python library.
+#
+# COPYRIGHT:   (C) 2020-2025 by Veronica Andreo, and the GRASS development team
```
ninsbl:
```suggestion
# COPYRIGHT:   (C) 2020-2024 by Veronica Andreo, and the GRASS development team
```

src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
```diff
@@ -204,110 +188,174 @@ def main():
...
-            try:
-
-                ee.download(
-                    identifier=i, output_dir=outdir, timeout=int(options["timeout"])
+        # Should use other provider other than USGS,
+        # as there was a bug in USGS API, fixed here
+        # https://github.com/CS-SI/eodag/issues/1252
+        # TODO: set provider to USGS when the above changes goes into production
+        if options["datasource"] == "usgs" and not (
+            eodag.__version__ == "3.0.0b3" or eodag.__version__ >= "3.0.0"
```
ninsbl: You should check versions as numbers or tuples of numbers, assuming semantic versioning. I do not think >= works with strings. Also no need to check for specific beta versions...

ninsbl (changes requested):
Looks good. Two minor comments. Ready to merge when they are addressed...

src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
```diff
@@ -204,110 +188,174 @@ def main():
...
-
-                ee.download(
-                    identifier=i, output_dir=outdir, timeout=int(options["timeout"])
+        # Should use other provider other than USGS,
+        # as there was a bug in USGS API, fixed here
+        # https://github.com/CS-SI/eodag/issues/1252
+        # TODO: set provider to USGS when the above changes goes into production
+        if options["datasource"] == "usgs" and not (
+            int(eodag.__version__.split(".")[0]) >= 3
+        ):
```
ninsbl:
```suggestion
        if options["datasource"] == "usgs" and int(eodag.__version__.split(".")[0]) < 3:
```

ninsbl (approved):

PR #1145, opened 2024-07-18
i.eodag: add timeout & wait options

HamedElgizery:
Adding timeout & wait options, which would be specially helpful when encountering OFFLINE products.

### DIFF
--- a/src/imagery/i.eodag/i.eodag.py
+++ b/src/imagery/i.eodag/i.eodag.py
@@ -98,6 +98,14 @@
 # % guisection: Region
 # %end
 
+# %option
+# % key: pattern
+# % type: string
+# % description: Filter products by title name (using regex)
+# % required: no
+# % guisection: Region
+# %end
+
 # %option G_OPT_V_OUTPUT
 # % key: footprints
 # % description: Name for output vector map with footprints
@@ -725,6 +733,11 @@ def filter_result(search_result, geometry=None, queryables=None, **kwargs):
                     ).data
                     tmp_search_result_list.extend(filtered_search_result_list)
             search_result = SearchResult(tmp_search_result_list)
+
+    if options["pattern"]:
+        import re
+        pattern = re.compile(options["pattern"])
+        search_result = SearchResult(filter(lambda p: pattern.fullmatch(p.properties["title"]), search_result))
 
     # Remove duplictes that might be created while filtering
     search_result = remove_duplicates(search_result)
--- a/src/imagery/i.eodag/i.eodag.py
+++ b/src/imagery/i.eodag/i.eodag.py
@@ -103,7 +103,7 @@
 # % type: string
 # % description: Filter products by title name (using regex)
 # % required: no
-# % guisection: Region
+# % guisection: Filter
 # %end
 
 # %option G_OPT_V_OUTPUT
--- a/src/imagery/i.eodag/i.eodag.py
+++ b/src/imagery/i.eodag/i.eodag.py
@@ -736,8 +736,11 @@ def filter_result(search_result, geometry=None, queryables=None, **kwargs):
 
     if options["pattern"]:
         import re
+
         pattern = re.compile(options["pattern"])
-        search_result = SearchResult(filter(lambda p: pattern.fullmatch(p.properties["title"]), search_result))
+        search_result = SearchResult(
+            filter(lambda p: pattern.fullmatch(p.properties["title"]), search_result)
+        )
 
     # Remove duplictes that might be created while filtering
     search_result = remove_duplicates(search_result)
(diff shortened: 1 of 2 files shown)


veroandreo (commented):
Did you find a test case for these options, i.e. a product that needs the wait in order to test it effectively works? Otherwise, it looks good to go

src/imagery/i.eodag/i.eodag.py
```diff
@@ -98,6 +98,14 @@
 # % guisection: Region
 # %end
 
+# %option
+# % key: pattern
+# % type: string
+# % description: Filter products by title name (using regex)
+# % required: no
+# % guisection: Region
```
veroandreo: Shouldn't this be filter section better?

Also, we will need an example in the manual page with this

src/imagery/i.eodag/i.eodag.py
```diff
@@ -98,6 +98,14 @@
 # % guisection: Region
 # %end
 
+# %option
+# % key: pattern
+# % type: string
+# % description: Filter products by title name (using regex)
```
```suggestion
# % description: Filter products by file name using a regular expression, e.g. 'S1.*_SDV.*'
```
An example would help in the description, e.g. as suggested above, but you may have a more relevant one...

src/imagery/i.eodag/i.eodag.py
```diff
@@ -710,6 +734,14 @@ def filter_result(search_result, geometry=None, queryables=None, **kwargs):
                     tmp_search_result_list.extend(filtered_search_result_list)
             search_result = SearchResult(tmp_search_result_list)
 
+    if options["pattern"]:
+        import re
+
+        pattern = re.compile(options["pattern"])
+        search_result = SearchResult(
+            filter(lambda p: pattern.fullmatch(p.properties["title"]), search_result)
```
```suggestion
            filter(lambda p: pattern.fullmatch(p.properties["title"]), search_result)
```
It is suggested to use list comprehensions instead of lambda expression. See e.g.:
https://stackoverflow.com/questions/1632902/lambda-versus-list-comprehension-performance

Speaking of code-style. I reccomend checking your code with ruff, it will give you suggestions for how code-style could be improved. Checking code with ruff is quite instructive. I learned a lot from it. What you do is roughly the following:
```
python3 -m pip install ruff
ruff check --preview i.eodag.py
```

The point is not to do everything ruff suggests but to choose the suggestions that seem reasonable....

src/imagery/i.eodag/i.eodag.py
```diff
@@ -710,6 +734,14 @@ def filter_result(search_result, geometry=None, queryables=None, **kwargs):
                     tmp_search_result_list.extend(filtered_search_result_list)
             search_result = SearchResult(tmp_search_result_list)
 
+    if options["pattern"]:
+        import re
```
ninsbl: You can import the builtin at the module top...

ninsbl (commented):
This looks overall good. Please check my one comment about the option description...

src/imagery/i.eodag/i.eodag.py
```diff
@@ -356,7 +380,7 @@ def search_by_ids(products_ids):
     search_result = []
     for query_id in products_ids:
         gs.verbose(_("Searching for {}".format(query_id)))
-        product, count = dag.search(id=query_id, provider=options["provider"] or None)
+        product, count = dag.search(id=query_id, provider=options["provider"] or None, count=True)
```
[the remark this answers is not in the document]
ninsbl: The black test still fails...

src/imagery/i.eodag/i.eodag.py
ninsbl: BTW. the term "title name" is a bit confusing, at least to me. Is "file name" a correct interpretation?

src/imagery/i.eodag/i.eodag.py
HamedElgizery: > It is suggested to use list comprehensions instead of lambda expression. See e.g.: https://stackoverflow.com/questions/1632902/lambda-versus-list-comprehension-performance

On it.
> Speaking of code-style. I reccomend checking your code with ruff, it will give you suggestions for how code-style could be improved. Checking code with ruff is quite instructive. I learned a lot from it. What you do is roughly the following:
> 
> ```
> python3 -m pip install ruff
> ruff check --preview i.eodag.py
> ```
> 
> The point is not to do everything ruff suggests but to choose the suggestions that seem reasonable....

Nice thanks, it caught some typos/bugs.

src/imagery/i.eodag/i.eodag.py
HamedElgizery: There are two terms:
Title: Name of the product on a provider, and it can vary for different providers...
UUID: Unified for a given product across different providers.

Title is actually what we pass to the `id` option, when we are trying to download products by their id.
So I will just change it to: 

> Filter products by id using a regular expression, e.g. 'LC09.*T1'

HamedElgizery:
> Did you find a test case for these options, i.e. a product that needs the wait in order to test it effectively works? Otherwise, it looks good to go

I couldn't find any yet.

lucadelu:
Can we merge  or not?

PR #1150, opened 2024-07-31
i.sentinel.download: use EODAG as a unified API for downloading sentinel products

HamedElgizery:
This PR will reimplement i.sentinel.download making use of the i.eodag module to use the EODAG API as the main source of Sentinel products.


src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
```diff
@@ -1240,89 +266,113 @@ def main():
...
+    try:
+        gs.find_program("i.eodag", "--help")
+    except ImportError:
+        gs.fatal(_("Addon i.eodag not found. Please intall it with g.extension."))
+
+    # Check if eodag config file is created
+    file_path = os.path.join(os.path.expanduser("~"), ".config/eodag/eodag.yml")
+    if not os.path.isfile(file_path):
+        dag = EODataAccessGateway()
+        gs.info(
```
lucadelu: @HamedElgizery are you sure this is needed? is the config file already exists this is superfluous

lucadelu:
It seems a big simplification, as soon it will be ready to test let us know.

src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
HamedElgizery: I think it is important because Copernicus Data Space Ecosystem, which has almost all the products, requires credentials for searching. So it can be an indication for the user to fill in the credentials.
However, I think having the settings file option back is the best workaround, but I need to implement it in i.eodag first...

src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
```diff
@@ -206,1025 +201,56 @@ def create_dir(directory):
...
-                sortby, ascending=[asc] * len(sortby), ignore_index=True
-            )
-        else:
-            self._products_df_sorted = scenes_df
-        gs.message(
-            _("{} Sentinel product(s) found").format(len(self._products_df_sorted))
-        )
+DATASOURCE_MAP = {
+    "ESA_CDSE": "cop_dataspace",
+    "GCS": "earth_search_gcs",
```
griembauer: In the case `"GCS": "earth_search_gcs"` I suppose google credentials are required to be configured via eodag? In the former version of `i.sentinel.download` this was not required because the GCS download URL was automatically composed by a given scene ID.
I think it is a great solution to use eodag as a one-stop-shop now, but if google credentials are required, this would also have to be indicated in the .html (or it is also possible to omit the Google cloud storage version entirely)

griembauer: Sorry for annoying so early - I noticed you only just began with this draft PR ;) I am looking forward to this solution!

HamedElgizery: Hello @griembauer :D

> (or it is also possible to omit the Google cloud storage version entirely)

I think that would be a reasonable choice since Google cloud storage only provides one Sentinel product in the EODAG context (S2_MSI_L1C) which is already provided by Copernicus.

HamedElgizery:
@neteler @ninsbl 
I was looking for these two products **S3OL1RAC & S3OL1SAC** on EODAG but they don't exist on Copernicus. 
I did find the first one, S3OL1RAC, on two other providers (WEKEO & SARA), but for the second product, S3OL1SAC, I don't seem to find it anywhere. Do you have any suggestion about them and if they are necessary to have?

src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
```diff
@@ -206,1025 +201,78 @@ def create_dir(directory):
...
+        return normalized_datetime.strftime("%Y-%m-%dT%H:%M:%S")
+    return normalized_datetime.astimezone(timezone.utc).strftime("%Y-%m-%dT%H:%M:%S")
+
+
+EODAG_PRODUCTTYPE_MAP = {
+    "S2MSI1C": "S2_MSI_L1C",
+    "S2MSI2A": "S2_MSI_L2A",
+    # Only found in wekeo, is S2MSI2Ap needed anymore?
+    # https://sentiwiki.copernicus.eu/web/s2-products#S2Products-XMLSchemaDefinitions(XSD)S2-Products-XML-Schema-Definitionstru
+    "S2MSI2Ap": "S2_MSI_L2AP",
```
veroandreo: This was a pilot L2 product IIRC, @neteler do you think we can safely remove?

neteler: Yes, IMHO no need to keep pilot products.

src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
```diff
@@ -1240,89 +288,145 @@ def main():
...
+            try:
+                acquisition_time = normalize_time(
+                    scene["properties"][headers_mapping[eodag_provider]["datetime"]]
+                )
+            except:
+                acquisition_time = scene["properties"][
+                    headers_mapping[eodag_provider]["datetime"]
+                ]
+            product_line += " " + acquisition_time
+            if headers_mapping[eodag_provider]["cloud_cover"] in scene["properties"]:
```
ninsbl: Instead of an if-in you can use the get method here...

src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
lucadelu: Thanks @griembauer, it is correct to add to documentation since it is an important change

src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
HamedElgizery: > @HamedElgizery are you sure this is needed? is the config file already exists this is superfluous

I found a way to have the settings option back, without needing to add anything to i.eodag.
I have put back the settings option, and removed the config file check in these two commits [d32374b](https://github.com/OSGeo/grass-addons/pull/1150/commits/d32374bdab064b750149090f650d00a77afbd762) [70f5c87](https://github.com/OSGeo/grass-addons/pull/1150/commits/70f5c87b52a4bc13c33fd63d998fa00ea3cc1c0a)
I am using the environment variables for that.

src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
```diff
@@ -97,7 +98,7 @@
 # % guisection: Filter
 # %end
 # %option
-# % key: uuid
+# % key: id
 # % type: string
 # % multiple: yes
 # % description: List of UUID to download
```
veroandreo: I guess this needs editing too if we are not using the uuid anymore?

src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
```diff
@@ -125,9 +126,9 @@
 # %option
 # % key: datasource
 # % description: Data-Hub to download scenes from.
-# % label: Default is ESA Copernicus Open Access Hub (ESA_COAH), but Sentinel-2 L1C data can also be acquired from USGS Earth Explorer (USGS_EE) or Google Cloud Storage (GCS)
-# % options: ESA_COAH,USGS_EE,GCS
-# % answer: ESA_COAH
+# % label: Default is ESA Copernicus Data Space Ecosystem (ESA_CDSE), but Sentinel-2 L1C data can also be acquired from Google Cloud Storage (GCS)
+# % options: ESA_CDSE,GCS
+# % answer: ESA_CDSE
```
veroandreo: GCS appears as deprecated below, shouldn't this be changed then?

src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
```diff
@@ -206,1004 +209,157 @@ def create_dir(directory):
...
+    "S3_SY_SYN": "S3SY2SYN",
+    "S3_SY_VGP": "S3SY2VGP",
+    "S3_SY_VG1": "S3SY2VG1",
+    "S3_SY_V10": "S3SY2V10",
+    "S3_SY_AOD": "S3SY2AOD",
+    "S3_RAC - SARA/WEKEO": "S3OL1RAC",
+    "S3OL1SAC": "DEPRECATED",  # Can not be found anywhere within EODAG
+}
+
+CLOUDCOVER_PRODUCTS = ["S2MSI1C", "S2MSI2A"]
```
veroandreo: S1 products do not have cloudcover, that's right because they are SAR data. But S3 is optic afaik, don't they have cloud cover property?? Just checking, haven't used S3.

src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
```diff
@@ -1221,108 +377,133 @@ def main():
...
-    elif options["datasource"] == "USGS_EE":
-        try:
-            import landsatxplore.api
-            from landsatxplore.errors import EarthExplorerError
-        except ImportError as e:
-            gs.fatal(_("Module requires landsatxplore library: {}").format(e))
+    try:
+        gs.find_program("i.eodag", "--help")
+    except ImportError:
+        gs.fatal(_("Addon i.eodag not found. Please intall it with g.extension."))
```
veroandreo: Shouldn't this be a single message? Users will need to have i.eodag installed which requires eodag.

veroandreo (commented):
I left a couple of comments. Would you mind updating the manual page too, so it's easier to test?

HamedElgizery:
> I left a couple of comments. Would you mind updating the manual page too, so it's easier to test?

Thanks @veroandreo, I will update the manual ASAP.

src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
```diff
@@ -206,1004 +209,157 @@ def create_dir(directory):
...
+    "S3SY2SYN": "S3_SY_SYN",
+    "S3SY2VGP": "S3_SY_VGP",
+    "S3SY2VG1": "S3_SY_VG1",
+    "S3SY2V10": "S3_SY_V10",
+    "S3SY2AOD": "S3_SY_AOD",
+    "S3OL1RAC": "S3_RAC - SARA/WEKEO",
+    "S3OL1SAC": "DEPRECATED",  # Can not be found anywhere within EODAG
+}
+
+REVERSE_PRODUCTTYPE_MAP = {
```
ninsbl: Instead of duplicating the dict with values as keys, you could use a generator to get the dict key for an eodag product like:

```
product_to_find = "S3_SY_SYN"
key = next(option_key for option_key, eodag_product in PRODUCTTYPE_MAP.items() if eodag_product == product_to_find)
```

ninsbl:
> Do you have any suggestion about them and if they are necessary to have?

S3 OL 1 RAC is an internal product. Perfectly fine to drop it... Not sure what the background of the SAC product has been. But I did not find documentation either. So fine to drop as well...

src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
```diff
@@ -206,1004 +209,157 @@ def create_dir(directory):
...
+    if normalized_datetime.tzinfo is None:
+        return normalized_datetime.strftime("%Y-%m-%dT%H:%M:%S")
+    return normalized_datetime.astimezone(timezone.utc).strftime("%Y-%m-%dT%H:%M:%S")
+
+
+PRODUCTTYPE_MAP = {
+    "S2MSI1C": "S2_MSI_L1C",
+    "S2MSI2A": "S2_MSI_L2A",
+    "OCN": "S1_SAR_OCN",
+    "GRD": "S1_SAR_GRD",
```
ninsbl, line 247: GRD data are now provided in SAFE format with internal COG, see: https://github.com/CS-SI/eodag/pull/1277

This illutrates how hardcoding product types limits `i.sentinel.download` compared to `i.eodag`.

You should add a note in the manual that i.eodag is used under the hood and offers more comprehensive download possibilities...

ninsbl (changes requested):
A general question is how backwards-compatibility is handled or if we should put a lot of effort into it?

Some features will disappear in this PR (e.g. possibility to download from DHUS portals = sentinelsat).

Personally, I would give priority to basic tests in i.eodag...

src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
HamedElgizery: After some testing, these are the Sentinel-3 products that have cloud coverage percentage:
* S3SY2SYN
* S3SY2VGP
* S3SY2VG1
* S3SY2V10
* S3SY2AOD

The rest do have the cloud coverage attribute but it is always 0.

src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
HamedElgizery: > GRD data are now provided in SAFE format with internal COG, see: [CS-SI/eodag#1277](https://github.com/CS-SI/eodag/pull/1277)

Cool, shall I add it or wait till it is officially released? 

> This illutrates how hardcoding product types limits `i.sentinel.download` compared to `i.eodag`.

Agree I think in most cases it is almost always better to use i.eodag directly, for more flexibility. Except for pre-existing code I guess.

> You should add a note in the manual that i.eodag is used under the hood and offers more comprehensive download possibilities...
 
Will do that.

HamedElgizery:
Thanks for the review, @ninsbl.

> Personally, I would give priority to basic tests in i.eodag...

I will be working on that in parallel with this PR now. 
This PR should be ready soon, as I think what is remaining for the most part is:
1. Updating the manual.
2. Emulate the -p flag to give a similar output as the previous implementation.

Is there any points that I might have not recognized?

src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.py
ninsbl: > > GRD data are now provided in SAFE format with internal COG, see: [CS-SI/eodag#1277](https://github.com/CS-SI/eodag/pull/1277)
> 
> Cool, shall I add it or wait till it is officially released?

there is a pre-release out. So I guess it is OK to add it...

ninsbl:
> I will be working on that in parallel with this PR now. This PR should be ready soon, as I think what is remaining for the most part is: (...)
> 2. Emulate the -p flag to give a similar output as the previous implementation.

For me it is no problem if the format of printed info changes a bit after switching to eodag... Not an issue I would invest a lot of time in. As mentioned, I think there will be breaking changes anyway...

src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.html
```diff
@@ -111,62 +111,46 @@ <h2>NOTES</h2>
...
+<ul>
+  <li>orbitNumber</li>
+  <li>timeliness (e.g. NT)</li>
+  <li>processingBaseline</li>
+  <li>storageStatus</li>
+</ul>
+
+Note that text based queryables have to be upper-case.
+Scenes with unavailbe queryable information are filtered out
+(e.g. if a product have the timeliness queryable set as <b>Null</b>,
```
ninsbl:
```suggestion
(e.g. if a product has the timeliness queryable set as <b>Null</b>,
```

src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.html
```diff
@@ -402,10 +386,10 @@ <h3>Download Sentinel products by scene name</h3>
 
 <h2>REQUIREMENTS</h2>
 
+<!-- Should i.eodag be added or is EODAG enough? -->
```
ninsbl: I would mention it here...

src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.html
```diff
@@ -111,62 +111,46 @@ <h2>NOTES</h2>
...
 
 <p>
 The <b>output</b> directory is created if not yet available. Default is current working directory.
 
-<p>
-The query string sent to ESA can be shown with <b>--verbose</b>.
-
 <!-- Not sure if Data Space Ecosystem doesn't have an archive, but apparently not -->
 <h3>Long Term Archive (DEPRECATED)</h3>
-Access to the product URL for data that are no longer available online in the Copernicus Data Space Ecosystem will automatically trigger
```
ninsbl: Data on CSDE is "Immediately available data (IAD)". Ao no archive. Not sure about other providers. You can remove the entire section I guess, but I would mention the providers used somewhere...

src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.html
```diff
@@ -111,62 +111,46 @@ <h2>NOTES</h2>
...
+Scenes with unavailbe queryable information are filtered out
+(e.g. if a product have the timeliness queryable set as <b>Null</b>,
+then it will be filtered out if you use query="timeliness=NT").
+
 
 <p>
-<!-- Leave a note about uuid deprecation -->
 <em>i.sentinel.download</em> also allows downloading of Sentinel products
-by specifying a (list of) ID. This operation is performed by the <b>id</b>
+by specifying a (list of) ID, where ID refers to the scene name on
```
ninsbl, line 142: File name?

src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.html
```diff
@@ -153,59 +111,46 @@ <h2>NOTES</h2>
...
+The <b>output</b> directory is created if not yet available. Default is current working directory.
 
+<!-- Not sure if Data Space Ecosystem doesn't have an archive, but apparently not -->
+<h3>Long Term Archive (DEPRECATED)</h3>
 To work with LTA you need to define the <b>sleep</b> parameter (and you can set the maximum number of retries with
 the <b>retry</b> parameter) to retry a download after the <b>sleep</b> value in minutes.
 
-<p>
-As an alternative to triggering re-upload Sentinel-2 data from the LTA, scenes can be downloaded
-from <b>USGS Earth Explorer</b> or <b>Google Cloud Storage</b> where all data remain online and directly accessible.
```
ninsbl, line 207: There are still a couple of references in the examples and in the python script...

src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.html
```diff
@@ -153,59 +111,46 @@ <h2>NOTES</h2>
...
-However, this applies only to Sentinel-2 data in levels 1C (USGS and GCS) and 2A (only GCS).
```
ninsbl, line 208: It would be great if you could structure examples with sub-captions, e.g. by Satellte (h3 for Sentinel-1, -2, and -3, printing providers, products queriables. ...).

src/imagery/i.sentinel/i.sentinel.download/i.sentinel.download.html
HamedElgizery: > but I would mention the providers used somewhere...

CDSE is the only provider used in the script currently, this is what you meant right?

HamedElgizery: > > but I would mention the providers used somewhere...
> 
> CDSE is the only provider used in the script currently, this is what you mean right?

veroandreo:
There are unresolved conversations here, a merging conflict and then we need an approval to merge. Can you please check the unresolved conversations and merge conflict @HamedElgizery ? After that I'd merge and do a second round of review once all is in

ninsbl:
As @veroandreo mentioned, backwards compatibility is not that a hard requirement for addons, But I think users should get a heads-up (GRASS-dev / GRASS-users ML) before breaking changes are submittet.

veroandreo:
> As @veroandreo mentioned, backwards compatibility is not that a hard requirement for addons, But I think users should get a heads-up (GRASS-dev / GRASS-users ML) before breaking changes are submittet.

I agree we can do a  courtesy heads up, but that's not common practice for addons really. COAH and sentinelsat stopped working almost one year ago, so users of i.sentinel should already know, I suppose. 

I am moreover for merging this, update the manual with a notice about COAH and sentinelsat not working anymore, and sending an email afterwards.

ninsbl:
> I am moreover for merging this, update the manual with a notice about COAH and sentinelsat not working anymore, and sending an email afterwards.

No objections from my side...

ninsbl (approved):

wenzeslaus:
Please wait for https://github.com/OSGeo/grass/pull/4205 to be merged, then restart the failing job.

PR #1152, opened 2024-08-01
i.eodag: additional options and modifications for i.sentinel.download

HamedElgizery:
- [x] Remove default clouds value.
- [x] Add -b flag to use AOI borders instead of bounding box.
- [x] Add -s flag to skip already downloaded products
- [x] Add footprint to the "sort" option.


src/imagery/i.eodag/i.eodag.py
```diff
@@ -254,59 +260,73 @@
...
-        "latmin": info["s"],
-        "lonmax": info["e"],
-        "latmax": info["n"],
-    }
+    args = {}
+    if vector:
+        args["vector"] = vector
+
+    # are we in LatLong location?
+    s = gs.read_command("g.proj", flags="j")
```
ninsbl: Here you could use `parse_command` right away...

ninsbl (changes requested):
Removal of cloud default looks good. As for the -b-flag, it does not really feel like an improvement to me. If needed at all, I probably would rather implement it in `i.sentinel.download`?

HamedElgizery:
> Removal of cloud default looks good. As for the -b-flag, it does not really feel like an improvement to me. If needed at all, I probably would rather implement it in `i.sentinel.download`?

I don't think it is possible to implement it in i.sentinel.download directly, since we can't pass the WKT to i.eodag, and the only interface method between them is the **map** option which is limiting. But we can remove the **b** flag from both i.eodag and i.sentinel.download, but I am afraid it might break some pre-existing code for users?

src/imagery/i.eodag/i.eodag.py
```diff
@@ -153,8 +164,8 @@
 # % key: sort
 # % type: string
 # % description: Field to sort values by
-# % options: ingestiondate,cloudcover
-# % answer: cloudcover,ingestiondate
+# % options: ingestiondate,cloudcover,footprint
+# % answer: cloudcover,ingestiondat,footprint
```
```suggestion
# % answer: cloudcover,ingestiondate,footprint
```

src/imagery/i.eodag/i.eodag.py
```diff
@@ -42,6 +42,18 @@
...
 
+# %flag
+# % key: b
+# % description: Use the borders of the AOI polygon and not the region of the AOI
+# % guisection: Filter
+# %end
+
+# %flag
+# % key: s
+# % description: Skip scenes that have already been downloaded after ingestiondate
```
```suggestion
# % description: Skip scenes that have already been downloaded
```

src/imagery/i.eodag/i.eodag.py
```diff
@@ -799,6 +829,47 @@ def products_compare(first, second):
...
+                os.path.getctime(existing_file[0])
+            )
+            ingestion_time = datetime.fromisoformat(
+                scene.properties["startTimeFromAscendingNode"]
+            ).replace(tzinfo=None)
+            if ingestion_time <= creation_time:
+                gs.message(
+                    _("Skipping scene: {} which is already downloaded.").format(
+                        scene.properties["title"]
+                    )
```
veroandreo: Why check for creation and ingestion time to decide to skip a file or not, isn't the matching file name enough? I do not understand the logic of comparing creation and ingestion time of existing files.

HamedElgizery: I wasn't sure why we need to compare creation and ingestion time as well, but I wrote in that way to maintain the logic of the previous implementation in i.sentinel.download.
i.sentinel.download had the same comparison done when skipping scenes, though I can't think of any scenario where that comparison would matter... Shall I remove it?

ninsbl: When data gets re-processed (e.g. with a new processing baseline) then users may have downloaded a product with the previous processing baseline, but the provider has a newer version of the same product. That has been the motivation behind the original check.

That said, I do not think _"startTimeFromAscendingNode"_ translates to what has been _"ingestiondate"_. Here properties like _"modificationDate"_ or _"publicationDate"_ are more appropriate I guess. Not sure if all providers support those...

When data was downloaded with _sentinelsat_ "in-progess" files got a suffix and then renamed when downloaded completely. So, `(Path(output) / scene.properties["title"]).exists()` would not find incomplete downloads... How does eodag handle filenames during download?

src/imagery/i.eodag/i.eodag.py
```diff
@@ -799,6 +829,47 @@ def products_compare(first, second):
...
+    :param search_results: EO products to be checked for existence in 'output' directory.
+    :type search_result: class'eodag.api.search_result.SearchResult'
+
+    :return: Sorted EO products
+    :rtype: class:'eodag.api.search_result.SearchResult'
+    """
+    filtered_result = []
+    # Check for previously downloaded scenes
+    existing_files = [os.path.join(output, f) for f in os.listdir(output)]
+    if len(existing_files) <= 1:
```
```suggestion
    if existing_files:
```
If a single file exists in _output_ it should be checked, no? And if you check for an empty list, don`t use _len()_

src/imagery/i.eodag/i.eodag.py
```diff
@@ -799,6 +829,47 @@ def products_compare(first, second):
...
+            sfile for sfile in existing_files if scene.properties["title"] in sfile
+        ]
+        if existing_file:
+            creation_time = datetime.utcfromtimestamp(
+                os.path.getctime(existing_file[0])
+            )
+            ingestion_time = datetime.fromisoformat(
+                scene.properties["startTimeFromAscendingNode"]
+            ).replace(tzinfo=None)
+            if ingestion_time <= creation_time:
```
```suggestion
            ingestion_time = scene.properties.get("modificationDate")
            if ingestion_time and datetime.fromisoformat(ingestion_time).replace(tzinfo=None) <= creation_time:
```

src/imagery/i.eodag/i.eodag.py
```diff
@@ -799,6 +829,47 @@ def products_compare(first, second):
...
+    # TODO: Implement a way to check if a file is completely downloaded
+    #       or if it was interrupted
+    for scene in search_result:
+        existing_file = [
+            sfile for sfile in existing_files if scene.properties["title"] in sfile
+        ]
+        if existing_file:
+            creation_time = datetime.utcfromtimestamp(
+                os.path.getctime(existing_file[0])
+            )
```
ninsbl: The following should be significantly faster:

```suggestion
    output = Path(output)
    for scene in search_result:
        scene_file = output / scene.properties["title"]
        if scene_file.exists():
            creation_time = datetime.utcfromtimestamp(
                os.path.getctime(scene_file)
            )
```
In that case you do not need the _os.path.join_ and the loop in line 846. You could remove 846 all together change 847 to:

```if next(os.scandir(output), None):```, so no _existing_files_ variable needed either...
That is probably the fastest way to check if _output_ folder is empty...

File: src/imagery/i.eodag/i.eodag.py [hunk 8]
```diff
@@ -799,6 +829,47 @@ def products_compare(first, second):
...
+            ).replace(tzinfo=None)
+            if ingestion_time <= creation_time:
+                gs.message(
+                    _("Skipping scene: {} which is already downloaded.").format(
+                        scene.properties["title"]
+                    )
+                )
+                continue
+        filtered_result.append(scene)
+    return SearchResult(filtered_result)
```
ninsbl: Could you do:
```suggestion
                search_result.remove(scene)
    return search_result
```
? That would avoid creating a new list, and I think _SearchResult_ is a _list_ object, so it should have a _.remove()_ method....

ninsbl (dismissed):
Some suggestions after looking into the existense check....

src/imagery/i.eodag/i.eodag.py
HamedElgizery: > When data gets re-processed (e.g. with a new processing baseline) then users may have downloaded a product with the previous processing baseline, but the provider has a newer version of the same product. That has been the motivation behind the original check.
> 
> That said, I do not think _"startTimeFromAscendingNode"_ translates to what has been _"ingestiondate"_. Here properties like _"modificationDate"_ or _"publicationDate"_ are more appropriate I guess. Not sure if all providers support those...

Ah I see that makes sense now, thanks for the explanation. 

> When data was downloaded with _sentinelsat_ "in-progess" files got a suffix and then renamed when downloaded completely. So, `(Path(output) / scene.properties["title"]).exists()` would not find incomplete downloads... How does eodag handle filenames during download?

When eodag downloads a file it just uses the scene's ID (aka. title) without any suffixes. So there isn't a direct way of checking if the product was completely downloaded or not. But there is another method... we could use the hidden directory `.download` which is automatically created by eodag when it downloads a scene.
Inside this directory, eodag creates a file for each product **_that is completely downloaded_**, and saves the source URL of that product. So we could use that to detect if a product was completely downloaded or if it was interrupted.

I will experiment with that more.

ninsbl: > we could use the hidden directory `.download` which is automatically created by eodag when it downloads a scene. Inside this directory, eodag creates a file for each product **_that is completely downloaded_**, and saves the source URL of that product. So we could use that to detect if a product was completely downloaded or if it was interrupted.

Nice! I did not know. Maybe worth to handle incomplete downloads in an "atexit" procedure?

HamedElgizery: > Maybe worth to handle incomplete downloads in an "atexit" procedure?

Sounds good, will give it a try.

src/imagery/i.eodag/i.eodag.py
HamedElgizery: Cool, that works yes.

src/imagery/i.eodag/i.eodag.py
HamedElgizery: @ninsbl I implemented the check for whether to skip a scene or not based on the existence of the corresponding file in `.downloaded`, and I have pushed these changes.

For the "atexit" procedure, there is a scenario that I think can be quite dangerous:
What if a user fully download a scene Q in directory X, and then they moved that scene manually to directory Y.
Later on, if they decide to run i.eodag with the `output` option pointing to directory Y, and the same scene Q manages to appear again in the search result... now if scene Q fails to download this time.. Both the fully downloaded scene Q and the new partially downloaded will be deleted...

I think multiple similar scenarios might accidentally delete files that shouldn't be deleted?

src/imagery/i.eodag/i.eodag.py
```diff
@@ -500,6 +526,7 @@ def list_products(products):
                 product_line += " "
             product_line += product_attribute_value
         print(product_line)
+        print(product.remote_location)
```
```suggestion
```

This (and the print-line above) were introduced for debuging, right? They should be removed before merging...

src/imagery/i.eodag/i.eodag.py
```diff
@@ -789,6 +816,10 @@ def products_compare(first, second):
                     continue
                 first_value = first.properties["cloudCover"]
                 second_value = second.properties["cloudCover"]
+            elif sort_key == "footprint":
```
ninsbl, line 818: You could simplify sorting like this:

```
products.sort(reverse=sort_order == "desc", key=lambda product: [product.properties.get(property) for property in sort_keys])
```

Not essential to merge this now. But something you can consider.

ninsbl (commented):
Nice.

Please remove the left-over debug messages and we are ready to merge IMHO.

As for terminology, the sort option name you added is "footprint", but you actually sort by "title". That is a bit confusing. For a later PR I suggested a simplified sort method that allows any property of an EOProduct to be used for sorting.

src/imagery/i.eodag/i.eodag.py
HamedElgizery: Yup my bad, only 529 is for debugging.

src/imagery/i.eodag/i.eodag.py
HamedElgizery: > You could simplify sorting like this:
> 
> ```
> products.sort(reverse=sort_order == "desc", key=lambda product: [product.properties.get(property) for property in sort_keys])
> ```

Cool thanks! I will use it in a future PR.

src/imagery/i.eodag/i.eodag.py
HamedElgizery: Shall we keep this since we are keeping the modification date condition?

veroandreo: the description was not right either, even if we keep the logic... so, no, I'd still remove the "after ingestiondate" part

veroandreo (approved):

veroandreo:
@ninsbl the changes required were made, that's why I dismissed the review, only to be able to move forward with this. Thanks!!

PR #1156, opened 2024-08-04
i.landsat.download: put back settings option for credentials

HamedElgizery:
This PR will return the `settings` option that can be used to set credentials either manually or using a text file.


src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
```diff
@@ -169,6 +176,119 @@ def normalize_time(datetime_str: str):
...
+        return planetary_computer_config
+
+    if datasource == "usgs":
+        if username is None or password is None:
+            gs.warnign(_("No user or password given for USGS"))
+        else:
+            if eodag_config is None:
+                # Shall we always just ignore any configuration in the eodag config file
+                # to eliminate confusion between i.eodag and i.landsat.download?
+                eodag_config = {"usgs": build_default_usgs_config(username, password)}
```
HamedElgizery, line 259: I think ignoring all of the eodag.yml file configurations will make it simpler by eliminating their effect on the module's behaviour, right?

src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
```diff
@@ -337,23 +470,10 @@ def main():
...
         import eodag
         from eodag import EODataAccessGateway
-    except:
-        gs.fatal(_("Cannot import eodag. Please intall the library first."))
+        from eodag.utils.exceptions import AuthenticationError
 
-    try:
         gs.find_program("i.eodag", "--help")
     except ImportError:
         gs.fatal(_("Addon i.eodag not found. Please intall it with g.extension."))
```
veroandreo:
```suggestion
        gs.fatal(_("Addon i.eodag not found. Please install it with g.extension."))
```

src/imagery/i.landsat/i.landsat.download/i.landsat.download.html
```diff
@@ -103,29 +95,86 @@ <h2>DESCRIPTION</h2>
 aborting (default is 300 seconds).
 
 <h2>NOTES</h2>
-USGS is still under development and can't be used for searching nor downloading.
+USGS is still under development and can't be used for searching
+nor downloading.
```
veroandreo: This sounds weird. It's the usage of the USGS provider which is under development in the eodag end, not USGS per se. Also, I thought only download by id was not possible...

src/imagery/i.landsat/i.landsat.download/i.landsat.download.html
```diff
@@ -103,29 +95,86 @@ <h2>DESCRIPTION</h2>
...
+
+<h3>Settings</h3>
+<p>
+<em>i.landsat.download</em> reads the user credentials either from the
+terminal or from the <b>settings</b> file. This file must contain two
+lines in case of USGS, or one line in case of Planetary Computer.
```
veroandreo:
```suggestion
lines in case of USGS provider, or one line in case of Planetary Computer provider.
```

src/imagery/i.landsat/i.landsat.download/i.landsat.download.html
```diff
@@ -103,29 +95,86 @@ <h2>DESCRIPTION</h2>
...
+Insert password:
+</pre></div>
+
+
+<h3>Planetary Computer</h3>
+<p>
+Most datasets are anonymously accessible, but a subscription
+key may be needed to
+<a href="https://planetarycomputer.microsoft.com/docs/concepts/sas/#rate-limits-and-access-restrictions">
+  increase rate limits and access private datasets.</a>
```
veroandreo:
```suggestion
<a href="https://planetarycomputer.microsoft.com/docs/concepts/sas/#rate-limits-and-access-restrictions">increase rate limits and access private datasets.</a>
```

src/imagery/i.landsat/i.landsat.download/i.landsat.download.html
```diff
@@ -103,29 +95,86 @@ <h2>DESCRIPTION</h2>
...
+Create an account <a href="https://planetarycomputer.microsoft.com/account/request">here</a>,
+then view your keys by signing in with your Microsoft account
+<a href="https://planetarycomputer.developer.azure-api.net/">here</a>.
```
veroandreo:
```suggestion
Users can create an account <a href="https://planetarycomputer.microsoft.com/account/request">here</a>,
and then view their keys by signing in with their Microsoft account
<a href="https://planetarycomputer.developer.azure-api.net/">here</a>.
```

veroandreo (commented):
I only had a quick look to the text and messages

src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
HamedElgizery: The way it is currently working is that:
The code reads the default eodag config file in '~/.config/eodag/eodag.yml', using the yaml library.
Then it edits the providers configurations and insert the credentials that the user have given either interactively or through a text file.
After that, a temp YAML file is created with the customized configuration, which is the same configuration the user have except that the credentials are inserted.
Finally, the code call i.eodag normally, and passes the custom temp YAML configuration file through the i.eodag `config` option.

What I am thinking, is that maybe it would be better if we ignore the default eodag configuration file, and just create a new empty one from scratch. Thus, any configurations set by the user will be invisible to i.landsat.download, and thus i.landsat.download would be isolated from i.eodag, and the user won't have to worry about their configurations in the default eodag.yml config file.

src/imagery/i.landsat/i.landsat.download/i.landsat.download.html
HamedElgizery: Yes right USGS is working fine and the only problem is with the IDs. I am not sure where did that comment come from I guess it was from the previous PR.

src/imagery/i.landsat/i.landsat.download/i.landsat.download.py
veroandreo: I do not like the idea of ignoring the eodag config file. 

Why not read the config file if available? Then, if credentials are passed via settings (file or interactive), maybe do what you suggest, i.e. create another config file for landsat? But will users end up with different eodag config files for different providers then? I'm thinking I search with usgs one day and another day with planetary, so I'll have my 2 text files and hidden there will be two config files one for usgs and one for planetary? or just one extra config file with both usgs and planetary? or will these be temporary? What do others think?

HamedElgizery: > Why not read the config file if available?

What I was thinking is that it could be simpler to not have users worry about the eodag config file... for example, they may have changed any of the parameters values, e.g. delete_url_params, or any other parameter when they were working with i.eodag, so it could be confusing if then use i.landsat.download and see that the module's behaviour changed.

> But will users end up with different eodag config files for different providers then? I'm thinking I search with usgs one day and another day with planetary, so I'll have my 2 text files and hidden there will be two config files one for usgs and one for planetary? or just one extra config file with both usgs and planetary? or will these be temporary? 

No there won't be any extra config files for each provider... they will be temporary files that are created by the module automatically and filled with the credentials. So users won't have to create these config YAML files manually, instead they will just use the settings text file normally, and then i.landsat.download will create temporary file with the credentials and pass that temporary file to i.eodag.

veroandreo: Ok, I see, in this way i.landsat.download will work as before (sort of)... Still users will need eodag and i.eodag installed. 

I would still keep the possibility to use the deafult eodag config file if it exists in the default location. Does that make sense? i.e., if I do not use the settings option, the default eodag config file should be used... is that too messy in terms of code? @ninsbl @lucadelu what do you think? Also, should the same logic be applied to i.sentinel.download then?

HamedElgizery: >  is that too messy in terms of code? 

No it is not messy it will be a small change, it is almost what the code currently does. Currently, it creates a temporary copy of the default config file, if it exists, with the credentials added to it.
If the default config file doesn't exist, it will create one as needed.

lucadelu: > Ok, I see, in this way i.landsat.download will work as before (sort of)... Still users will need eodag and i.eodag installed.
> 
> I would still keep the possibility to use the deafult eodag config file if it exists in the default location. Does that make sense? i.e., if I do not use the settings option, the default eodag config file should be used... is that too messy in terms of code? @ninsbl @lucadelu what do you think? Also, should the same logic be applied to i.sentinel.download then?

I also like the idea to use the general eodag config, I understood that in this way for user it is a simpler change to use the new implementation, however could you check if eodag config file exists otherwise use the settings option?

lucadelu (approved):

lucadelu:
@veroandreo @ninsbl I tried and with planetary computer and USGS and everything works as expected, I'm not able to merge the pull request. It seems I have no permission to do that.

veroandreo:
> @veroandreo @ninsbl I tried and with planetary computer and USGS and everything works as expected, I'm not able to merge the pull request. It seems I have no permission to do that.

Some months ago there was a re-structuring of github [teams](https://github.com/orgs/OSGeo/teams). Indeed I do not see you there. 

I can merge this one in any case. Have a question though: @HamedElgizery is the unresolved conversation above resolved? and how?

HamedElgizery:
> @HamedElgizery is the unresolved conversation above resolved? and how?

Yes it is resolved as following:
1. If `settings` option is set then get credentials from user and add it to the pre-existing eodag config file, thus all the set options will be maintained and recognized by EODAG. Otherwise, if the settings option is not used, then the module will attempt to use the credentials from the eodag config file.
2. If the EODAG config file haven't been created before, it will create a temporary one and use it normally as described in point 1.

Thus, supporting both the EODAG config file and the settings option, while giving priority to credentials from the `settings` option, as well as using any other configurations from the EODAG config file.

PR #1159, opened 2024-08-06
i.eodag: add dynamic setting for Creodias OTP 

HamedElgizery:
This PR will allow users to input their Creodias OTP on runtime, only if there is at least one scene that requires it.

### DIFF
--- a/src/imagery/i.eodag/i.eodag.html
+++ b/src/imagery/i.eodag/i.eodag.html
@@ -282,6 +282,19 @@ cop_dataspace:
           password: password
 </pre></div>
 
+<h3>Creodias</h3>
+Using the provider Creodias is slightly different, since it requires TOTP for authentication.
+To register to credoias create an account
+<a href="https://portal.creodias.eu/register.php">here</a>,
+then use your username, password in eodag configuration file.
+You will also need totp, a temporary 6-digits OTP
+(One Time Password, see Creodias documentation) to be able to authenticate
+and download scenes, which shall not be set through the eodag configuration file,
+but instead a prompt will show asking for OTP in case <em>i.eodag</em> attempts to
+download a scene from creodias. To discard these scenes the prompt will
+ask you to input <b>"-"</b> instead of the OTP. Note that interactive prompt
+does not work in the graphical user interface.
+
 <p>
 See <a href="https://eodag.readthedocs.io/en/stable/getting_started_guide/configure.html" target="_blank">Configure EODAG</a>
 section for more details about configuration of the
--- a/src/imagery/i.eodag/i.eodag.html
+++ b/src/imagery/i.eodag/i.eodag.html
@@ -282,18 +282,19 @@ cop_dataspace:
           password: password
 </pre></div>
 
-<h3>Creodias</h3>
-Using the provider Creodias is slightly different, since it requires TOTP for authentication.
-To register to credoias create an account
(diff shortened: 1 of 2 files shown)


ninsbl (approved):
This is OK from my perspective.
It should probably be addressed better at EODAG side.

An option would be to use a KeePassXC database and [pykeepass](https://pypi.org/project/pykeepass/) with an OTP python library to fetch tokens dynamically, without manual user interaction. But as mentioned, that should be handled by the creodias-plugin in EODAG (IMHO)...

src/imagery/i.eodag/i.eodag.py
```diff
@@ -1143,10 +1143,34 @@ def main():
...
+                    dag.providers_config["creodias"].auth.credentials[
+                        "totp"
+                    ] = creodias_otp
+                    dag._plugins_manager.get_auth_plugin("creodias").authenticate()
             custom_config = {
                 "timeout": int(options["timeout"]),
                 "wait": int(options["wait"]),
             }
+            if not search_result:
+                gs.message(_("Nothing to download.\nexiting..."))
```
```suggestion
                gs.message(_("Nothing to download.\nExiting..."))
```

veroandreo:
@HamedElgizery would you please document this in the manual page?

src/imagery/i.eodag/i.eodag.html
```diff
@@ -282,6 +282,19 @@ <h3>EODAG configuration</h3>
...
+To register to credoias create an account
+<a href="https://portal.creodias.eu/register.php">here</a>,
+then use your username, password in eodag configuration file.
+You will also need totp, a temporary 6-digits OTP
+(One Time Password, see Creodias documentation) to be able to authenticate
+and download scenes, which shall not be set through the eodag configuration file,
+but instead a prompt will show asking for OTP in case <em>i.eodag</em> attempts to
+download a scene from creodias. To discard these scenes the prompt will
+ask you to input <b>"-"</b> instead of the OTP. Note that interactive prompt
+does not work in the graphical user interface.
```
```suggestion
<h4>Creodias</h4>
To register to Creodias, users should create an account
<a href="https://portal.creodias.eu/register.php">here</a>, and
then use their username and password in the eodag configuration file.
Users will also need TOTP, a 6-digits temporary one time password, 
to be able to authenticate and download scenes (product search 
within creodias can be done without registering). 
This TOTP is only valid for very short time, i.e., 30 to 60 seconds, so it shall
not be set through the eodag configuration file.
When <em>i.eodag</em> attempts to download a scene from creodias,
users will be prompted to input the TOTP. If they prefer to discard these
scenes, they should enter <b>"-"</b> instead. Note that interactive 
prompt does not work in the graphical user interface.
```
I checked the text and try fixing it but I'm not sure I understand this correctly; users need two TOTP for creodias?? One in the config file and then another one for download?? Or they are the same thing?

HamedElgizery: No they are the same thing, what I am trying to convey is that users shouldn't set the totp in the config file, since it changes every 30 seconds or so anyways...
So they need to input it on the spot, when the prompt shows up, to be able to authenticate within the 30 seconds time frame.

veroandreo: updated now, check if that makes sense

HamedElgizery: > updated now, check if that makes sense

It is more clear now thanks!

lucadelu (approved):

PR #1165, opened 2024-08-08
i.eodag: update i.eodag to be compatible with EODAG 3

HamedElgizery:


ninsbl (approved):

veroandreo:
Download by id works now with usgs and cop_dataspace :partying_face: Let's fix the examples/description by informing users that provider and producttype must be specified as well when downloading by id. After that, this is good to go :) 

These examples could be used:
```
i.eodag id=S2A_MSIL2A_20240728T154941_N0511_R054_T17SPV_20240728T235651,S2A_MSIL2A_20240618T154941_N0510_R054_T17SPV_20240618T222157 provider=cop_dataspace producttype=S2_MSI_L2A

i.eodag id=LC08_L2SP_016035_20240715_20240722_02_T1 provider=usgs producttype=LANDSAT_C2L2
```

src/imagery/i.eodag/i.eodag.html
```diff
@@ -359,6 +360,19 @@ <h2>EXAMPLES</h2>
     pattern="LC09.*T1"
 <pre></div>
 
+Search and list Sentinel 2 scenes using IDs from Collection Data Space Ecosystem:
```
```suggestion
Search and list Sentinel 2 scenes using IDs from Copernicus Data Space Ecosystem. 
To download the selected scenes, simply remove the l flag.
```

src/imagery/i.eodag/i.eodag.html
```diff
@@ -359,6 +360,19 @@ <h2>EXAMPLES</h2>
...
+<div class="code"><pre>
+i.eodag -l provider=cop_dataspace producttype=S2_MSI_L2A \
+    id="S2A_MSIL2A_20240728T154941_N0511_R054_T17SPV_20240728T235651,
+    S2A_MSIL2A_20240618T154941_N0510_R054_T17SPV_20240618T222157"
+<pre></div>
+
+Search and list for a Landsat scene using its ID from UGSS:
```
```suggestion
Search and list for a Landsat scene using its ID from USGS. To download the 
selected scene, remove the l flag.
```

src/imagery/i.eodag/i.eodag.html
```diff
@@ -52,7 +52,8 @@ <h4>Searching using scenes IDs</h4>
 Users have a set of scenes IDs that they want to search for and download.
 They can either use the <b>id</b> option or use the <b>file</b> option
 and pass a text file, with one ID per line.
-In both cases, specifying a provider is optional. In case users do
+When searching with ID, user must specify the product type that the IDs belong to.
+Specifying a provider is optional. In case users do
```
```suggestion
When searching with ID, users must specify the product type that the ID 
belongs to. Specifying a provider is not mandatory, but it is recommended. In case users do
```

src/imagery/i.eodag/i.eodag.html
```diff
@@ -52,7 +52,8 @@ <h4>Searching using scenes IDs</h4>
...
 not specify the provider, products might be offered
 from different providers as long as the user provides the
 credentials in the configuration file.
```
```suggestion
credentials in the configuration file and the IDs are the same among providers.
```

veroandreo (approved):
Check if the changes make sense, and commit them if yes. I think it ts ready.

src/imagery/i.eodag/i.eodag.html
HamedElgizery: I updated this part here https://github.com/OSGeo/grass-addons/pull/1165/commits/dae18f367b6af74f6d5942c902e51d5f857e1775

I am trying to say that, if a provider is not specified then each scene could be offered using a different provider, but each scene will only be offered once, using one provider.

src/imagery/i.eodag/i.eodag.py
```diff
@@ -468,8 +477,14 @@ def no_fallback_search(search_parameters, provider):
...
-        server_poke = dag.search(**search_parameters, provider=provider)
-        if server_poke[1] == 0:
+        server_poke = dag.search(**search_parameters, provider=provider, count=True)
+        if int(eodag.__version__.split(".")[0]) < 3 and server_poke[1] == 0:
+            gs.verbose(_("No products found"))
+            return SearchResult([])
+        if (
+            int(eodag.__version__.split(".")[0]) >= 3
+            and server_poke.number_matched == 0
+        ):
```
ninsbl: This can be combined and shortened in one if condition:
```suggestion
        if (int(eodag.__version__.split(".")[0]) < 3 and server_poke[1] == 0) or
        (
            int(eodag.__version__.split(".")[0]) >= 3
            and server_poke.number_matched == 0
        ):
```

src/imagery/i.eodag/i.eodag.py
```diff
@@ -65,7 +65,7 @@
 # % key: producttype
 # % type: string
 # % description: Imagery product type to search for
-# % required: no
+# % required: yes
```
ninsbl: Not 100% sure what the right choice is here, as when you want to list available product types, you would not specify that option here...
You could leave it as not required, but use rules to allow or require certain combinations of options and flags: https://grass.osgeo.org/grass83/manuals/g.parser.html#conditional-parameters
Just something to consider...

ninsbl (commented):
Two comments to consider...

src/imagery/i.eodag/i.eodag.py
veroandreo: Agreed, the best option might be to set rules for producttype. I don't think it should be always required as otherwise listing of available products won't work...

HamedElgizery: I put it back to **no**. Couldn't add the rules though as it will cause unnecessary limitations, esp. with the the file option when the file is referring to a geojson and not a text file. Instead, I added a warning that the producttype is not set, when it is expected to be preferably set.

PR #1169, opened 2024-08-13
style: use overwrite and not 'o' flag in Python scripts

neteler:
The flag `o` has been renamed to `--overwrite` long time ago.

This PR changes to `overwrite=` syntax.

### DIFF
--- a/src/raster/r.basin/r.basin.py
+++ b/src/raster/r.basin/r.basin.py
@@ -388,7 +388,7 @@ def main():
             "r.stream.distance",
             stream_rast=r_outlet,
             direction=r_drainage_e,
-            flags="o",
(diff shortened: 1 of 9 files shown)


ninsbl:
Hmm,   in r.in.gdal or r.external -o means override projection check and seems independent from --o...

neteler:
> Hmm, in r.in.gdal or r.external -o means override projection check and seems independent from --o...

Of course, restored in 68a681ca0ef74e86b5fc2a7d7c3be797034898b1

neteler:
> Hmm, in r.in.gdal or r.external -o means override projection check and seems independent from --o...

r.unpack, too. 

I'll close this PR as mostly invalid.

PR #1170, opened 2024-08-13
i.eodag: handle cases when sorting parameters are not available 

HamedElgizery:


veroandreo:
It seems ok to me, but could you provide an example to test this?

src/imagery/i.eodag/i.eodag.py
```diff
@@ -813,15 +813,47 @@ def products_compare(first, second):
...
             if sort_key == "ingestiondate":
                 if "startTimeFromAscendingNode" not in first.properties:
+                    gs.warning(
+                        _("Could not sort by ingestion date, please report this issue")
+                    )
+                    continue
+                if "startTimeFromAscendingNode" not in second.properties:
+                    gs.warning(
+                        _("Could not sort by ingestion date, please report this issue")
+                    )
```
ninsbl: The if statements can be merged with an "or" (see suggestion below).

Maybe I am confusing things here, so please correct me if I am wrong, but I don`t think _startTimeFromAscendingNode_ translates to _ingestion time_ but rather _start os sensing time_ ...

src/imagery/i.eodag/i.eodag.py
```diff
@@ -813,15 +813,47 @@ def products_compare(first, second):
...
+                if "title" not in first.properties:
+                    gs.warning(
+                        _("Could not sort by footprint, please report this issue")
+                    )
+                    continue
+                if "title" not in second.properties:
+                    gs.warning(
+                        _("Could not sort by footprint, please report this issue")
+                    )
+                    continue
```
ninsbl: Here (as well as above) if conditions can be merged.

```suggestion
                if "title" not in first.properties or "title" not in second.properties:
                    gs.warning(
                        _("Could not sort by footprint, please report this issue")
                    )
                    continue
```

A question is: how or where to "report this issue"? Maybe better to have a similar warning message as for cloudcover or the like...

src/imagery/i.eodag/i.eodag.py
```diff
@@ -813,15 +813,47 @@ def products_compare(first, second):
...
-                first_value = first.properties["cloudCover"]
-                second_value = second.properties["cloudCover"]
+                    gs.warning(
+                        _(
+                            "Cloud cover not available for scene {}".format(
+                                first.properties.get("title", "NA")
+                            )
+                        )
+                    )
+                if "cloudCover" not in second.properties:
```
ninsbl: Merge with if above...

ninsbl (dismissed):
Please see my comments on these particular changes. This would also be an opportunity to consider also my arlier comment on sorting: https://github.com/OSGeo/grass-addons/pull/1152#discussion_r1705080040.

src/imagery/i.eodag/i.eodag.py
HamedElgizery: Yes, ingestiondate should be transalted to _creationTime_ instead. But is that what we are interested in, isn't the sensing time the one we care about?

veroandreo: To me, *creationTime* sounds like the time the data provider actually process the data. This was a question of mine some time ago, and Hamed came up with *startTimeFromAscendingNode*. We are indeed interested in when the data was sensed or ingested by the sensor. Can we please re-check that? @HamedElgizery do you have a link with definitions for these variables?

HamedElgizery: > Can we please re-check that? @HamedElgizery do you have a link with definitions for these variables?

I could not find a source for definitions in EODAG except for this table [1], where they also state:

> EODAG maps each provider’s specific metadata parameters to a common model using [OGC OpenSearch Extension for Earth Observation](http://docs.opengeospatial.org/is/13-026r9/13-026r9.html). Extra parameters having no equivalent in this model are mapped as is.

And from the OGC definitions table and manually checking a scene from Copernicus Data Space Ecosystem, _startTimeFromAscendingNode_ is indeed the time when the scene was sensed.


[1] https://eodag.readthedocs.io/en/stable/add_provider.html?highlight=metadata#opensearch-parameters-csv
[2] https://docs.ogc.org/is/13-026r9/13-026r9.html

HamedElgizery:
> This would also be an opportunity to consider also my arlier comment on sorting: [#1152 (comment)](https://github.com/OSGeo/grass-addons/pull/1152#discussion_r1705080040).

Thanks @ninsbl, I also came across this new sorting feature that will be released with EODAG 3 https://github.com/CS-SI/eodag/pull/943
I think we can use it instead of the current implementation, I will look more into how it can be used.

HamedElgizery:
@ninsbl I updated the implementation as in here: https://github.com/OSGeo/grass-addons/pull/1152#discussion_r1705080040

For the _ingestiondate_, shall we rename it to _sensingtime_? Yet, I believe that in i.sentinel.download _ingestiondate_ is used to refer to the time the scene was indexed in the database...  [1]

[1]https://scihub.copernicus.eu/twiki/do/view/SciHubUserGuide/FullTextSearch#:~:text=beginposition%3A%0Aendposition%3A-,ingestiondate,-%3A

src/imagery/i.eodag/i.eodag.py
```diff
@@ -825,33 +822,32 @@ def sort_result(search_result):
...
-
-    search_result.sort(key=cmp_to_key(products_compare))
+    sort_keys = list()
+    for sort_key in options["sort"].split(","):
+        if sort_key == "ingestiondate":
+            sort_keys.append(
+                (
+                    "startTimeFromAscendingNode",
+                    (
+                        "3000-12-31T00:00:00"
```
lucadelu: @HamedElgizery maybe we could use datetime.today() + datetime.timdelta(1) to non use a fixed data

src/imagery/i.eodag/i.eodag.py
ninsbl: Thanks, @HamedElgizery That is very valuable information and should probably be added to the manual.

src/imagery/i.eodag/i.eodag.py
ninsbl: Another approach without artificial keys could be like this: https://stackoverflow.com/a/18411610

veroandreo (approved):

veroandreo:
@ninsbl are you satisfied with @HamedElgizery changes? Would you mind approving them so we can merge? We are blocked at time.

PR #1179, opened 2024-08-18
r.in.vect: add a flag

ecodiv:
- corrections in the manual page
- add -g flag to set the region's extent to match the bounding box of the vector layer (for this flag, the -v flag is required).

### DIFF
--- a/src/raster/r.in.vect/r.in.vect.py
+++ b/src/raster/r.in.vect/r.in.vect.py
@@ -61,12 +61,22 @@
 # %end
 
 # %flag
+# % key: g
+# % label: Match region's extent to vector bounding box
+# % description: Set region extent to match that of the bounding box of the vector layer.
+# %end
+
+# %flag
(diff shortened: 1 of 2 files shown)


src/raster/r.in.vect/r.in.vect.py
```diff
@@ -60,6 +60,12 @@
 # % description: Set this flag if the whole vector layer needs to be converted. By default, only the part overlapping with the computational region is converted.
 # %end
 
+# %flag
+# % key: g
+# % label: Match region's extent to vector bounding box
+# % description: Set region extent to match that of the bounding box of the vector layer.
```
neteler: What about using instead `-a` (as done in `g.region`) to "adjust" things? Since `-g` is commonly used for script-parsable output...

ecodiv: Good point, changed.

PR #1190, opened 2024-08-27
v.in.gbif: bugfixes

ecodiv:
- Change how CSV are opened (text mode instead of binary mode).
- Import the gbifid field as Integer64 to avoid error messages.


hellik (approved):
No chance to test right now, but it looks ok.

PR #1192, opened 2024-08-27
i.sentinel.coverage: update module to be compatible with the i.sentinel.download EODAG migration 

HamedElgizery:
As of the i.sentinel.download migration to the EODAG API, multiple changes are needed to make i.sentinel.coverage compatible. The main problem was because the key referring to the scenes' files names  are changed from _identifier_ to _title_

This PR will also fix the example of using the **name** option, by omitting the **clouds** option, and by removing the exclusivity relation between the **names** and the **producttype** options.


src/imagery/i.sentinel/i.sentinel.coverage/i.sentinel.coverage.py
```diff
@@ -310,30 +308,31 @@ def main():
...
+                column="a_title,title",
                 quiet=True,
             )
             columns_dict = grass.parse_command("v.info", map=temp_overlay, flags="c")
             drop_columns = [
                 col.split("|")[1]
                 for col in columns_dict
-                if col.split("|")[1] not in ["cat", "identifier"]
+                if col.split("|")[1]
+                not in ["cat", "title"]  # What does cat refer to here?
```
veroandreo: cat is the geometry identifier in GRASS vector maps, so, like the id in other formats

HamedElgizery:
Hello @veroandreo, sorry for the late follow up here.
I have updated the manual page and the i.sentinel.coverage and it works fine as far as I have tested.

There have been a problem with EODAG < 3.0.0... The results of searching wasn't consistent, as if you were to do the same searching query twice, the resulting scenes could vary, and that was causing troubles with i.sentinel.coverage. But, it is fixed now with EODAG 3, and the module works fine now.

I will add a check to make sure that the version of EODAG installed is > 3.0.0 here. (and within i.sentinel.download in another PR)

PR #1206, opened 2024-09-19
CI: add liblapacke-dev as dependency

nilason:
Add `liblapacke-dev` as dependency, to adopt to changes in GRASS GIS main.

### DIFF
--- a/.github/workflows/apt.txt
+++ b/.github/workflows/apt.txt
@@ -8,7 +8,7 @@ libgdal-dev
 libgl1-mesa-dev
 libglu1-mesa-dev
 libnetcdf-dev
-libopenblas-dev
+liblapacke-dev
 libpng-dev
 libproj-dev
 libreadline-dev


nilason:
Updates are needed for the cron jobs too.

nilason:
> Updates are needed for the cron jobs too.

It not clear to me where this is run, and where its dependencies are declared. But `--with-blas-includes=/usr/include/atlas/` and `--with-lapack-includes=/usr/include/atlas/` are no longer valid configure flags.

wenzeslaus:
Some, or hopefully all, of the code is in grass-addons. @neteler can comment more.

wenzeslaus (approved):

neteler:
> Some, or hopefully all, of the code is in grass-addons. @neteler can comment more.

Installation of dependencies on grass.osgeo.org (DONE):

`neteler@grasslxd:~$ sudo apt install $(cat $HOME/src/main/.github/workflows/apt.txt)`

Flags need to be adjusted in the relevant `cron*.sh` scripts (TODO via PR):

https://github.com/OSGeo/grass-addons/tree/grass8/utils/cronjobs_osgeo_lxd/

PR #1225, opened 2024-10-16
i.eodag: Support AOIs from other mapsets (fix existence check)

ninsbl:
This PR implements the existence check for AOI vector maps according to the [style guide](https://github.com/OSGeo/grass/blob/main/doc/development/style_guide.md#checking-inputs-of-a-tool)

The previous implementation fails for find AOI vector maps from other mapsets or if mapset qualifiers are used in general.


ninsbl:
Taking the liberty to merge without review as the change is tiny and towards the style guide...

PR #1230, opened 2024-10-24
r.maxent.train: bugfix and clarification help text

ecodiv:
Fixes a bug and some small rewrites of help text

### DIFF
--- a/src/raster/r.maxent.train/r.maxent.train.html
+++ b/src/raster/r.maxent.train/r.maxent.train.html
@@ -135,10 +135,14 @@ Use the output of <em>v.maxent.swd</em> as input for
 The outputs will be written to this folder.
 
 <p>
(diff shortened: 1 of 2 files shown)


neteler:
There is an unexpected `gs` file included (38 MB, apparently PDF). Accidentally committed?

ecodiv:
Yes, not sure how that happened, definitely unintentional. Thanks for catching that, I'll remove it and check when back behind the computer.

ecodiv:
@neteler Removed the gs file. Now one of the test fails. From the error message, I can't really see what is wrong, and whether there is a problem with my code. Do you have an idea?

neteler:
The CI says " 7 successful checks" - which error do you mean?

ecodiv:
> The CI says " 7 successful checks" - which error do you mean?

Sorry for the noise, there was an error, but running the checks again solved it, so apparently not related to my code/text.

PR #1237, opened 2024-11-11
r.vif: bugfix

ecodiv:


ecodiv:
This was already committed, not sure what is going wrong here, but closing

PR #1245, opened 2024-11-18
i.eodag: remove experimental tool warning

veroandreo:
I suggest removing the warning since the GSoC was completed successfully some months ago, EODAG released 3+ version that solved several issues we dealt with during the development of the tool and, last but not least, the warning clutters every text output in the terminal 

@HamedElgizery what do you say?

### DIFF
--- a/src/imagery/i.eodag/i.eodag.py
+++ b/src/imagery/i.eodag/i.eodag.py
@@ -1323,13 +1323,6 @@ def main():
 
 
 if __name__ == "__main__":
-    gs.warning(_("Experimental Version..."))
-    gs.warning(
-        _(
-            "This module is still under development, \
-            and its behaviour is not guaranteed to be reliable"
-        )
-    )
     options, flags = gs.parser()
 
     try:


ninsbl (approved):

HamedElgizery:
> I suggest removing the warning since the GSoC was completed successfully some months ago, EODAG released 3+ version that solved several issues we dealt with during the development of the tool and, last but not least, the warning clutters every text output in the terminal
 
Yes, I agree. EODAG 3 has solved most, if not all, of the issues that we faced.
I also think IMHO, that it would be helpful if we restrict the module to only use EODAG 3+, and throw a fatal error if the version is below 3.

```py
if int(eodag.__version__.split(".")[0]) < 3:
    gs.fatal(
        _(
            "eodag 3.0.0 or higher is required. eodag {} was found.".format(
                eodag.__version__
            )
        )
    )
```

src/imagery/i.eodag/i.eodag.py
```diff
@@ -1323,13 +1323,6 @@ def main():
 
 
 if __name__ == "__main__":
```
HamedElgizery, line 1325: Maybe we can add this after the `try except` block to force the usage of EODAG 3:

```py
if int(eodag.__version__.split(".")[0]) < 3:
    gs.fatal(
        _(
            "eodag 3.0.0 or higher is required. eodag {} was found.".format(
                eodag.__version__
            )
        )
    )
```
Not sure if there is a different format for these type of error messages?

veroandreo: @ninsbl what do you think?

veroandreo: as of now, and because I teach tomorrow, I'll merge the removal of the warning only

PR #1248, opened 2024-11-22
v.in.gbif: bugfix - add fallback if encoding text file does not work

ecodiv:
On Windows (and perhaps on other OS), in some cases, issues arise with the encoding of a GBIF file. With this patch, an attempt is made to read the file using the default encoding. If that fails, it is read in using utf-8. I considered the option to have the script guess the encoding, but that requires extra dependencies, so I opted for the simpler solution to use utf-8 and if there are still problematic characters, to substitute them. A warning is issued in these cases to inform the user.

### DIFF
--- a/src/vector/v.in.gbif/v.in.gbif.py
+++ b/src/vector/v.in.gbif/v.in.gbif.py
@@ -56,7 +56,83 @@ import csv
 import math
 import shutil
 import tempfile
-import grass.script as grass
+import grass.script as gs
+
+import csv
+
+
+def process_csv(
+    gbifraw, new_gbif_csv, primary_encoding=None, fallback_encoding="utf-8"
+):
+    """
+    Process a CSV file using a primary encoding and fallback to another if the first fails.
+    Substitutes problematic characters during the second fallback.
+
+    :param gbifraw: Path to the input CSV file.
+    :param new_gbif_csv: Path to the output CSV file.
+    :param primary_encoding: Primary encoding to try first. If None, uses system default.
+    :param fallback_encoding: Fallback encoding to use if the primary fails.
+    """
+
+    def process_file(input_encoding, handle_errors=False):
+        """
+        Process the file with the specified encoding.
+        If `handle_errors` is True, substitutes problematic characters.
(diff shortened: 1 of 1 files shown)


hellik (commented):
I can't test it with a real life example at the moment. Do you have an example where such fallback is needed?

Though,, it looks good.

ecodiv:
> I can't test it with a real life example at the moment. Do you have an example where such fallback is needed?
> 

@hellik Yes, I had several students reporting this problem. [This is one example](https://github.com/user-attachments/files/17879163/0023319-241107131044228.csv) of a GBIF CSV file that one of my students could not import, with the following error: 

```
File "C:\Users\AAA\AppData\Roaming\GRASS8\addons/scripts\v
.in.gbif.py", line 93, in main
    for row in gbifreader:
  File "C:\Program Files\GRASS GIS
8.4\Python312\Lib\encodings\cp1252.py", line 23, in decode
    return
codecs.charmap_decode(input,self.errors,decoding_table)[0]
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
UnicodeDecodeError: 'charmap' codec can't decode byte 0x81
in position 5215: character maps to <undefined>
(Wed Nov 20 12:06:29 2024) Command ended with non-zero return code 1 (1 sec) 
```

Which implies, if I understand right, that Python on their Windows defaults to legacy window encodings (cp1252). With the patch, they were able to import the same file. 

Note that on my Window computer, I had no issue importing the same file. One difference is that I installed Windows using OSGeo4W and the student installed it using the stand-alone installer. But, I can't be sure that is what makes the difference because I had other students who all installed GRASS using the stand-alone installer, and who did not have problems importing  (other) GBIF CSV files. 

Are you OK with merging this, or you would like to do some further testing?

hellik:
Thanks for the insights. Unfortunately there are so many different ways how Windows boxes behave.

Lets go ahead to merge. :-)

hellik (approved):
Ready to merge.

PR #1255, opened 2024-12-06
v.maxent.swd: Fix overwrite behavior for existing species SWD file

ecodiv:
When the species swd file already existed, and the option “overwrite” was selected, new species data was appended to the existing species swd file. With this fix, when there is already a species swd file with the user-defined name, and the option “overwrite” is selected, the existing species swd file is removed before a new one is created.


neteler:
(would you mind to add a line what the bugfix is about? Ideally also in the PR title)

PR #1276, opened 2025-01-07
CI: silence confusing --help output in workflows/test.sh

neteler:
This PR silences some confusing help text output during data installation by `g.download.location`:

For the error, see
https://github.com/neteler/grass-addons/actions/runs/12647834206/job/35240965201

Additionally: comment added.

### DIFF
--- a/.github/workflows/test.sh
+++ b/.github/workflows/test.sh
@@ -4,14 +4,16 @@
 set -e
 
 # Install module from addons if not available (in v7).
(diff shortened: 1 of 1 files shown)


neteler:
TBH, maybe only the silencing part is needed...?

neteler:
Re: broken PDAL test in the log: it was disabled in #871.

neteler:
> Re: broken PDAL test in the log: it was disabled in #871.

... which only seems to be triggered during backports.

PR #1280, opened 2025-01-14
r.stream.snap: Fix snapping to immediate NW neighbor cell

HuidaeCho:
This PR fixes snapping to the immediate NW neighbor cell. It was using `di=-1` and `dj=-1` to indicate the unsnapped status, but they are valid deltas for the NW cell. This PR uses the existing `status` variable to store this status.


metzm (approved):
Looks good to me.

PR #1283, opened 2025-01-20
r.slopeunits.optimize: clean up tmp mapsets

metzm:
r.slopeunits.optimize: On error or Ctrl-C, switch back to the original mapset and remove temporary mapsets


neteler (approved):
Looks good.

PR #1284, opened 2025-01-21
r.slopeunits: fix import of grass.scripts

neteler:
Fixes

```
  File "/home/neteler/.grass8/addons/r.slopeunits/scripts/r.slopeunits.clean", line 82, in <module>
    import gs.script as gs
ModuleNotFoundError: No module named 'gs'
```

(and related)

Discovered in https://grass.osgeo.org/addons/grass8/logs/r.slopeunits.log


metzm (approved):
Thanks for the fix!

PR #1330, opened 2025-02-19
cronjobs: Add Markdown and generalize addon build

wenzeslaus:
- Add Markdown path to compilation.
- Make running of the URL check script optional, so the build does not fail if the script is not available (local server path is currently hardcoded, so that's tricky or maybe even impossible to change for me).
- Fix the misleading typo in the usage info.

PR #1331, opened 2025-02-19
cronjonb: simplify hugo cronjob by using rsync

neteler:
This PR simplifies the hugo cronjob by using rsync:

- avoids need of tmp-dir
- reduces rm/cp operations to single rsync call
- include header

In tests on grass.osgeo.org this modified script needs approx 80% of the time the original script required.

### DIFF
--- a/utils/cronjobs_osgeo_lxd/hugo_clean_and_update_job.sh
+++ b/utils/cronjobs_osgeo_lxd/hugo_clean_and_update_job.sh
@@ -23,5 +23,4 @@ cd /home/neteler/grass-website/ && \
    nice /home/neteler/go/bin/hugo && \
    rsync -a --delete /home/neteler/grass-website/public/ /var/www/html/ && \
    ln -s /var/www/code_and_data/* /var/www/html/ && \
-   (cd /var/www/html/ ; /home/neteler/bin/fix_link_timestamp.sh .) && \
-   exit 0
+   (cd /var/www/html/ ; /home/neteler/bin/fix_link_timestamp.sh .)
--- a/utils/cronjobs_osgeo_lxd/hugo_clean_and_update_job.sh
+++ b/utils/cronjobs_osgeo_lxd/hugo_clean_and_update_job.sh
@@ -26,10 +26,29 @@
 #  6. restore timestamps of links from their original time stamps in src directory
 ####
 
+# function to update timestamp of link to the source timestamp
+fix_link_timestamp()
+{
+ if [ -z "$1" ] ; then
+   echo 'ERROR: Parameter missing. Specify the folder (. for current)!'
+   exit
+ fi
+
+ for mylink in $(find . -type l) ; do
+  LINK="$(namei ${mylink} | grep '^ l ' | tr -s ' ' ' ' | cut -d' ' -f3)"
+  ORIG="$(namei ${mylink} | grep '^ l ' | tr -s ' ' ' ' | cut -d' ' -f5-)"
+
+  echo "Updating timestamp of link <$ORIG> ---> <$LINK> timestamp"
+
+  # transfer timestamp
+  touch -h -m -r "$ORIG" "$LINK"
+ done
+}
+
 cd /home/neteler/grass-website/ && \
    git pull origin master && \
    rm -rf /home/neteler/grass-website/public/* && \
    nice /home/neteler/go/bin/hugo && \
    rsync -a --delete /home/neteler/grass-website/public/ /var/www/html/ && \
    ln -s /var/www/code_and_data/* /var/www/html/ && \
-   (cd /var/www/html/ && /home/neteler/bin/fix_link_timestamp.sh .)
+   (cd /var/www/html/ && fix_link_timestamp .)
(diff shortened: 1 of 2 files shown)


utils/cronjobs_osgeo_lxd/hugo_clean_and_update_job.sh
```diff
@@ -3,31 +3,25 @@
...
-   mkdir /var/www/html_new && \
-   \cp -rp /home/neteler/grass-website/public/* /var/www/html_new/ && \
-   rm -fr /var/www/html/* && \
-   \mv /var/www/html_new/* /var/www/html/ && \
+   rsync -a --delete /home/neteler/grass-website/public/ /var/www/html/ && \
    ln -s /var/www/code_and_data/* /var/www/html/ && \
-   rmdir /var/www/html_new && \
-   (cd /var/www/html/ ; /home/neteler/bin/fix_link_timestamp.sh .)
+   (cd /var/www/html/ ; /home/neteler/bin/fix_link_timestamp.sh .) && \
+   exit 0
```
wenzeslaus: Any special reason for adding `exit 0`?

utils/cronjobs_osgeo_lxd/hugo_clean_and_update_job.sh
```diff
@@ -3,31 +3,25 @@
...
    nice /home/neteler/go/bin/hugo && \
-   mkdir /var/www/html_new && \
-   \cp -rp /home/neteler/grass-website/public/* /var/www/html_new/ && \
-   rm -fr /var/www/html/* && \
-   \mv /var/www/html_new/* /var/www/html/ && \
+   rsync -a --delete /home/neteler/grass-website/public/ /var/www/html/ && \
    ln -s /var/www/code_and_data/* /var/www/html/ && \
-   rmdir /var/www/html_new && \
-   (cd /var/www/html/ ; /home/neteler/bin/fix_link_timestamp.sh .)
+   (cd /var/www/html/ ; /home/neteler/bin/fix_link_timestamp.sh .) && \
```
wenzeslaus: Not directly related to this PR, but is the timestamp fixing needed? I don't see the script in this repo. Maybe you have a comment there?

wenzeslaus (commented):
The rsync is indeed much more readable and cutting some time is nice.

utils/cronjobs_osgeo_lxd/hugo_clean_and_update_job.sh
neteler: It indicates that the entire `utils/cronjobs_osgeo_lxd/hugo_clean_and_update_job.sh` was successfully completed and not the last executed script within (here: `fix_link_timestamp.sh`).

utils/cronjobs_osgeo_lxd/hugo_clean_and_update_job.sh
neteler: It has been my (personal) policy to not damage timestamps when copying over files. I tried to carefully keep the server timestamps since the late 1990s :-) Random example: https://grass.osgeo.org/grass42/source/

The [fix_link_timestamp.sh](https://github.com/neteler/script_collection/blob/master/filename_tricks/fix_link_timestamp.sh) script I can add to this PR if desired.

utils/cronjobs_osgeo_lxd/hugo_clean_and_update_job.sh
wenzeslaus: I don't see that. With `&&`, if cd and fix_link_timestamp.sh return non-zero, that's what will be returned from the script. If they return zero, zero will be returned from the script anyway. Or not? Or is this some Bash best practice? We don't end this with `|| exit 1`.

I'm just trying to avoid wondering what this is later on. We can merge whatever.

utils/cronjobs_osgeo_lxd/hugo_clean_and_update_job.sh
wenzeslaus: Lovely. Can you put there a comment that his makes the time stamps of the symlinks the same as time stamps of the original files? (If that's what this is doing.)

wenzeslaus:
Maybe we can merge it around the same time as #1330 and you can watch if the server explodes due to either of those PRs.

utils/cronjobs_osgeo_lxd/hugo_clean_and_update_job.sh
neteler: Okay, agreed. Reverted in 768661da66b9d9e2381c6a0885285aa2cc63b013

utils/cronjobs_osgeo_lxd/hugo_clean_and_update_job.sh
neteler: Time stamp restore function included in 9963b614d7fcb1a0dfad68cb1baca07a1941d5ec

neteler:
I have already tested the latest PR version on grass.osgeo.org. Appears to successfully work.

wenzeslaus (approved):
Perfect!

neteler:
The updated script is now in place on the server. 
Let's watch if anything goes wrong on https://grass.osgeo.org/ in the next 24 hs...

wenzeslaus:
Thanks!

PR #1332, opened 2025-02-20
doc: Convert HTML to Markdown

wenzeslaus:
This converts HTML markup in .md files to Markdown using the ./utils/grass_html2md.sh script from core.

In addition to the automatic conversion it:

- Does basic whitespace cleanup (esp. double whitespace in lists).
- Removes empty first line for v.area.stats.
- Uses text code blocks where appropriate for i.sentinel.preproc.
- Specifies text for r.fuzzy.system code blocks with syntax and explanation examples.

It further adjusts linting to accommodate the auto-converted Markdown.

It does not touch r.landscape.evol because that tool has already had Markdown documentation (which needs to be revised manually as needed).


wenzeslaus:
Running `pre-commit run markdownlint-fix -a` locally, I get suggestions, but I don't know how to actually fix the code or disable the checks.

The suggestion is to remove `$ `, but the code is correct, it is the syntax of a language specific to that tool. I though using `text` as language will make it work, but I still get the wrong suggestion (and thus failure in CI).

```diff
--- a/src/raster/r.fuzzy.system/r.fuzzy.system.md
+++ b/src/raster/r.fuzzy.system/r.fuzzy.system.md
@@ -14,7 +14,7 @@
     of rule file. Lines beginning with **\#** are comments.
     
     ```text
-    $ set_name {side; points; boundary_shape; hedge; height }
+    set_name {side; points; boundary_shape; hedge; height }
     ```
     
 
@@ -44,7 +44,7 @@
     An example of fuzzy sets definition:
     
     ```text
-    $ moderate {both; 90,100,120,130; sshaped; 0; 1}
+    moderate {both; 90,100,120,130; sshaped; 0; 1}
     ```
     
 
@@ -79,7 +79,7 @@
     An example of fuzzy rules definition:
     
     ```text
-    $ small {distance = high & elev = high}
+    small {distance = high & elev = high}
     ```
```

wenzeslaus:
Fortunately, General linting / GitHub Super Linter workflow clarified for me what are the issues, so I'm ignoring them for now, but unfortunately markdownlint triggers the warning for language set to text. Maybe, we need to use a custom language. (?)

cwhite911:
> unfortunately markdownlint triggers the warning for language set to text. Maybe, we need to use a custom language.

@wenzeslaus Which Rule was trigger this warning?

cwhite911:
> > unfortunately markdownlint triggers the warning for language set to text. Maybe, we need to use a custom language.
> 
> @wenzeslaus Which Rule was trigger this warning?

Never mind I see it.

wenzeslaus:
> Maybe, we need to use a custom language. (?)

I mean later, in some following PR.

cwhite911 (approved):
From the files I've spot check this looks good to me. The only issues I've found were existing bugs in the origin html, so unrelated to this PR.

petrasovaa (approved):

PR #1348, opened 2025-02-26
docs: fix HTML errors in r.hand.html

neteler:
At time `r.hand` cannot be compiled ([log](https://grass.osgeo.org/addons/grass8/logs/r.hand.log)) against GRASS GIS 8.4 (which is the version to build the addon HTML pages).

This PR fixes some HTML tags (as understood by the build system) in the manual page.


neteler:
An observation: there is no MD file yet in https://github.com/OSGeo/grass-addons/tree/grass8/src/raster/r.hand/

cwhite911 (approved):

Issue #1414, opened 2025-06-03
[Bug] r.hydro.flatten incaccurate  water elevations next to shore during DEM creation


dnewcomb:
**Name of the addon**
.r.hydro.flatten
**Describe the bug**
When  using the Raster map representing filled digital elevation model option, the water elevations within 2 cells of the shoreline are higher near the shoreline and do not match the elevation of the water  in the standalone water elevation product. 

**To Reproduce**

1. Install the r.hydro.flatten addon
2. Start r.hydro.flatten and fill in the Raster map representing filled digital elevation model with an output raster name
3. Open both the water elevation raster and the filled digital elevation model raster  in the map display. 
4. Uncheck the filled digital elevation model raster in the layers panel
5. Select both the water elevation and filled digital elevation model raster
6. Select the Query Raster tool for the top toolbar.
7. Select pixels along the edge of the water elevation raster in the water.
8. The water values near the edge of the water will not match the values for the  filled digital elevation model raster
9. The values for the water elevation raster and the  filled digital elevation model raster should be the same.

**System description (please complete the following information):**

- Operating System: Linux Mint 
- GRASS GIS version :  8.5.0 dev
GRASS nc6543/PERMANENT:~ > g.version -rge
version=8.5.0dev
date=2025
revision=bd88d25baf
build_date=2025-05-29
build_platform=x86_64-pc-linux-gnu
build_off_t_size=8
libgis_revision=1df16fe61e
libgis_date=2025-05-19T15:56:07+00:00
proj=9.4.0
gdal=3.8.4
geos=3.12.1
sqlite=3.45.1

**Additional context**
The final step for creating the  filled digital elevation model raster is a call to r.patch using the r.fill.stats created layer and the water elevation layer

    if options["filled_elevation"]:
        gs.run_command(
            "r.patch",
            input=[tmp_rfillstats, options["water_elevation"]],
            output=options["filled_elevation"],
        )
        gs.run_command("r.colors", map=options["filled_elevation"], raster=ground)
        gs.raster_history(options["filled_elevation"])

Flipping the order of the input raster layers to r.patch appears to fix the issue.
 
if options["filled_elevation"]:
        gs.run_command(
            "r.patch",
            input=options["water_elevation"],[tmp_rfillstats ],
            output=options["filled_elevation"],
        )
        gs.run_command("r.colors", map=options["filled_elevation"], raster=ground)
        gs.raster_history(options["filled_elevation"])

[...]

PR #1415, opened 2025-06-03
Update r.hydro.flatten.py to change order of r.patch to create the optional filled DEM raster layer

dnewcomb:
This changes the order of the input rater layers to r.patch . Previously the values from the r.fill.stats result were overwriting the calculated water level near the edges of the water.


dnewcomb:
This should fix https://github.com/OSGeo/grass-addons/issues/1414

petrasovaa (approved):
Didn't verify, but I think it makes sense!

petrasovaa (approved):

PR #1424, opened 2025-06-18
add section to create null for high std dev then fill with rfillnulls

dnewcomb:
There are many situations were void areas with high standard deviation edge values should not be modeled as flat in a filled elevation raster. This modification adds an input option , stddev_limit.  Any void areas with edge standard deviation values above this limit are removed in an initial filled elevation raster using  r.mapcalc . The resulting raster is passed to r.fillnulls with the default values with the exception of tension,which is hard coded to 150 and memory, which is hard coded to 4000 MB  .

It may be that a memory input option with a default of 4000 MB might be better.  The reason I chose a tension of 150 is that is works with minimal overshoots for the data set in the tutorial that is in progress.

### DIFF
--- a/src/raster/r.hydro.flatten/r.hydro.flatten.py
+++ b/src/raster/r.hydro.flatten/r.hydro.flatten.py
@@ -1,10 +1,10 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 #
 #########################################################################
 #
 # MODULE:     r.hydro.flatten
 #
-# AUTHOR(S):  Anna Petrasova <kratochanna gmail com>
+# AUTHOR(S):  Anna Petrasova <kratochanna gmail com> 
 #
 # PURPOSE:    Derive elevation of water bodies for hydro-flattening
 #
@@ -46,6 +46,12 @@
 # % key: filled_elevation
 # % required: no
 # % description: Raster map representing filled digital elevation model
+# %end
+# %option G_OPT_R_OUTPUT
+# % key: stddev_limit
+# % required: no
+# % description:Limit of standad deviation values filled as flat in filled elevation layer 
+# % Label: Standard deviation limit for flat fill 
 # %end
 # %option
 # % key: percentile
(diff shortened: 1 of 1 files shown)


petrasovaa:
See my other PR, I think r.hydro.flatten doesn't need to do filling, that can be left as postprocessing and this way we don't have to guess any tension values or methods used in r.fillnulls.

dnewcomb:
That's a better idea. Not everything needs to be solved in a single script.

petrasovaa:
@dnewcomb do you want to test the other PR or should I just merge it? I did test it.
I am closing this one then.

dnewcomb:
I can test tonight

dnewcomb:
Results are a bit different than I expected.  It had not occurred to me to
leave the original 3 cell idw partial interpolation in the high standard
deviation areas . I note the water elevations are missing for the high
standard deviation areas. I will look again in the morning.

petrasovaa:
> Results are a bit different than I expected. It had not occurred to me to leave the original 3 cell idw partial interpolation in the high standard deviation areas .

Well, this was the simplest implementation I could think of. I don't think it's a big deal, but if you find a better way, we can change that.

 I note the water elevations are missing for the high standard deviation areas. I will look again in the morning.

That seems fine to me since we are not using them.

dnewcomb:
On Fri, Jun 20, 2025 at 10:09 AM Anna Petrasova ***@***.***>
wrote:

> *petrasovaa* left a comment (OSGeo/grass-addons#1424)
> <https://github.com/OSGeo/grass-addons/pull/1424#issuecomment-2991776050>
>
> Results are a bit different than I expected. It had not occurred to me to
> leave the original 3 cell idw partial interpolation in the high standard
> deviation areas .
>
> Well, this was the simplest implementation I could think of. I don't think
> it's a big deal, but if you find a better way, we can change that.
>
My original thought was to punch a hole for the footprint of the modeled
water body and let the user fill the larger voids, but if we are confident
about the 3 cell idw interpolations into those voids, then the users'
filling the remaining voids should be as accurate.  Which is my long-winded
way of saying I agree.


> I note the water elevations are missing for the high standard deviation
> areas. I will look again in the morning.
>
> That seems fine to me since we are not using them.
>
Yes. The user can run it again without the stddev_max value if they desire
that data.

PR #1435, opened 2025-07-03
r.hydro.flatten: stddev update

dnewcomb:
This request builds off of Dr. Petrasova's r.hydro.flatten branch to remove areas of high standard deviation voids from the final filled dem. Instead of leaving the 3 cell idw interpolation in the high standard deviation areas that are above the limit, this modification trims back the voids to the original shoreline.

Some initial comparisons in recreating the surface from artificially created voids  using r.fillnulls with and without the 3 cell idw interpolation and the full void indicate that using r.fillnulls from the original void is more accurate.  A bit surprising.


petrasovaa (commented):
I didn't test, but looks fine to me! In #1426 I included change in documentation, do you want to add it here, so that we can close the other one?

dnewcomb:
That sounds great!

petrasovaa (approved):

PR #1443, opened 2025-07-24
r.lfp: Add a new longest flow path addon

HuidaeCho:
This PR adds a new longest flow path addon `r.lfp` that ports the standalone [Memory-Efficient Longest Flow Path (MELFP)](https://github.com/HuidaeCho/melfp) (Cho, 2025) OpenMP parallel algorithm to GRASS.


nilason:
Impressive work!

Clang issues some minor warnings:

```
direction.c:88:44: warning: using integer absolute value function 'abs' when argument is of floating point type [-Wabsolute-value]
   88 |                     DIR(row, col) = pow(2, abs(dir_buf[col] / 45.));
      |                                            ^
direction.c:88:44: note: use function 'fabs' instead
   88 |                     DIR(row, col) = pow(2, abs(dir_buf[col] / 45.));
      |                                            ^~~
      |                                            fabs
1 warning generated.
gettimeofday.c:28:7: warning: ISO C requires a translation unit to contain at least one declaration [-Wempty-translation-unit]
   28 | #endif
      |       ^
1 warning generated.
In file included from main.c:35:
./lfp_funcs.h:607:9: warning: variable 'up' is used uninitialized whenever 'if' condition is false [-Wsometimes-uninitialized]
  607 |     if (up_stack->n > 0) {
      |         ^~~~~~~~~~~~~~~
./lfp_funcs.h:618:12: note: uninitialized use occurs here
  618 |     return up;
      |            ^~
./lfp_funcs.h:607:5: note: remove the 'if' if its condition is always true
  607 |     if (up_stack->n > 0) {
      |     ^~~~~~~~~~~~~~~~~~~~
./lfp_funcs.h:605:5: note: variable 'up' is declared here
  605 |     struct cell up;
      |     ^
1 warning generated.
write.c:144:73: warning: unused parameter 'dir_map' [-Wunused-parameter]
  144 |                        struct outlet_list *outlet_l, struct raster_map *dir_map)
      |                                                                         ^
write.c:223:72: warning: unused parameter 'dir_map' [-Wunused-parameter]
  223 |                       struct outlet_list *outlet_l, struct raster_map *dir_map)
      |                                                                        ^
2 warnings generated.
```

The `gettimeofday.c` warning should probably be ignored. Ideally that file should be filtered out in the Makefile, if not being MSVC, but that would perhaps be difficult to implement without changes to GRASS core. Or is there a way?

HuidaeCho:
> The `gettimeofday.c` warning should probably be ignored. Ideally that file should be filtered out in the Makefile, if not being MSVC, but that would perhaps be difficult to implement without changes to GRASS core. Or is there a way?

Ideally, it should reuse `msvc/sys/gettimeofday.c` from https://github.com/OSGeo/grass/pull/6110 for Windows only. I'm not aware of any platform-specific mechanism in the core's Makefile system.

nilason:
If `_MSC_VER` is set only with CMake builds, then this there will be no issue for 8.4.x releases. In that case https://github.com/OSGeo/grass/pull/6110 would be the best solution.

PR #1459, opened 2025-08-08
i.eodag: use build in search method and refactor module

ninsbl:
This PR removes the cusom _no_fallback_search_ and replaces it with the build in search method which now can also be restricted to one provider (if the user chooses so). The build in method is significantly more performant and the custom code is no longer necessary to maintain.

At that occasion, also I choose to restructure the the main function to kind of have a decision tree for different types of search a user may request...

Finally, the PR also includes some work (and minor formating changes) towards increased ruff-check-compliance.

@HamedElgizery : Tests and any comments / suggestions would be very welcome.


HamedElgizery:
Hi @ninsbl,
Nice update! I will be taking a look ASAP. I will need a couple more days, sorry for the delay.
Thanks!

src/imagery/i.eodag/i.eodag.py
```diff
@@ -383,54 +390,58 @@ def get_aoi(vector=None):
...
 
     :return: EO products found by searching with 'search_parameters'
     :rtype: class:'eodag.api.search_result.SearchResult'
     """
-    gs.verbose("Searching for products...")
+    # Remove empty string
+    ids_set.discard("")
+    # Search for products found from options["file"] or options["id"]
+    gs.verbose(_("Searching for {} distinct ID(s).").format(len(ids_set)))
+    # Search for products found from options["file"] or options["id"]
```
HamedElgizery:
```suggestion
    # Search for products found from options["file"] or options["id"]
    gs.verbose(_("Searching for {} distinct ID(s).").format(len(ids_set)))
```
nit: double comment

src/imagery/i.eodag/i.eodag.py
```diff
@@ -383,54 +390,58 @@ def get_aoi(vector=None):
...
-                provider=options["provider"] or None,
-                productType=options["producttype"] or None,
-                count=True,
-            )
-            count = product.number_matched
-        if count > 1:
+        if module_options["producttype"] is None:
+            gs.warning(_("The producttype option is not set"))
+        product = eodag_api.search(
+            id=query_id,
```
HamedElgizery:
```suggestion
        if not module_options["producttype"]:
            gs.warning(_("The producttype option is not set"))
        product = eodag_api.search(
            id=query_id,
```

Would be an empty string.

src/imagery/i.eodag/i.eodag.py
```diff
@@ -615,60 +585,60 @@ def parse_query(query=None):
...
                     _(
-                        "Queryable <{}> could not be parsed\nDate must be ISO formated"
-                    ).format(parameter)
+                        "Queryable <{}> could not be parsed\nDate must be ISO formated",
+                    ).format(parameter),
                 )
             continue
         operator = None
         values_operators = []
         for value in map(str.strip, values.split("|")):
```
HamedElgizery:
```suggestion
        values_operators = []
        for value in map(str.strip, values.split("|")):
            operator = None
```
Must have slipped from before, default value for an operator should be None, later translated to 'eq'.

src/imagery/i.eodag/i.eodag.py
```diff
@@ -1166,69 +1172,74 @@ def main():
...
-        search_result = search_by_ids(ids_set)
-    elif "search_result" not in locals():
+        except OSError:
+            gs.fatal(
+                _(
+                    "Unable to read product IDs from file <{}>.",
+                ).format(options["file"]),
+            )
+    else:
+        # Search using given search parameters
```
HamedElgizery:
```suggestion
                ).format(options["file"]),
            )
    elif id_file:
        gs.fatal(
            _(
                "File type '{}' is not supported. Please use a '.geojson' or '.txt' file.",
            ).format(id_file.suffix.lower()),
        )
    else:
        # Search using given search parameters
```
Handle other files.

src/imagery/i.eodag/i.eodag.py
```diff
@@ -1238,40 +1249,36 @@ def main():
...
+        # Conduct parameter search
+        search_result = dag.search_all(**search_parameters)
+
+        gs.verbose(_("Filtering results..."))
+        search_result = filter_result(
+            search_result,
+            geometry if "geometry" in locals() else None,
+            queryables,
+            **options,
+        )
```
HamedElgizery, line 1271: Putting the filtering step inside the `else` block, will allow only for filtering when searching for products from scratch. 

If you agree, I think it would be better to move the filtering out of the `else` block... so that users could use the filtering when searching with IDs or files.

src/imagery/i.eodag/i.eodag.py
```diff
@@ -1110,49 +1096,69 @@ def get_range(info):
...
-    DEFAULT_OPERATOR = "eq"
+        query_dict["cloudCover (le)"] = kwargs["clouds"]
+    default_operator = "eq"
+    query_dict["queryables"] = {}
     for k, v in queryables:
         for value in v:
-            operator = value[1] if value[1] else DEFAULT_OPERATOR
-            print(f"{k} ({operator}): {value[0]}")
+            operator = value[1] or default_operator
+            query_dict["queryables"][f"{k} ({operator})"] = value[0]
```
HamedElgizery, line 1130: I have just noticed that there is a problem here when printing queryables with logical operators, for instance if we have a query as:
`query="cloudCover=30;lt | 60;gt, cloudCover=40;lt"` (The query doesn't make sense but just for illustration).

When using the `-p` flag this would print:
```
{
...
    "limit": "50",
    "queryables": {
        "cloudCover (lt)": 40.0,
        "cloudCover (gt)": 60.0
    }
}
```

While it would make sense if it prints:
```
{
...
    "limit": "50",
    "queryables": {
        "cloudCover": "( 30(lt) OR 60(gt) ) AND 40(lt)",
    }
}
```

I will create another PR after this one to patch this.

HamedElgizery (commented):
Hello @ninsbl, I have done some testing and things is looking good... disabled fallback searching is also working fine. I have left some extra fixes in suggestions. Thanks!

src/imagery/i.eodag/i.eodag.py
ninsbl: Hm, yes, indeed a question of expected behavior.

What concernes me a bit with always filtering is that `geometry` is set to the computational region by default, and also `start` and `end` are set to default values, if the user does not provide any input here. So in order to not filter away earlyer search results users would have to set those values deliberately in a way that matches the earlier search...

That said, removing duplicates would make sense also with ID-search...

HamedElgizery: Right, yeah that makes sense, since in most cases if users are searching by IDs they probably want the full list of products without filtering. Though, one of the use cases I can think of for a user to want to filter products from a file is if they have already done a search before, and have saved the footprints to a geojson. Then later, they decide they want to research for the products in that geojson file with extra filtering. Maybe then it would be a case where filtering can be helpful.

I think another way to go about it, could be to:
1. If a user is searching from scratch, use either the current computational region, or a vector map if provided. (The same exact flow as of now)
2. If a user is searching by IDs or Files, then set all filtering options to None, unless the user explicitly specifies any of them.

ninsbl: Implemented in [4dc1a51](https://github.com/OSGeo/grass-addons/pull/1459/commits/4dc1a5177741ff6db0852290bc82bd66c21b3b3d)

PR #1489, opened 2025-09-22
r.curvenumber: update the unittest

mabdazzam:
This PR fixes the unit test for the `r.curvenumber` addon.

The test failure was caused by a change in the ARC I / ARC II lookup table.
In the original implementation, ARC I and ARC II were named incorrectly.
This was corrected in PR #1432 (see commit f7bde2c), where the two tables
were swapped to match their proper definitions.

As a result, the expected curve number in the test raster changed:

- previously **87** (line 891 in the added lines, green section)  
- now **73** (line 887 in the added lines, green section)

This PR updates the expected test raster.


petrasovaa (approved):
Thanks!

PR #1508, opened 2025-10-22
r.timeofconcentration: Add a new time of concentration addon

mabdazzam:
This PR adds a new addon which calculates the hydrologic time of concentration using the Kirpich Equation.


src/raster/r.timeofconcentration/r.timeofconcentration.md
```diff
@@ -0,0 +1,118 @@
...
+- **Stream derivation.** If 'streams' is omitted, `r.watershed` generates it,
+  using 'threshold' to control density.
+- **Static Tc** Tc is a steady-state estimate; no temporal
+  routing or storage is modeled.
+
+## EXAMPLE
+
+These examples use the North Carolina sample dataset.
+
+Calculate time of concentration using r.watershed and r.flowaccumulation:
```
petrasovaa:
```suggestion
Calculate time of concentration using r.watershed and r.timeofconcentration:
```

src/raster/r.timeofconcentration/r.timeofconcentration.md
```diff
@@ -0,0 +1,118 @@
...
+
+**sbar:** Optional raster of path-average slope per cell [unitless], calculated
+as $S_{\text{avg}} = \max(\frac{\Delta z}{L}, \text{slope}_\text{{min}})$,
+where $\Delta z$ is the drop and $L$ is the length.
+
+## Notes
+
+- **SI units only.** The module processes inputs and outputs in metric system
+- **CRS-aware areas.** `r.stream.distance` returns lengths in meters, adjusted
+  to the coordinate system.
```
petrasovaa: This is somewhat confusing formulation. Either you need to explicitly state user should run it only in projects with CRS using meters, or I would try to fix the code to run in feet-based CRSs, I think I was suggesting a fix for that in one of the previous emails.

src/raster/r.timeofconcentration/r.timeofconcentration.py
```diff
@@ -0,0 +1,260 @@
...
+# % required: yes
+# % guisection: inputs
+# %end
+
+# %option G_OPT_R_INPUT
+# % key: direction
+# % description: Name of input flow direction raster map (from r.watershed or r.stream.extract)
+# % required: yes
+# % guisection: inputs
+# %end
```
petrasovaa, line 40: I realized this could be optional, no? The same way the streams are optional? I don't see why direction needs to be required and not streams. I am thinking the easiest workflow is to provide only elevation and derive everything afterward (in one r.watershed run), or use first r.watershed to derive directions and streams (or use r.stream.extract) and pass both to this tool. I just want to discourage (not necessarily forbid) user running r.watershed first, providing only directions and then internally running r.watershed again, seems inefficient.

src/raster/r.timeofconcentration/r.timeofconcentration.md
```diff
@@ -0,0 +1,118 @@
...
+These examples use the North Carolina sample dataset.
+
+Calculate time of concentration using r.watershed and r.flowaccumulation:
+
+```sh
+# set the region
+g.region -p raster=elevation
+
+# calculate positive flow accumulation and drainage directions using r.watershed
+r.watershed -sa elevation=elevation drainage=fdr threshold=10
```
petrasovaa: The threshold here doesn't do anything, did you want to output streams?

src/raster/r.timeofconcentration/r.timeofconcentration.md
```diff
@@ -0,0 +1,118 @@
...
+# use length_min parameter for coarser tc on important streams only
+r.timeofconcentration elevation=elevation direction=fdr tc=tc_nc length_min=100
+```
+
+![r_timeofconcentration example](tc_nc.png)
+*Figure: Output from r.timeofconcentration on NC dataset*
+
+![r_timeofconcentration example](tc_nc_zoomed.png)
+*Figure: Output from r.timeofconcentration on NC dataset zoomed near the
+watershed outlet*
```
petrasovaa: I think I would leave only the zoomed in figure

src/raster/r.timeofconcentration/tc_nc_zoomed.png
petrasovaa, line 1: Showing this on top of a shaded relief would look better.

src/raster/r.timeofconcentration/testsuite/test_r_timeofconcentration.py
```diff
@@ -0,0 +1,178 @@
...
+
+class TestRTimeOfConcentrationRegression(TestCase):
+    """
+    Regression test for r.timeofconcentration
+
+    what it checks:
+    verifies run-to-run stability: for a fixed region and inputs, repeated
+    executions must produce identical output statistics. each scenario runs the
+    module 'reps' times; run 1 becomes the baseline, and runs 2–reps must match
+    its r.univar statistics (min, max, mean, stddev, sum) and non-null count.
```
petrasovaa: Sorry, this is misunderstanding, with regression test we mean that the you check the output values against fixed reference values in the test (or reference data). With that we are trying to catch some regression either in the tool itself (when a change is introduced) or in the underling tools. You don't need to rerun the tool multiple times, maybe that's useful for tools that have some random component, but not here.

petrasovaa (changes requested):
Thanks!

There is missing Makefile and please see other comments.

src/raster/r.timeofconcentration/r.timeofconcentration.md
mabdazzam: Added.

src/raster/r.timeofconcentration/r.timeofconcentration.md
mabdazzam: Addressed this issue. 
r.stream.distance produces outputs (distance and difference) in meters no matter the projection. So introduced a conversion factor for the only edge case where the vertical units are not meters. If the elevation input is in feet, conversion to meters is done internally.

src/raster/r.timeofconcentration/r.timeofconcentration.py
mabdazzam: Agreed and removed the internal streams generation to keep it modular. 
Now both direction and streams should be generated outside this addon.

src/raster/r.timeofconcentration/r.timeofconcentration.md
mabdazzam: Fixed.

src/raster/r.timeofconcentration/r.timeofconcentration.md
mabdazzam: Removed the other one.

src/raster/r.timeofconcentration/tc_nc_zoomed.png
mabdazzam: I added 50% transparent relief as the basemap (100% was too dark). Please check.

src/raster/r.timeofconcentration/testsuite/test_r_timeofconcentration.py
mabdazzam: I came up with another test, where I manually calculated the tc on a calculator.
Length was calculated using r.lfp and slope and tc were calculated manually by reading the elevation raster values of start and end points of the lfp. 
Those 5 outlets and the manual tc values are added as the standard to be compared against.

src/raster/r.timeofconcentration/r.timeofconcentration.md
petrasovaa: You are still missing the streams as output.

src/raster/r.timeofconcentration/r.timeofconcentration.md
```diff
@@ -0,0 +1,122 @@
...
+Calculate time of concentration using r.watershed and r.timeofconcentration:
+
+```sh
+# set the region
+g.region -p raster=elevation
+
+# calculate positive flow accumulation and drainage directions using r.watershed
+r.watershed -sa elevation=elevation drainage=fdr threshold=10
+
+# compute the time of concentration (use same threshold as r.watershed)
```
petrasovaa:
```suggestion
# compute the time of concentration
```

src/raster/r.timeofconcentration/r.timeofconcentration.md
petrasovaa: Did you accidentally removed the zoomed one?

petrasovaa (changes requested):
Thanks, almost done, but please check the remaining comments.

src/raster/r.timeofconcentration/r.timeofconcentration.md
mabdazzam: Fixed the commands.
I made a new map but it wasn't zoomed in enough. See the new one with elevation_shade background.

src/raster/r.timeofconcentration/r.timeofconcentration.md
```diff
@@ -0,0 +1,128 @@
...
+r.timeofconcentration elevation=elevation direction=fdr streams=str tc=tc_nc
+
+# use length_min parameter for coarser tc on important streams only
+r.timeofconcentration elevation=elevation direction=fdr streams=str tc=tc_nc_250 length_min=250
+
+# if the vertical units of the DEM are in feet
+r.timeofconcentration elevation=elevation vunits=feet dir=fdr str=str tc=tc
+
+# if the vertical units of the DEM are in units other than meters or feet (e.g., cm)
+r.timeofconcentration elevation=elevation vunits=factor factor=0.1 dir=fdr str=str tc=tc
```
petrasovaa: To fit with the comment:

```suggestion
r.timeofconcentration elevation=elevation vunits=factor factor=0.01 dir=fdr str=str tc=tc
```

src/raster/r.timeofconcentration/r.timeofconcentration.html
```diff
@@ -0,0 +1,106 @@
...
+r.timeofconcentration elevation=elevation direction=fdr streams=str tc=tc_nc
+
+# use length_min parameter for coarser tc on important streams only
+r.timeofconcentration elevation=elevation direction=fdr streams=str tc=tc_nc_250 length_min=250
+
+# if the vertical units of the DEM are in feet
+r.timeofconcentration elevation=elevation vunits=feet dir=fdr str=str tc=tc
+
+# if the vertical units of the DEM are in units other than meters or feet (e.g., cm)
+r.timeofconcentration elevation=elevation vunits=factor factor=0.1 dir=fdr str=str tc=tc
```
petrasovaa: To fit with the comment:

```suggestion
r.timeofconcentration elevation=elevation vunits=factor factor=0.01 dir=fdr str=str tc=tc
```

src/raster/r.timeofconcentration/r.timeofconcentration.py
```diff
@@ -0,0 +1,263 @@
...
+    # optional
+    outlets = options.get("outlets")
+
+    slope_min = float(options["slope_min"]) if options.get("slope_min") else 1e-4
+    length_min = float(options["length_min"]) if options.get("length_min") else 10.0
+
+    vertical_units = options.get("vertical_units") or "meters"
+    if vertical_units in ("meters", "feet"):
+        if options["factor"]:
+            gs.fatal(_("Factors must be used only when vertical_units=factor"))
```
petrasovaa:
```suggestion
            gs.fatal(_("Factor must be used only when vertical_units=factor"))
```

petrasovaa (approved):
I made the image smaller. Looks good now, thank you!

PR #1515, opened 2025-10-29
t.rast.import.netcdf: Support ACDD time definition

ninsbl:
This PR adds (minimalistic) support for the data following the Attribute Convention for Data Discovery ([ACDD](https://wiki.esipfed.org/Category:Attribute_Conventions_Dataset_Discovery)) instead of CF-1.X.
Some format improvements were done for improved ruff-compliance.

Getting time from ACDD metadata was needed for reading SWOT raster data. Which comes with variables with only 2 dimensions and start and end time in a global dataset attribute... ACDD seems common enough to support such cases...

P.S.: Not sure why those old commits show up. I guess my grass8 branch is not perfectly clean. But only the two relevant files were modified anyway...


src/temporal/t.rast.import.netcdf/testsuite/test_t_rast_import_netcdf_sentinel.py
```diff
@@ -0,0 +1,264 @@
...
+
+AUTHOR(S): Stefan Blumentrath <stefan dot blumentrath at nina dot no>
+
+PURPOSE:   Test of t.rast.import.netcdf (example of a simple test of a module)
+
+COPYRIGHT: (C) 2021-2025 by Stefan Blumentrath and the GRASS Development Team
+
+This program is free software under the GNU General Public
+License (>=v2). Read the file COPYING that comes with GRASS
+for details.
```
nilason: You may change here and other files you changes.

```suggestion
SPDX-License-Identifier: GPL-2.0-or-later
```

See https://grass.osgeo.org/grass85/manuals/style_guide.html#copyright-header

ninsbl: Implemented in [03f3251](https://github.com/OSGeo/grass-addons/pull/1515/commits/03f32512669042a4a79b84a19f86540eccd1e86f) and  [269e7f7](https://github.com/OSGeo/grass-addons/pull/1515/commits/269e7f7230227ee4c2a52a87791f39731759113b)

PR #1532, opened 2025-11-19
Fixed modules: removed six dependency

### LINKED ISSUE #1346: [Bug] Remove usage of the six package
**Describe the bug**

Couple tools are still using six which is no longer a dependency and not needed with Python 3 only code.

**To Reproduce**

In an environment without _six_, you will get:

```
ModuleNotFoundError: No module named 'six'
```

**Expected behavior**

We probably have six lingering here and there in the CI and any dependency can bring six as a transitive dependency for us, but removing usage of six from our code is the first, and possibly the only, step to do.

**Screenshots**

```bash
grep -IrnE "from six|import six"
```

```text
src/raster/r.in.srtm.region/r.in.srtm.region.py:125:from six.moves.urllib import request as urllib2
src/raster/r.in.nasadem/r.in.nasadem.py:119:from six.moves.urllib import request as urllib2
src/raster/r.in.usgs/r.in.usgs.py:159:from six.moves.urllib.request import urlopen
src/raster/r.in.usgs/r.in.usgs.py:160:from six.moves.urllib.error import URLError, HTTPError
src/raster/r.in.usgs/r.in.usgs.py:161:from six.moves.urllib.parse import quote_plus
```

**Additional context**

I found this through the addon build logs artifact in the new Documentation workflow in the main repo.

JoyalMPaul:
This change resolves #1346 by removing outdated six.moves imports from r.in.srtm.region, r.in.nasadem, and r.in.usgs modules. Please let me know if I need to fix anything as this is my first contribution!

### DIFF
--- a/.github/workflows/requirements.txt
+++ b/.github/workflows/requirements.txt
@@ -1,3 +1,2 @@
 matplotlib
 numpy
-six
--- a/src/raster/r.in.nasadem/r.in.nasadem.py
+++ b/src/raster/r.in.nasadem/r.in.nasadem.py
@@ -116,7 +116,7 @@ import time
 import zipfile as zfile
 import subprocess
 import urllib
-from six.moves.urllib import request as urllib2
+import urllib.request as urllib2
 
 try:
     from http.cookiejar import CookieJar
--- a/src/raster/r.in.srtm.region/r.in.srtm.region.py
+++ b/src/raster/r.in.srtm.region/r.in.srtm.region.py
@@ -122,7 +122,7 @@ import os
 import atexit
 import numpy as np
 import subprocess
-from six.moves.urllib import request as urllib2
+import urllib.request as urllib2
 
 try:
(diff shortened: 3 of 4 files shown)


neteler (approved):

echoix (changes requested):
Can you also actually remove the installation (as a requirement) from https://github.com/OSGeo/grass-addons/blob/0df3f4a109ebb8a965e35e13bb943299cfc1352c/.github/workflows/requirements.txt#L3?

If you are able to try these modules with the change (making sure six is not available), it would be great. But keep in mind some other packages might install six as a dependency, so it might not be possible to really do that verification.

JoyalMPaul:
I removed six from requirements.txt since no other files used it, since everything else was using urllib. I also updated my branch since it was a few commits behind grass8. Hope that helps!

echoix (approved):

PR #1534, opened 2025-11-21
v.stream.profiler: Update gs.parser() call to avoid `ImportError: Cannot load backend 'WXAgg'

neteler:
Update `gs.parser()` call to avoid `ImportError: Cannot load backend 'WXAgg' (see addon compilation [log](https://grass.osgeo.org/addons/grass8/logs/v.stream.profiler.log)), inspired by [t.rast.boxplot.py](https://github.com/OSGeo/grass-addons/blob/grass8/src/temporal/t.rast.boxplot/t.rast.boxplot.py):

```python
  File "/home/neteler/.grass8/addons/v.stream.profiler/scripts/v.stream.profiler", line 404, in <module>
    main()
  File "/home/neteler/.grass8/addons/v.stream.profiler/scripts/v.stream.profiler", line 179, in main
    lazy_import_matplotlib()
  File "/home/neteler/.grass8/addons/v.stream.profiler/scripts/v.stream.profiler", line 164, in lazy_import_matplotlib
    from matplotlib import pyplot as plt
  File "/usr/lib/python3/dist-packages/matplotlib/pyplot.py", line 2336, in <module>
    switch_backend(rcParams["backend"])
  File "/usr/lib/python3/dist-packages/matplotlib/pyplot.py", line 284, in switch_backend
    raise ImportError(
ImportError: Cannot load backend 'WXAgg' which requires the 'wx' interactive framework, as 'headless' is currently running
```

Also change from `is not ""` to `!= ""` to fix these warnings:

```python
/home/neteler/.grass8/addons/v.stream.profiler/scripts/v.stream.profiler:252: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if options["outstream"] is not "":
/home/neteler/.grass8/addons/v.stream.profiler/scripts/v.stream.profiler:279: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if options["window"] is not "":
/home/neteler/.grass8/addons/v.stream.profiler/scripts/v.stream.profiler:301: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if options["window"] is not "":
/home/neteler/.grass8/addons/v.stream.profiler/scripts/v.stream.profiler:323: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if options["window"] is not "":
/home/neteler/.grass8/addons/v.stream.profiler/scripts/v.stream.profiler:357: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if options["outfile_original"] is not "":
/home/neteler/.grass8/addons/v.stream.profiler/scripts/v.stream.profiler:374: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if options["outfile_smoothed"] is not "":
```

### DIFF
--- a/src/vector/v.stream.profiler/v.stream.profiler.py
+++ b/src/vector/v.stream.profiler/v.stream.profiler.py
@@ -110,8 +110,8 @@
 # IMPORT MODULES #
 ##################
 # PYTHON
-import numpy as np
 import sys
+import numpy as np
 
 # GRASS
 from grass.pygrass.modules.shortcuts import general as g
@@ -157,15 +157,15 @@ def lazy_import_matplotlib():
     """Lazy import matplotlib modules"""
     global mpl
     global plt
+
+    # lazy import matplotlib
     try:
         import matplotlib as mpl
 
         mpl.use("WXAgg")
         from matplotlib import pyplot as plt
     except ModuleNotFoundError:
-        gs.fatal(
-            _("Matplotlib (python-matplotlib) is not installed. Please, install it.")
-        )
+        gs.fatal(_("Matplotlib is not installed. Please, install it."))
 
 
 def main():
@@ -175,10 +175,8 @@ def main():
     means that the river exits the map.
     """
 
-    # lazy import modules
+    # lazy import py modules
     lazy_import_matplotlib()
-
-    options, flags = gs.parser()
 
     # Parsing
     window = float(options["window"])
@@ -401,4 +399,4 @@ def main():
 
 
 if __name__ == "__main__":
-    main()
+    sys.exit(main(*gs.parser()))
--- a/src/vector/v.stream.profiler/v.stream.profiler.py
+++ b/src/vector/v.stream.profiler/v.stream.profiler.py
@@ -249,7 +249,7 @@ def main():
         """
 
     # Network extraction
-    if options["outstream"] is not "":
+    if options["outstream"] != "":
         selected_cats_str = list(np.array(selected_cats).astype(str))
         selected_cats_csv = ",".join(selected_cats_str)
         v.extract(
@@ -276,7 +276,7 @@ def main():
             _i += 1
         DEM.close()
         z = np.array(z)
-        if options["window"] is not "":
+        if options["window"] != "":
             x_downstream, z = moving_average(x_downstream_0, z, window)
         gs.core.percent(1, 1, 1)
     else:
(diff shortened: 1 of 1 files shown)


echoix (approved):

wenzeslaus:
If I'm reading the change correctly, this will fail because the parser result is not passed to main. All you need here is to move the parser call before the lazy load to the beginning of the main function, you don't need to deal with the `if __name__` block or parameters of the main function.

src/vector/v.stream.profiler/v.stream.profiler.py
```diff
@@ -401,4 +399,4 @@ def main():
 
 
 if __name__ == "__main__":
-    main()
+    sys.exit(main(*gs.parser()))
```
wenzeslaus, line 402: Revert this little piece.

src/vector/v.stream.profiler/v.stream.profiler.py
```diff
@@ -175,11 +175,9 @@ def main():
     means that the river exits the map.
     """
 
-    # lazy import modules
+    # lazy import py modules
```
wenzeslaus, line 178: If you want to be descriptive, say "lazy import optional dependencies". Otherwise, no comment at all would be fine too as the function name pretty much describes it except the optional piece.

src/vector/v.stream.profiler/v.stream.profiler.py
```diff
@@ -175,11 +175,9 @@ def main():
...
     lazy_import_matplotlib()
 
-    options, flags = gs.parser()
-
     # Parsing
```
wenzeslaus, line 181: Move this few lines up before the lazy import.

wenzeslaus (commented):
Per-line comments just to clarify what I meant.

src/vector/v.stream.profiler/v.stream.profiler.py
neteler: Addressed in #1535

src/vector/v.stream.profiler/v.stream.profiler.py
neteler: Addressed in #1535

src/vector/v.stream.profiler/v.stream.profiler.py
neteler: Addressed in #1535

PR #1535, opened 2025-11-21
v.stream.profiler: fix lazy import and manual references

neteler:
This PR fixes #1534 (unmerged requested changes from review) and the URLs to addon manual pages.


wenzeslaus (approved):
Looks good!

PR #1550, opened 2025-11-29
r.curvenumber: replace python implementation with C version

mabdazzam:
This PR refactors the r.curvenumber addon from Python to C to improve performance.
During a project, the current Python version was used on a nominal sized region (nrows = 47104, ncols = 36785) and `/usr/bin/time` reported that wall-clock time was 1h33m52s, whereas this C implementation took 1m37.69s (a speedup of ~57.6x).
The module interface and outputs are unchanged; only the implementation has been ported to C.


mabdazzam:
@petrasovaa could you please take a look?

petrasovaa (approved):
Great improvement, the tests pass so this is good to go!

PR #1593, opened 2026-01-08
r.in.vect: option added to select features to import

ecodiv:
New option to query attributes for selecting features

### DIFF
--- a/src/raster/r.in.vect/r.in.vect.html
+++ b/src/raster/r.in.vect/r.in.vect.html
@@ -24,12 +24,8 @@ the vector map's attribute table. The values in that column will be
 used as raster values in the output raster map.
 
 <p>
-Optionally, the only features selected with the <b>where</b> option can be 
+Optionally, only features selected with the <b>where</b> option can be
 imported.
-
-v.in.ogr where="highway >< ''"
-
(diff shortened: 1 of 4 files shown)


src/raster/r.in.vect/r.in.vect.html
```diff
@@ -23,6 +23,14 @@ <h2>DESCRIPTION</h2>
 the vector map's attribute table. The values in that column will be
 used as raster values in the output raster map.
 
+<p>
+Optionally, the only features selected with the <b>where</b> option can be 
+imported.
+
+v.in.ogr where="highway >< ''"
```
neteler: Please mark this line as GRASS code (rather than plain text)

neteler:
Please remove `src/raster/r.in.vect/.Rhistory` from the PR

src/raster/r.in.vect/r.in.vect.html
ecodiv: That example should not have been there (something for in the example section). Removed it.

ecodiv:
> Please remove `src/raster/r.in.vect/.Rhistory` from the PR

Thanks Markus, that code example shouldn't have been there, I removed it.

PR #1603, opened 2026-01-19
r.in.vect: new heatmap (count overlapping features) added

ecodiv:
- Some refactoring of existing options
- New option to create a heatmap of overlapping features: count number of overlapping features per raster cell and assign that count as raster value
- New option to print column names and data type of attribute table (work around because it is not possible, as far as I know, to create a dropdown with column names in those fields were the user need to select/define column name.


nilason:
All green again!

ecodiv:
Great, thanks!

PR #1611, opened 2026-01-27
r.forestfrag: added memory and nprocs options

ecodiv:
Memory and nprocs options added to be used in `r.neighbors`

### DIFF
--- a/src/raster/r.forestfrag/r.forestfrag.py
+++ b/src/raster/r.forestfrag/r.forestfrag.py
@@ -335,7 +335,7 @@ def main(options, flags):
     else:
         indexfin2 = opl
     mapcalc(
-        "eval(" "dpf = $pf - $pff,"
+        "eval(dpf = $pf - $pff,"
         # Individual classes
         "patch = if($pf < 0.4, 1, 0),"
         "transitional = if($pf >= 0.4 && $pf < 0.6, 2, 0),"
--- a/src/raster/r.forestfrag/r.forestfrag.py
+++ b/src/raster/r.forestfrag/r.forestfrag.py
@@ -335,7 +335,8 @@ def main(options, flags):
     else:
         indexfin2 = opl
     mapcalc(
-        "eval(dpf = $pf - $pff,"
+        "eval("
+        "dpf = $pf - $pff,"
         # Individual classes
         "patch = if($pf < 0.4, 1, 0),"
         "transitional = if($pf >= 0.4 && $pf < 0.6, 2, 0),"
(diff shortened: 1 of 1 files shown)


src/raster/r.forestfrag/r.forestfrag.py
```diff
@@ -325,8 +335,7 @@ def main(options, flags):
     else:
         indexfin2 = opl
     mapcalc(
-        "eval("
-        "dpf = $pf - $pff,"
+        "eval(dpf = $pf - $pff,"
```
petrasovaa: This is unrelated change, so perhaps remove that.

src/raster/r.forestfrag/r.forestfrag.py
```diff
@@ -325,8 +335,7 @@ def main(options, flags):
     else:
         indexfin2 = opl
     mapcalc(
```
petrasovaa, line 337: The mapcalc could be parallelized too. But that may need more testing. I believe passing nprocs should not result in error in 8.4 (mapcalc was parallelized in 8.5), it should be ignored. The other aspect is the problem with too many open files, the expressions here have very many rasters, so together with many cores this might be a problem (number of cores multiplies that). So again, maybe add that only if you are able to test it well.

ecodiv: Yes, I wasn't sure about the effect on 8.4, but it would be nice to implement. Do you really think there will be an issue with too many open files? The one that opens most is this mapcalc statement, and there are not that many maps, aren't there? And how does this (internally) work with eval, does it at each of the sub-statements open files?

What about making a feature request for this, and for now implement what is proposed here? And with that, would you be OK with merging this?

ecodiv: see below (I was answering in VS Code, but I see that doesn't put the reply directly under your comment )

src/raster/r.forestfrag/r.forestfrag.py
ecodiv: It is only a formatting thing, isn't it? Changed it back.

src/raster/r.forestfrag/r.forestfrag.py
ecodiv: > I believe passing nprocs should not result in error in 8.4 (mapcalc was parallelized in 8.5), it should be ignored. 

Not in general, it seems. I added it to r.mapcalc calls in another addon (_r.mess_) and running it in 8.4.1 gave an error about an unknown parameter.  I solved it by adding a version check, so easy to solve.

PR #1612, opened 2026-01-27
r.sun.hourly: add option to provide horizon information map as input.

ecodiv:
The module has the `m` to enable the low-memory version of the program. However, using it fails with the message that using this flag requires the user to provide a *horizon information input map basename* and the *horizon_step*.

I did not find a clear description of this in the manual page of r.sun or r.sun.hourly, but adding the input options does solve the error message.

P.s., as the module takes a lot of memory and therefore easily runs out of memory, would it be an idea to expose the `npartitions` parameter from `r.sun` in the `r.sun.hourly` as well?

### DIFF
--- a/src/raster/r.sun.hourly/r.sun.hourly.py
+++ b/src/raster/r.sun.hourly/r.sun.hourly.py
@@ -103,9 +103,8 @@
 # % required: no
 # % description: Name of input raster map containing latitudes [decimal degrees]
 # %end
-# %option
+# %option G_OPT_R_BASENAME_INPUT
 # % key: horizon_basename
-# % type: string
 # % required: no
 # % description: The horizon information input map basename
 # %end
--- a/src/raster/r.sun.hourly/r.sun.hourly.py
+++ b/src/raster/r.sun.hourly/r.sun.hourly.py
@@ -103,21 +103,18 @@
 # % required: no
 # % description: Name of input raster map containing latitudes [decimal degrees]
 # %end
-
 # %option
 # % key: horizon_basename
 # % type: string
 # % required: no
 # % description: The horizon information input map basename
 # %end
-
 # %option
 # % key: horizon_step
 # % type: double
 # % required: no
 # % description: Angle step size for multidirectional horizon [degrees]
 # %end
-
 # %option G_OPT_R_INPUT
 # % key: long
 # % required: no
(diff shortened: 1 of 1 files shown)


src/raster/r.sun.hourly/r.sun.hourly.py
```diff
@@ -103,6 +103,18 @@
 # % required: no
 # % description: Name of input raster map containing latitudes [decimal degrees]
 # %end
+# %option
+# % key: horizon_basename
+# % type: string
```
petrasovaa:
```suggestion
# %option G_OPT_R_BASENAME_INPUT
# % key: horizon_basename
```

petrasovaa:
> P.s., as the module takes a lot of memory and therefore easily runs out of memory, would it be an idea to expose the `npartitions` parameter from `r.sun` in the `r.sun.hourly` as well?

I have never used it, but we can expose it, no harm in that.

ecodiv:
> > P.s., as the module takes a lot of memory and therefore easily runs out of memory, would it be an idea to expose the `npartitions` parameter from `r.sun` in the `r.sun.hourly` as well?
> 
> I have never used it, but we can expose it, no harm in that.

I included it, tested it on a dataset, worked fine. It tests pass, OK, to merge?

ecodiv:
@petrasovaa based on your suggestion for r.forestfrag, I thought I try to set nprocs when running r.mapcalc and r.slope.aspect in this addon (now this is only used in r.sun). And I ran a few tests without a problem on 8.5 and 8.4.1. Would there be a reason to not add those to aforementioned functions in this add-on?

src/raster/r.sun.hourly/r.sun.hourly.py
```diff
@@ -227,6 +238,12 @@
 # % description: If not specified, r.sun default will be used.
 # %end
 # %option
+# % key: npartitions
```
petrasovaa, line 241: I am not sure what is the default when not specified but I would add required: no

src/raster/r.sun.hourly/r.sun.hourly.py
```diff
@@ -589,6 +628,7 @@ def main():
         rsun_flags += "m"
     if flags["p"]:
         rsun_flags += "p"
+    partitions = int(options["npartitions"])
```
petrasovaa: Since npartition is not required, I would check if not empty before converting to int.

src/raster/r.sun.hourly/r.sun.hourly.py
ecodiv: The default is set here, and in r.sun, to 1 (= no partitioning).

src/raster/r.sun.hourly/r.sun.hourly.py
ecodiv: Done.. not sure it is needed, but added a else: partition = 1. Can't hurt either as that is the default in r.sun anyway.

src/raster/r.sun.hourly/r.sun.hourly.py
ecodiv: Added required: no

petrasovaa (approved):

PR #1592, opened 2026-01-06
cmake: add initial support for CMake build

nilason:
Added support for:

- i.hyper
- i.landsat
- i.landsat8.swlst
- i.points.auto
- r.flowaccumulation
- r.hypso
- r.in.vect
- r.skyline
- wx.metadata

Can be tested in combination with https://github.com/OSGeo/grass/pull/6878.

PR #1618, opened 2026-02-13
cmake: add math lib to r.flowaccumulation and r.skyline

nilason:
Add math library as dependency to r.flowaccumulation and r.skyline.

Addresses issue raised in https://github.com/OSGeo/grass-addons/pull/1592#issuecomment-3898922641.

### DIFF
--- a/src/raster/r.flowaccumulation/CMakeLists.txt
+++ b/src/raster/r.flowaccumulation/CMakeLists.txt
@@ -4,6 +4,8 @@ project(r.flowaccumulation)
 
 include(build_addon)
 
+find_M()
+
 set(sources
     accumulate_c.c
     accumulate_cm.c
@@ -52,5 +54,6 @@ set(doc_files
 build_addon(
   ${PROJECT_NAME}
   GRASSLIBS gis raster
+  DEPENDS LIBM
   SOURCES ${sources}
   DOCFILES ${doc_files})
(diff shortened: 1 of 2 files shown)


wenzeslaus (approved):
Thanks. I was looking forward to seeing the dependency handling in addons with CMake.

src/raster/r.flowaccumulation/CMakeLists.txt
```diff
@@ -52,5 +54,6 @@ set(doc_files
 build_addon(
   ${PROJECT_NAME}
   GRASSLIBS gis raster
```
cwhite911, line 56: Based on the original Makefile it looks like the `vector` , `dbmibase`,  `dbmiclient`, and `dbmidriver` need to be included as well.

cwhite911 (dismissed):

src/raster/r.flowaccumulation/CMakeLists.txt
nilason: It is my experience, that the Makefile doesn't always tell the truth. There is not seldom more dependencies included than needed. I usually check for `#include` in the addon directory, which gives an indication of what is definitely needed, and test the addon for building with them. Sometimes it happens there is a "#include" (and thus library) missing, then both the `#include` line and the dependency should be added.

A related note: `vector` and `dbmi*` often goes hand in hand, however `dbmidriver` is rarely needed (worth testing and check the code).

nilason: Forgot to mention explicitly: either `vector` or `dbms*` is needed for r.flowaccmulation.

petrasovaa (approved):
I managed to compile it, and it works! Thank you

PR #1621, opened 2026-02-14
docs: Partial HTML formatting

echoix:
Used a combination of prettier, biome, and dprint to solve some issues, that prevent some html parsers to work.

Only committed what I ended up having when my time available for it ended up.


echoix:
Maybe have to wait for the super-linter PR to be merged.

PR #1628, opened 2026-02-18
cmake: add support for temporal Addons

nilason:
Add CMake support for temporal Addons.


src/temporal/t.stac/libstac/CMakeLists.txt
cwhite911, line 1: Glad to see you addressed how to handle the python lib dependencies here!

cwhite911 (approved):
Looks good to me!

src/temporal/t.stac/libstac/CMakeLists.txt
nilason: 😄 
(we already have this in repo with `i.hyper`)

PR #1634, opened 2026-02-22
t.rast.import.netcdf: fix GDAL subdataset handle leak in parse_netcdf

ashnaaseth2325-oss:
## Summary

This PR fixes improper GDAL dataset cleanup in `parse_netcdf()` located in:

```
src/temporal/t.rast.import.netcdf/t.rast.import.netcdf.py
```

When importing multi-variable netCDF files, the function opened all subdatasets at once and did not correctly release them during processing. Additionally, the parent dataset (`ncdf`) remained open longer than necessary in the subdataset branch.

This could lead to excessive simultaneous open file descriptors, particularly when running with `nprocs > 1`.

---

## Problem Description

Subdatasets are opened using:

```python
sds = [[gdal.Open(s[1])] + s for s in sds]
```

Inside the processing loop, the intended cleanup was:

```python
# Close open GDAL datasets
s_d = None
```

However, `s_d` is only a loop variable. Reassigning it does not affect the dataset stored in `sds[i][0]`. As a result, all opened GDAL handles remained alive until:

```python
sds = None
```

at the end of the function.

In addition, the parent dataset:

```python
ncdf = gdal.Open(in_url)
```

was not explicitly released in the subdataset path and remained open until function return.

---

## Steps to Reproduce

1. Use a netCDF file containing many subdatasets (e.g., ERA5/CMIP6 multi-variable file).
2. Run:

```bash
t.rast.import.netcdf input=large_file.nc output=test_strds nprocs=8
```

3. Monitor file descriptors during execution:

```bash
ls /proc/<pid>/fd | wc -l
```

Before this fix:

* File descriptors increase proportionally to number of subdatasets.
* With multiprocessing, the count scales by number of workers.
* On systems with default `ulimit -n`, this may lead to:

  ```
  OSError: [Errno 24] Too many open files
  ```

---

## Impact

This affects:

* Multi-variable netCDF imports
* Batch workflows
* Multiprocessing usage (`nprocs > 1`)
* Large climate or forecast datasets

Peak simultaneous handles per worker were:

```
N subdatasets + 1 parent dataset
```

With multiprocessing, this multiplies across workers.

This can cause:

* File descriptor exhaustion
* Import interruption
* Increased memory pressure

---

### 1. Correct per-subdataset cleanup

Changed:

```python
s_d = None
```

to:

```python
s_d[0] = None
```

This correctly drops the reference to the GDAL dataset stored in the list and allows the GDAL destructor to close the handle immediately.

---

### 2. Release parent dataset earlier

After subdataset enumeration:

```python
sds = [[gdal.Open(s[1])] + s for s in sds]
ncdf = None
```

The parent container is no longer required and is now explicitly released before entering the processing loop.

---

## Result After Fix

* Only the currently processed dataset remains open.
* Parent dataset is closed as soon as possible.
* File descriptor usage remains stable.
* Multiprocessing imports no longer scale FD usage with number of variables.
* Resource handling now matches the comment:

  ```
  # Close open GDAL datasets
  ```

---

## Notes on Testing

A follow-up improvement could include a regression test verifying that file descriptor counts remain stable across repeated `parse_netcdf()` calls.

This would align with the Addons unit testing recommendations.

### DIFF
--- a/src/imagery/i.ann.maskrcnn/i.ann.maskrcnn.detect/i.ann.maskrcnn.detect.py
+++ b/src/imagery/i.ann.maskrcnn/i.ann.maskrcnn.detect/i.ann.maskrcnn.detect.py
@@ -495,6 +495,7 @@ def parse_instances(
                 target.SetGeoTransform(trans)
                 target.SetProjection(proj)
                 target.FlushCache()
+                del target
                 mList.append(maskName)
                 if class_ids[index] not in cList:
                     cList.append(class_ids[index])
@@ -573,6 +574,7 @@ def parse_instances(
                 target.SetGeoTransform(trans)
                 target.SetProjection(proj)
                 target.FlushCache()
+                del target
                 mList.append(maskName)
                 if class_ids[index] not in cList:
                     cList.append(class_ids[index])
--- a/src/temporal/t.rast.import.netcdf/t.rast.import.netcdf.py
+++ b/src/temporal/t.rast.import.netcdf/t.rast.import.netcdf.py
@@ -891,6 +891,8 @@ def parse_netcdf(
 
         # Open subdatasets to get metadata
         sds = [[gdal.Open(s[1])] + s for s in sds]  # noqa: RUF005
+        # Parent dataset no longer needed; close before processing loop
+        ncdf = None
     elif not sds and ncdf.RasterCount == 0:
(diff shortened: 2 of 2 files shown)


wenzeslaus:
Please read the AI policy in the contributing file of our main repo (OSGeo/grass).

Issue #975, opened 2023-10-31
[Bug] testsuite: do not run tests for deprecated module


ninsbl:
**testsuite for addons**
For example, the _r.in.pdal_ module is deprecated in addons: https://github.com/OSGeo/grass-addons/blob/grass8/src/raster/r.in.pdal/DEPRECATED

However, the testsuite for that module runs never the less and of course fails, e.g.:
https://github.com/OSGeo/grass-addons/actions/runs/6706180778/job/18222154754#step:17:303

No testsuite should run for deprecated modules...
Not sure how to create that exception in the testsuite, or if the test map simply should be renamed so that tests are not picked up...

neteler:
I guess that things happen here:

https://github.com/OSGeo/grass-addons/blob/82d59ce1424be8d7fac485836b3af5c00bba3b33/.github/workflows/ci.yml#L91

which calls `make` in the respective subdirs of `src/`.

Here it appears that `DEPRECATED` modules should be ignored:

https://github.com/OSGeo/grass-addons/blob/82d59ce1424be8d7fac485836b3af5c00bba3b33/src/raster/Makefile#L5

but apparently that mechanism isn't used?

ninsbl:
It seems to me, that _r.in.pdal_ is not compiled, so the `DEPRECATED` file is taken into account in the make step.

When the unittest is started here: https://github.com/OSGeo/grass-addons/blob/82d59ce1424be8d7fac485836b3af5c00bba3b33/.github/workflows/test.sh#L15C16-L15C34
I guess all _testsuite_ directory in the source tree are picked up. Maybe a solution could be to add an exception for a `DEPRECATED` file in the function here:
https://github.com/OSGeo/grass/blob/7089dc6ef72591b807c8ab430e9b94f74f55f38b/python/grass/gunittest/loader.py#L67

wenzeslaus:
Before adding more code to _grass.gunittest_, I suggest:

1. Add `.gunittest.cfg` and exclude those tests. Deprecated code is to be removed some day which nicely fits with the line in the configure file which also should be removed eventually (other ignored files mean failing tests, so these need to be resolved too just like the line for deprecated code).
2. Consider removing the deprecated code. You can't get that with g.extension anyway, no? If that's the case, you need to use Git to get the code. In that case, you can dig up the code from an old version using Git if you really need it.

neteler:
[...]

PR #1635, opened 2026-02-26
Remove deprecated raster modules to prevent CI running tests (Fixes #…

ArthPatel1502:
As discussed in #975, the deprecated modules are removed entirely to prevent CI from running tests on obsolete code.


petrasovaa (approved):
Looks good! Thanks!

Issue #1637, opened 2026-02-27
[Bug] r.neighborhoodmatrix: Fix incorrect parameter in manual page example


tangelll:
**r.neighborhoodmatrix**

**Describe the bug**

The documentation (manual page) of the r.neighborhoodmatrix addon contains an incorrect example. The example uses the parameter n= which is not recognized by the module. It should be input= (or in=).

**To Reproduce**

1.Open a GRASS GIS session.

2.Install the addon: g.extension extension=r.neighborhoodmatrix

3.Run the example command from the manual:
```sh
r.mapcalc "bc_int = int(boundary_county_500m)"
r.neighborhoodmatrix in=bc_int sep=comma
r.neighborhoodmatrix -l n=some_raster sep=comma output=neighbors.csv
```
4.See error: ERROR: Unknown option: 'n'

**Expected behavior**
The command should run successfully, recognizing the input raster. The documentation example should use input= or in= instead of n=.

**System description (please complete the following information):**

- Operating System: Operating System: macOS (M4 chip)
- GRASS GIS version  8.5

**To Fix**

- Change the n=some_raster to in=some_raster

tangelll:
I would like to work on this issue and fix the documentation. Could you please assign it to me?

PR #1638, opened 2026-02-27
r.neighborhoodmatrix: Fix incorrect parameter

tangelll:
Describe the bug

The documentation (manual page) of the r.neighborhoodmatrix addon contains an incorrect example. The example uses the parameter n= which is not recognized by the module. It should be input= (or in=).

#1637 This pull request open for this Issue


tangelll:
@mlennert  Could you please review and approve the test results?

veroandreo (approved):

echoix (dismissed):
Please apply the same change in the .md file. From grass8.5 and further, it is the source of truth to create the html files, and web docs. 8.4 still needs the html files

petrasovaa (changes requested):
This is minor, but let's use full name of the parameter "input", so please change also the example above.

petrasovaa (approved):
Thanks!

PR #1640, opened 2026-02-27
cmake: add support for misc Addons

nilason:
Add CMake support for misc Addons.

(`m.eigensystem` being a Fortran based addon, is not included.)


echoix (approved):
Fortran works with CMake, but probably not on our side yet

nilason:
> Fortran works with CMake, but probably not on our side yet

I managed to compile it, without any particular changes, but g.extension failed for some reason. Will need to return to that one later.

PR #1664, opened 2026-03-31
Fix CI to fail on test failures, closes #946

### LINKED ISSUE #946: Require the tests to pass in CI
The CI is configured for the tests to run and one can inspect the results, but the CI does not fail if the tests don't succeed.

laasyapriya-code:
Fix CI to fail on test failures #946

Add || exit 1 to test.sh command in ci.yml so GitHub Actions fails when tests fail.

Fixes https://github.com/OSGeo/grass-addons/issues/946

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -142,7 +142,7 @@ jobs:
         timeout-minutes: 45
         run: |
           cd grass-addons/src
-          ../.github/workflows/test.sh --config ../.gunittest.cfg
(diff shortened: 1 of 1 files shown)


echoix:
There is an extra space or two at the beginning of the line that seem unexpected.

echoix:
I don't think this really does what you think it does.

The "problem" isn't there, it's because not all tests work, and there's a percentage of passing test _**files**_ that is set in the script called. The solution would be to bump the percentage to 100 instead of 81. But then no PR could be merged, because of test failures related to another addon than the one concerned.

PR #1665, opened 2026-03-31
CI: disable PDAL in cronjob compilation

neteler:
Since 2022, PDAL is no longer a Debian package (see [here](https://tracker.debian.org/pkg/pdal).

Subsequently, the cronjobs are adjusted in this PR to avoid `configure` failures (after today's Debian update on grass.osgeo.org).


nilason (approved):

neteler:
So, let's monitor the cronjobs in the next 24hs :)

echoix:
Does that degrade the Doxygen docs because of the missing dependency? (Some code paths now skipped, so not documented, like enum values or thing of the like)

neteler:
Possibly. Does anyone have an idea how to get PDAL back onto the Debian server? I'd like to avoid manual compilation.

PR #1678, opened 2026-04-12
r.learn.ml2/rlearnlib: np.object is deprecated, changing to object

ecodiv:
np.object is deprecated in recent versions of NumPy


echoix (approved):
The change was done according to https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated recommendations

PR #1682, opened 2026-04-29
r.lake.series: Parallelization of the tool

gureckyfrantisek:
This PR introduces parallel processing to the r.lake.series tool.

It was marked with a TODO note and the parameter nprocs was unused.


petrasovaa (changes requested):
Keep the script API, no need to rewrite it for pygrass. While the multiprocessing pool is fine and we have used it in the past, the modern solution would use ThreadPoolExecutor, threads are fine here and should be technically less overhead then processes, since the r.lake are already launching separate process.

gureckyfrantisek:
Hi, I discussed this with @landam and tried to implement a solution simillar to how t.rast.mapcalc does it.

The previous solution you reviewed didn't work for me and ended with an error, I had local uncommited changes before I was ready to open this PR for review.

I'll have a look at the ThreadPoolExecutor as you mentioned.

src/raster/r.lake.series/r.lake.series.py
```diff
@@ -122,6 +122,8 @@
 from grass.script import core as gcore
 import grass.temporal as tgis
 from grass.exceptions import CalledModuleError
+from grass.pygrass.modules import Module
```
landam: As discussed, there’s no need to use pygrass here, as it hasn’t been used so far. Continue using `grass.script` to run command as before.

landam (changes requested):

gureckyfrantisek:
@landam I used ThreadPoolExecutor now, I tried to keep the implementation simple. This version works correctly for me, the processing doesn't stop on first error, is that the expected behavior?

I also removed the stale comments related to this PR.

gureckyfrantisek:
> This version works correctly for me
> 

It seems I tested a different version than this code, I forgot to reinstall the extension, so this doesn't work, I had a working version I'll commit.

src/raster/r.lake.series/r.lake.series.py
```diff
@@ -206,33 +229,39 @@ def main():
...
+            outputs[i],
+            water_level,
+            kwargs,
+            pass_flags,
+            gcore.overwrite(),
+        )
+        for i, water_level in enumerate(water_levels)
+    ]
+
+    use_cores = min(int(nprocs), len(tasks))
```
petrasovaa: The issue is nprocs=0 (the default in GRASS 8.5), which with openmp code in C cleanly translates to available cores (negative values mean number of available cpus minus that value). But it will result in error here. There are more ways to address it here, ideally, we should add a library function for that you could use here, but since that is not available you could do something like that (I wrote that recently in a another PR):

```python
def _available_cpus():
    """Number of CPUs this process may actually use.

    Prefers affinity-aware sources over ``os.cpu_count()``, which reports
    the host total and overcounts in containers and cgroup-limited jobs.
    """
    if hasattr(os, "process_cpu_count"):  # Python 3.13+
        return os.process_cpu_count() or 1
    if hasattr(os, "sched_getaffinity"):  # Linux
        return len(os.sched_getaffinity(0))
    return os.cpu_count() or 1


def _resolve_nprocs(nprocs):
    """Resolve G_OPT_M_NPROCS into a worker count for ThreadPoolExecutor.

    Mirrors the semantics of G_set_omp_num_threads() in
    lib/gis/omp_threads.c: 0 means use all available cores, a positive
    number is used as-is, a negative number means cpu_count + nprocs
    (clamped to at least 1). Belongs in a library helper eventually.
    """
    nprocs = int(nprocs)
    if nprocs > 0:
        return nprocs
    available = _available_cpus()
    if nprocs == 0:
        return available
    return max(1, available + nprocs)
```

It would be fine to simplify the available cpus to os.cpu_count, in most use cases it will be ok:

```python
def _resolve_nprocs(nprocs):
    nprocs = int(nprocs)
    if nprocs > 0:
        return nprocs
    available = cpu_count()
    if nprocs == 0:
        return available
    return max(1, available + nprocs)
```

gureckyfrantisek: I didn't know that, thanks for the comment. I will work this into the logic.

gureckyfrantisek:
@petrasovaa I added the check but running r.lake.series with nproc=0 results in:

```console
ERROR: Value <0> out of range for parameter <nproc>
	Legal range: 1-
```

If 0 is the default, the parameter should be also adjusted to allow such value, the definition doesn't currently support it here:

```python
# %option
# % key: nproc
# % type: integer
# % label: Number of processes to run in parallel
# % required: no
# % answer: 1
# % options: 1-
# %end
```

I tested removing the options 1- line and the tool doesn't break with any nproc input with the nprocs resolution, so it should be safe to remove to fully match the behavior you described.

src/raster/r.lake.series/r.lake.series.py
```diff
@@ -89,7 +89,7 @@
 # %option
```
petrasovaa: Just use G_OPT_M_NPROCS

petrasovaa: And then the key is "nprocs"

src/raster/r.lake.series/r.lake.series.py
```diff
@@ -158,6 +159,57 @@ def remove_raster_maps(maps, quiet=False):
...
+        )
+
+    except CalledModuleError as e:
+        # Show the error message
+        gcore.error(f"r.lake failed for output <{output}>: {e}")
+        return None
+    return output
+
+
+def _available_cpus():
```
landam: Introducing `_available_cpus()` and `_resolve_nprocs()` at the level of a single addon is excessive. These functions belong in the core GRASS package/library; otherwise, they will end up duplicated across every GRASS tool that relies on ThreadPoolExecutor.

src/raster/r.lake.series/r.lake.series.py
```diff
@@ -206,33 +260,39 @@ def main():
...
+            pass_flags,
+            gcore.overwrite(),
+        )
+        for i, water_level in enumerate(water_levels)
+    ]
+
+    use_cores = _resolve_nprocs(nprocs)
+
+    # Run tasks
+    with ThreadPoolExecutor(max_workers=use_cores) as executor:
```
landam, line 271:
```suggestion
    with ThreadPoolExecutor(max_workers=int(nprocs)) as executor:
```

I think it would be enough for this PR. Anything better should be implemented on `grass` package level and not here.

landam (changes requested):

src/raster/r.lake.series/r.lake.series.py
petrasovaa: This is not going to work for nprocs=0, see my [comment](https://github.com/OSGeo/grass-addons/pull/1682#discussion_r3204694673). My intention is to include it in the library, but that would be for 8.6. I suggested a simpler solution in the comment.

gureckyfrantisek: I will use the simpler solution to not overcomplicate it here.

gureckyfrantisek: I left just the _resolve_nprocs using the given code with os.cpu_count() as fallback instead of the _available_cpus function.

src/raster/r.lake.series/r.lake.series.py
```diff
@@ -86,13 +86,12 @@
...
 # %end
-# %option
-# % key: nproc
+# %option G_OPT_M_NPROCS
+# % key: nprocs
 # % type: integer
-# % label: Number of processes to run in parallel (currently ignored)
+# % label: Number of processes to run in parallel
 # % required: no
 # % answer: 1
```
petrasovaa: Just keep the defaults
```suggestion
# %option G_OPT_M_NPROCS
```

gureckyfrantisek:
I have now pushed the final requested changes, so this should be ready to run final checks.

petrasovaa:
> I have now pushed the final requested changes, so this should be ready to run final checks.

Next time please use pre-commit to avoid the formatting issues:
https://github.com/OSGeo/grass/blob/main/doc/development/style_guide.md#using-pre-commit

gureckyfrantisek:
Sorry about that. Should be resolved now.

petrasovaa (approved):
Thanks!

landam (approved):

PR #1687, opened 2026-05-08
CI(Renovate): Override base branch and ignore pinning of grass action from main

echoix:
Since using the config from the main repo, a warning that the base branches listed in the main repo doesn't exist here has appeared. This PR tries to fix this by setting the base branch patterns, hoping that it will override the config that was extended.

Second part of the PR tries to ignore the pinning of the github action that comes from the main branch of the grass repo. We control both of these, and the grass-addons renovate PR is updated for each commit of the main repo, since we are not merging it. See https://github.com/OSGeo/grass-addons/pull/1683

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -7,6 +7,7 @@
         "github>OSGeo/grass:renovate.json5",
         "customManagers:biomeVersions",
     ],
+    "baseBranchPatterns": [ "$default" ],
 
     "packageRules": [
         {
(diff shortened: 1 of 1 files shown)


echoix:
How come the required checks were not respected here, but were respected in my other previous PR? Were the rules temporarily suspended 15 min ago? @wenzeslaus

echoix:
> How come the required checks were not respected here, but were respected in my other previous PR? Were the rules temporarily suspended 15 min ago? @wenzeslaus

Never mind, it was the mobile app that wasn’t updated.

But it still means it didn’t get an (optional) review to auto-merge

PR #1692, opened 2026-05-09
cronjob: updates required for GRASS 8.5.0 release

neteler:
This PR extends #1689 and updates all cronjobs as follows:

Overall:
- correctly define GRASS versions (as to be modified after GRASS 8.5.0 release)
- update of comments
- disabled "old" (8.4) as no longer supported

Scripts "preview" (8.6) and "current" (8.5):
- fetch HTML manual pages from GitHub (created via GitHub "documentation" action)
- remove internal creation of HTML manual
- incorporated `fetch_unpack_manual_GHA.sh` functionality
- ~~added/synced sphinx support~~ (removed in b90e6d3fce214f48777633a50b8862ed23e7a0ed)
- removed sitemap creation as being part of  GitHub "documentation" action artifacts
- remove addon complilation
- "preview": SEO: inject canonical link into versioned manual pages

Script `gh_cli_download_artifact.sh`:
- added multi branch support

~~Removed `fetch_unpack_manual_GHA.sh` as no longer needed and updated cronjob list accordingly.~~

On the server grass.osgeo.org: "grass-stable/" and "grass-devel/" are true directories (and neither Apache redirects nor links on disk). They are sync'ed in above scripts using `rsync` from versioned source directories.


utils/cronjobs_osgeo_lxd/gh_cli_download_artifact.sh
echoix, line 71: This doesn't go fetch the latest commit-wise I think. It's only the latest ran. So if we do a rerun of an older failed job, like 3 months later, it'll go get that artifact, if my understanding is right. And that's not what we want.

utils/cronjobs_osgeo_lxd/cron_grass_current_stable_build_binaries.sh
```diff
@@ -324,28 +263,15 @@ cp -p $MAINDIR/.grass$GMAJOR/addons/logs/* $TARGETMAIN/addons/grass$GMAJOR/logs/
...
 TARGETHTMLDIRSTABLE=$TARGETMAIN/grass-stable/manuals/
-mkdir -p $TARGETHTMLDIRSTABLE $TARGETHTMLDIRSTABLE/addons
-# cleanup from previous run
-rm -rf /tmp/addons
-\mv $TARGETHTMLDIRSTABLE/addons /tmp
-rm -f $TARGETHTMLDIRSTABLE/*.*
-(cd $TARGETHTMLDIRSTABLE ; rm -rf barscales colortables icons northarrows)
 # clone manual pages
-cp -rp $TARGETHTMLDIR/* $TARGETHTMLDIRSTABLE/
+rm -rf $TARGETHTMLDIRSTABLE
```
echoix, line 261: So that means that temporarily, the docs will be unavailable if they are deleted before being rebuilt? (If the process doesn't finish its even worse).

What about having multiple "staging" folders, not served directly, that are like grass85-20260509205104 or something unique enough and the rebuild/unzipping of artifact is done in there, and when done, a symlink swaps what folder is served, atomically. Keep last 2-3 versions on the server. You could swap the symlink to a previous version if something goes wrong.

For not having these folders served but available to be served as symlink, there might be some friction with the configuration of some web servers, which might try to prevent you from serving something outside what is already publicly served.

utils/cronjobs_osgeo_lxd/cron_grass_current_stable_build_binaries.sh
```diff
@@ -324,28 +263,15 @@ cp -p $MAINDIR/.grass$GMAJOR/addons/logs/* $TARGETMAIN/addons/grass$GMAJOR/logs/
...
+rsync -a $TARGETHTMLDIR/ $TARGETHTMLDIRSTABLE/
```
echoix, line 262: Is there a flag needed to propagate the deletions (if the line above is changed according to my other observation)?

utils/cronjobs_osgeo_lxd/cron_grass_preview_build_binaries.sh
```diff
@@ -13,13 +13,16 @@
...
 # Preparations, on server (neteler@grasslxd:$):
-
+# - git clone the correct GRASS source code version into $HOME/src/
 # - install dependencies:
 #     cd $HOME/src/main/ && git pull && sudo apt install $(cat .github/workflows/apt.txt)
-# - install further dependencies (not needed for mkdocs!):
-#     apt-get install texlive-latex-extra python3-sphinxcontrib.apidoc
+# - install further dependencies (libpython and developers docs):
+#     apt-get install texlive-latex-extra python3-sphinxcontrib.apidoc python3-sphinx-sitemap
+#     pip install sphinx-last-updated-by-git sphinx_material --break-system-packages # to avoid virtenv
```
echoix: When are these dependencies updated? Because we check in CI that we have the update of the docs generator and extensions working before using functionalities that were just added.

Here, it seems locally installed. And between the branches we might not have versions that work together, if there's only one single version for all branches

utils/cronjobs_osgeo_lxd/cron_grass_preview_build_binaries.sh
```diff
@@ -235,19 +240,16 @@ sh $MAINDIR/cronjobs/compile_addons_git.sh $GMAJOR \
...
+### fetch manual from GH actions
+# fetch artifacts for main (devel)
+bash $HOME/cronjobs/gh_cli_download_artifact.sh $BRANCH
+cd $TARGETHTMLDIR && rm -rf * && unzip -q /tmp/mkdocs-site.zip
+
+#### generate developer stuff: pygrass docs + gunittest docs
+# generate pyGRASS sphinx manual (in python/grass/docs/)
+# including source code
+cd $GRASSBUILDDIR
+$MYMAKE sphinxdoclib
```
echoix: At what point it isn't present in the artifact of GitHub? Should we be generating it there?

utils/cronjobs_osgeo_lxd/cron_grass_preview_build_binaries.sh
```diff
@@ -257,17 +259,57 @@ cp -p $MAINDIR/.grass$GMAJOR/addons/logs/* $TARGETMAIN/addons/grass$GMAJOR/logs/
...
 
+############################################
+# Cloning new manual pages into grass-devel/manuals/ (following the Python manual pages concept)
+# - inject canonical URL therein to point to versioned manual page (avoiding "duplicate content" SEO punishment)
+#   see https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls
+
+TARGETHTMLDIRDEVEL=$TARGETMAIN/grass-devel/manuals/
+# clone manual pages
+rm -rf $TARGETHTMLDIRDEVEL
+rsync -a $TARGETHTMLDIR/ $TARGETHTMLDIRDEVEL/
```
echoix, line 278: Similar observations as for stable versions

utils/cronjobs_osgeo_lxd/cron_grass_current_stable_build_binaries.sh
neteler: Due to the hard cleanup a line below we here left out the (potentially dangerous) delete flag.

utils/cronjobs_osgeo_lxd/cron_grass_preview_build_binaries.sh
neteler: > When are these dependencies updated?

Just manually, on the server which is not ideal for sure.
An option could be the use of python venv which would again add another level of complexity.

utils/cronjobs_osgeo_lxd/cron_grass_preview_build_binaries.sh
neteler: Indeed sphinx was redundant in this PR, hence removed (in b90e6d3fce214f48777633a50b8862ed23e7a0ed)

utils/cronjobs_osgeo_lxd/cron_grass_current_stable_build_binaries.sh
neteler: Generally I agree. And would welcome implementation support.

utils/cronjobs_osgeo_lxd/gh_cli_download_artifact.sh
neteler: Agreed. So we need a more sophisticated use of `gh` here.

neteler: Idea: name the documentation artifact with `YYYYMMDD-<branch>-mkdocs.zip` using the **commit** date (rather than the workflow date). Then download the latest file to the server etc. That would be done in the "documentation" workflow.

utils/cronjobs_osgeo_lxd/cron_grass_preview_build_binaries.sh
neteler: Indeed sphinx was redundant in this PR, hence removed (in b90e6d3fce214f48777633a50b8862ed23e7a0ed)

neteler:
Note: since the test implementation of this PR on the server (as there is no "dry-run" environment), I discovered that this sitemap is actually not existing on the server:
https://grass.osgeo.org/grass-devel/manuals/sitemap_manuals.xml
(404)

Means: there is a naming confusion. See

https://github.com/OSGeo/grass-addons/blob/0346f46caf05e23832b10f71b44582e485327168/utils/create_manuals_sitemap.py#L36

vs

https://github.com/OSGeo/grass/blob/5bd94dabee99556bf4956e0fd933617a61676085/.github/workflows/documentation.yml#L250
`     --output "${MKDOCS_DIR}/site/sitemap.xml" \`

Which name shall be used? Unless this is fixed, the entire "devel" manual is no longer shown when using search engines like Google etc:

```text
URL is not on Google
This page is not indexed. Pages that aren't indexed can't be served on Google.

```

echoix:
Keep in mind, from what I remember of playing around last year, the generated sitemap is not perfect. 

https://github.com/OSGeo/grass/blob/5bd94dabee99556bf4956e0fd933617a61676085/python/grass/docs/conf.py#L265-L267

and 

https://github.com/OSGeo/grass/blob/5bd94dabee99556bf4956e0fd933617a61676085/python/grass/docs/conf.py#L499-L502

collide, meaning we can't configure correctly, as that last line will win.

Either the canonical rel links are correct, or the sitemap is correct.

neteler:
> OSGeo/grass@`5bd94da`/python/grass/docs/conf.py#L499-L502

How about calling it `sitemap_sphinx.xml` here to avoid a name collision? It doesn't need to be named `sitemap.xml`.

echoix:
> > OSGeo/grass@`5bd94da`/python/grass/docs/conf.py#L499-L502
> 
> How about calling it `sitemap_sphinx.xml` here to avoid a name collision? It doesn't need to be named `sitemap.xml`.

No, it’s the html_baseurl that is assigned twice

neteler:
Why is that relevant? Sorry, I don't get it. 
I'm discussing the 404 of the sitemap file on the server.

neteler:
@cwhite911 I have done simplifications. To be checked if I got it all...

cwhite911 (approved):
Merge and pray

PR #1699, opened 2026-05-27
Fix cronjob staging

ArthPatel1502:
This PR implements the atomic directory swap suggested by @echoix to prevent 404 downtime during document builds. It also wraps variables in double quotes to resolve the Super Linter failures and explicitly names the sitemap outputs to fix the 404 indexing issue.


echoix:
Why are you changing so many files? Only stage the parts needed for your PR

PR #1701, opened 2026-05-28
r.boxplot: preserve active MASK when zonal run exits with an error

### LINKED ISSUE #1696: r.boxplot: active MASK deleted when zones= is used and script exits with an error
## Summary

When `r.boxplot` is run with `zones=` while a MASK is active, the module temporarily renames `MASK` so zonal statistics can be computed without masking.

If the script exits before restoring the original MASK, the cleanup routine deletes the temporary backup map, causing the user's MASK to be permanently lost.

## Steps to reproduce

1. Create a MASK:

   ```bash
   r.mask raster=some_raster
   ```

2. Run:

   ```bash
   r.boxplot map=x zones=y
   ```

   and make the command fail partway through (for example, on a headless system without a display, or by triggering a failure in `r.stats.quantile`).

3. Check whether MASK still exists:

   ```bash
   g.list type=raster pattern=MASK
   ```

4. Result:

   `MASK` no longer exists.

## Root cause

In `main()` (~line 1292), `backup_mask` is created using `create_temporary_name()`:

```python
backup_mask = create_temporary_name("maskbackup")
```

`create_temporary_name()` automatically adds the map to the global `clean_maps` list.

The original MASK is then renamed:

```python
Module("g.rename", raster=["MASK", backup_mask])
```

The MASK restoration only happens if `main()` exits normally:

```python
if mask_present and bool(options["zones"]):
    Module("g.rename", raster=[backup_mask, "MASK"])
```

However, `cleanup()` is registered with `atexit` and runs even when the script exits due to an exception. Since `backup_mask` is still present in `clean_maps`, it gets deleted during cleanup.

As a result, the user's original MASK is lost.

## Impact

Silent and irreversible data loss. A user's existing MASK may represent significant preprocessing work and is removed without warning or recovery.

## Possible fixes

- Remove `backup_mask` from `clean_maps` after creation, since it is not disposable temporary data, or
- Move MASK restoration into `cleanup()` so it is restored before temporary maps are deleted, regardless of how the script exits.

Happy to implement either approach if preferred.

Valyrian-Code:
## Summary

Fixes #1696.

When `r.boxplot` is executed with `zones=` while a MASK is active, the module temporarily renames `MASK` so zonal statistics can run against unmasked rasters. Previously, the backup MASK name was stored in `clean_maps`, causing the `atexit`-registered `cleanup()` handler to delete it on any exit path — including crashes or uncaught exceptions during statistics computation or plotting.

This could result in silent and irreversible loss of the user’s original MASK if the module terminated unexpectedly.

## Fix

Store the backup MASK name in a dedicated module-level variable (`mask_backup_name`) instead of adding it to `clean_maps`.

`cleanup()` now restores the original MASK before removing temporary maps, ensuring the MASK survives all exit paths, including failures during zonal statistics computation.

On successful execution, the existing inline restore logic still runs as before and clears `mask_backup_name`, making the cleanup restore step a no-op in the normal case.

## Note on `MaskManager`

The issue discussion proposed using `MaskManager` (GRASS 8.5+). Testing showed this does not disable an already existing `MASK` raster. Instead, it only sets the `GRASS_MASK` environment variable to a different name, while GRASS continues applying the existing `MASK` raster internally.

As a result, the zonal statistics workflow still failed, with `r.stats.quantile` reporting:

> No cells found where both base and cover are not NULL

Because of this behavior, physically renaming the MASK remains the correct approach. The updated `cleanup()` restoration logic provides crash safety without introducing version-specific branching.

## Test Plan

Verified manually with an active MASK in GRASS 8.5.0:

* [x] Normal run with `zones=` and active MASK → completes successfully, MASK preserved
* [x] Crash during `bxp_zones` (e.g. invalid output path) → MASK restored by `cleanup()`
* [x] Run without MASK → behavior unchanged

cc @ecodiv


Valyrian-Code:
Thanks for the review. Addressed both points in `adc3d1a`.

### Changes

* Moved MASK restoration into a shared `restore_mask_backup()` helper used by both `cleanup()` and the normal restore path
* Wrapped `g.rename` in `try/except`:

  * emits `gs.warning()` with manual recovery instructions
  * never raises, so cleanup still removes temp maps
* `mask_backup_name` is now cleared only after a successful restore
* `main()` assigns `mask_backup_name` only after a successful backup rename

### Re-tested in GRASS 8.5.0

* [x] Normal `zones=` + active MASK run → MASK restored
* [x] Forced crash in `bxp_zones` → `cleanup()` restores MASK via `atexit`

cc @ecodiv

ecodiv:
Hi @Valyrian-Code , nice work, thanks! I wonder though, at exit, cleanup() calls restore_mask_backup(), which tries to rename the backup back to MASK. But that will only work if the mask created for quantile computation is removed before. What if there is a failure before that?

Valyrian-Code:
> Hi @Valyrian-Code , nice work, thanks! I wonder though, at exit, cleanup() calls restore_mask_backup(), which tries to rename the backup back to MASK. But that will only work if the mask created for quantile computation is removed before. What if there is a failure before that?

Thanks {nice catch there}. I missed that compute_outliers() creates its own MASK around lines 721 to 742 for the outlier path with zones. If anything in between raises, MASK is left behind and restore_mask_backup() fails because MASK already exists.

### Two part fix I am thinking:

1. In compute_outliers(), wrap the r.mask raster=zones ... r.mask -r block in try/finally so the temporary mask is always removed, even on error.

2. As a backstop in restore_mask_backup(), if MASK already exists when we go to restore the backup, remove it first. At cleanup time any leftover MASK is one r.boxplot created itself, so removing it is safe.

> Does that match what you had in mind? I will implement once you confirm.

cc @ecodiv

ecodiv:
@Valyrian-Code sound like a good solution

> > Hi @Valyrian-Code , nice work, thanks! I wonder though, at exit, cleanup() calls restore_mask_backup(), which tries to rename the backup back to MASK. But that will only work if the mask created for quantile computation is removed before. What if there is a failure before that?
> 
> Thanks {nice catch there}. I missed that compute_outliers() creates its own MASK around lines 721 to 742 for the outlier path with zones. If anything in between raises, MASK is left behind and restore_mask_backup() fails because MASK already exists.
> ### Two part fix I am thinking:
> 
>     1. In compute_outliers(), wrap the r.mask raster=zones ... r.mask -r block in try/finally so the temporary mask is always removed, even on error.
> 
>     2. As a backstop in restore_mask_backup(), if MASK already exists when we go to restore the backup, remove it first. At cleanup time any leftover MASK is one r.boxplot created itself, so removing it is safe.
> 
> 
> > Does that match what you had in mind? I will implement once you confirm.
> 
> cc @ecodiv

@Valyrian-Code sound like a good solution

Valyrian-Code:
Thanks @ecodiv, implemented both parts in 1c9804728:

1. Wrapped the temporary zonal mask block in `compute_outliers()` in try/finally, so the mask is always removed even if a step in between raises.
2. Added a backstop in `restore_mask_backup()`: if a MASK already exists when we go to restore the backup, remove it first. At cleanup time any leftover MASK is one r.boxplot created itself, so removing it is safe.

Tested in GRASS 8.5 with an active MASK and `zones=`: with a forced crash inside `compute_outliers()` (after the zonal mask is created), the original MASK is now restored correctly. Without the fix the same crash left the zonal mask active (wrong cell count) and the original orphaned as a `maskbackup` raster. Ready for another look whenever you have time.

ecodiv:
Looks good. Tried it out, and it works for me. Thanks. I think this can be merged when the checks are completed.

Valyrian-Code:
Thanks a lot @ecodiv, I really appreciate you testing it and for the reviews along the way. Glad it works on your end too. The checks are almost through (just the 8.5 release branch job still running), so it should be ready shortly.

PR #1715, opened 2026-06-03
r.vif: preserve active MASK when read_data() exits with an error

### LINKED ISSUE #1702: r.vif: active MASK silently deleted (and replaced with random-sample mask) when read_data() exits with an error
## Summary

When `r.vif` is run with `n=` (random sampling), `read_data()` temporarily renames the active `MASK` so it can create a new random-sampling mask. The backup name is currently registered in `CLEAN_RAST` via `tmpname()`.

If an exception occurs between the backup rename and the final restore step (lines 195–209), the `atexit`-registered `cleanup()` removes the backup raster before it can be restored. This permanently deletes the user's original `MASK`, while the temporary random-sampling mask created by `r.mask raster=new_mask` is left behind as the active `MASK`.

This is the same antipattern recently fixed in r.boxplot (issue #1696, PR #1701).

## Reproduction

GRASS 8.5.0. Inside a mapset with three correlated rasters and an active MASK:

r.mapcalc "rvif_a = rand(1.0, 100.0)" --overwrite
r.mapcalc "rvif_b = rvif_a * 2 + rand(1.0, 10.0)" --overwrite
r.mapcalc "rvif_c = rand(1.0, 50.0)" --overwrite
r.mapcalc "rvif_mask = if(row() <= 5, 1, null())" --overwrite
r.mask raster=rvif_mask

Inject a controlled failure inside the protected window of the installed `r.vif` (added one line after `r.mask raster=new_mask`):

raise RuntimeError("INJECTED FAILURE for MASK loss reproduction")

Run:

r.vif maps=rvif_a,rvif_b,rvif_c n=100 seed=42

After the crash:
- `g.list type=raster pattern="rvifoldmask*"` returns empty — the backup was deleted by `cleanup()`
- `g.list type=raster pattern=MASK` shows MASK present, but it is now the random-sample reclass, not the user's original mask

## Root cause

In `read_data()` (`src/raster/r.vif/r.vif.py:194`):

```python
mask_backup = tmpname("rvifoldmask")
gs.run_command("g.rename", raster=["MASK", mask_backup], quiet=True)
tmpname() appends to CLEAN_RAST, and cleanup() (registered with atexit) deletes everything in that list on any exit, including exceptions raised between the rename and the rename-back at line 209.

Possible fix
Same direction as #1701: track the backup name in a module-level variable instead of CLEAN_RAST, and restore it from cleanup() first (before any temp-map removals). Happy to implement once approach is confirmed.
```

cc @wenzeslaus (original 2021 commit) @ecodiv

Valyrian-Code:
## Summary

Fixes #1702.

When `r.vif` is run with `n=` (random sampling), `read_data()` renames the active `MASK` to a temporary backup so a new random-sampling mask can be applied. The backup name was previously added to `CLEAN_RAST` via `tmpname()`, so the `atexit` `cleanup()` deleted it on any exit — including exceptions raised during `r.stats` or the numpy parse. The user's original mask was permanently lost, and the random-sample reclass left behind by `r.mask raster=new_mask` was left as the active `MASK`.

This is the same antipattern fixed in r.boxplot (#1696, PR #1701), adapted to r.vif's flow.

## Fix

1. Track the backup mask name in a module-level `mask_backup_name` variable instead of adding it to `CLEAN_RAST`. Set the variable only after a successful rename.

2. Add a `restore_mask_backup()` helper that:

   - Returns silently if no backup is tracked.
   - Removes any pre-existing `MASK` first (at cleanup time any MASK present is one r.vif created internally, so this is safe).
   - Wraps the rename in `try/except`, emits `gs.warning()` with the backup name and a manual restore command on failure, and never raises.
   
3. Call `restore_mask_backup()` from `cleanup()` before the temp-map removal loop, so the user's MASK is restored on any exit path.

4. Wrap the `r.stats` / `np.loadtxt` block in `try/finally` so on the normal-exception path the internal mask is removed and the backup is restored locally.

## Test plan

Verified manually in GRASS 8.5.0 with an active MASK and three correlated rasters. Controlled failure injected after `r.mask raster=new_mask` to simulate a crash in the protected window.

Before the fix:
- `g.list pattern="rvifoldmask*"` after crash → backup raster remained but
- `g.list pattern=MASK` after crash → MASK was the random-sample reclass, not the user's mask. Original MASK was deleted by `cleanup()`.

After the fix:
- `g.list pattern=MASK` → MASK present
- `r.info MASK` → "Reclass of rvif_mask" — the user's original mask
- `g.list pattern="rvifoldmask*"` → empty (backup was renamed back, not orphaned)
- No warnings during cleanup

cc @ecodiv @wenzeslaus


Valyrian-Code:
Thanks for the review. Addressed two of the three:

1. Widened try/finally to also cover gs.run_command("r.mask", raster=new_mask, ...). Even before this change the atexit cleanup() served as a backstop, but it is cleaner to localize the restoration inline.

2. Switched restore_mask_backup() from ["file"] to ["fullname"] for consistency with the rest of the function. Both keys work in gs.find_file output (both truthy on hit), but mixing reads as sloppy.

3. Did not change the "remove pre-existing MASK before restore" behavior. GRASS mapsets are locked during operation, so concurrent MASK creation by another process is not a supported scenario. At cleanup time any MASK present is one r.vif itself created (the random-sample reclass), and removing it before renaming the backup back is the right action. Renaming it to a temp would just create leftover state for a case that cannot happen.

Re-tested manually in GRASS 8.5.0 — normal run completes with MASK preserved, crash-injection test (raise after r.mask) leaves MASK intact via atexit cleanup.

wenzeslaus:
In 8.5, we have MaskManager for this. I would consider duplicatating the code here for compatibility with 8.4.

Valyrian-Code:
> In 8.5, we have MaskManager for this. I would consider duplicatating the code here for compatibility with 8.4.

Good point. Just to confirm before refactoring: the r.vif use case is "replace MASK with a different one" (random-sample), which is what MaskManager handles cleanly via manager.mask_name and GRASS_MASK. (This is different from r.boxplot's "disable the existing MASK" case where MaskManager alone is not sufficient see #1701.)

Plan:
  - On 8.5+: write the random-sample raster directly to manager.mask_name inside a MaskManager context. No manual rename, no try/finally needed locally, MaskManager handles cleanup.
  
  - On 8.4: keep the current manual approach (g.rename + restore_mask_backup() backstop in cleanup).
  
  - Version detection via try/import on grass.script.MaskManager.

Does that match what you had in mind?

Valyrian-Code:
Thanks @wenzeslaus. Implemented the dual path in the latest commit:

- On 8.5+ (MaskManager available): the random-sample mask is applied inside a `MaskManager` context, which restores the user's MASK automatically on exit including on error, so the manual backup and rename are not needed there.
- On 8.4: the previous manual approach (g.rename backup plus restore in cleanup) is kept, selected by a `try`-import of `grass.script.MaskManager`.
- The sample is drawn before entering the MaskManager context, so it still respects an active user MASK, matching the 8.4 behaviour.

Verified in GRASS 8.5 with a synthetic three-raster case: both paths give identical VIF on the same seed, the user's MASK is preserved after a normal run and after a forced crash inside the sampling block, and no internal mask is left behind. Ready for another look whenever you have time.

ecodiv:
Thanks @Valyrian-Code , tested it, works, merging.

PR #1727, opened 2026-06-06
r.cell.area: fix silent exit on unsupported projection units; support all linear CRS; add testsuite

### LINKED ISSUE #1716: r.cell.area: silently exits with success on unsupported projection units
## Summary

When the current location's projection units are anything other than "meters"/"Meters" or "degrees"/"Degrees", `r.cell.area` prints a message via raw `print()` and exits with status 0, producing no output raster. A wrapping script or pipeline cannot detect the failure — `r.cell.area output=foo units=m2` reports success, but `foo` does not exist.

## Affected code

`src/raster/r.cell.area/r.cell.area.py:100-101`

```python
else:
    print("Units: ", projunits, " not currently supported")
```

The supported branches above are explicit:

- line 78: `(projunits == "meters") or (projunits == "Meters")`
- line 83: `(projunits == "degrees") or (projunits == "Degrees")`

Any other value (e.g., "feet", "us-ft") falls through to the silent-failure branch.

## Reproduction (predicted from reading; not yet reproduced live)

In a location whose CRS reports feet as the unit (e.g., EPSG:6446 — NAD83 Massachusetts state plane, US Survey Foot):

```
r.cell.area output=cell_area units=m2
echo $?
g.list type=raster pattern=cell_area
```

Expected (after fix): non-zero exit code, error on stderr, no map created.
Actual (current): exit code 0, message on stdout, no map created.

## Fix

Replace `print(...)` at line 101 with `gs.fatal(...)` so the script exits with non-zero status and the error goes through GRASS's message infrastructure.

## Related concerns (lower priority)

- Line 76 (`projunits = str(projinfo["units"])`) can raise `KeyError` if `projinfo` doesn't contain a `units` key (unusual CRSes). Worth a defensive `.get("units", "")` in the same fix.
- This module lacks a testsuite. A small testsuite covering the meters, degrees, and unsupported-units paths would be a useful follow-up.

Happy to submit a PR once approach is confirmed.

cc @awickert (original author) @wenzeslaus (introduced the print line in 2021)

awickert:
## Summary

Fixes #1716.

- **Replace `print()` with `gs.fatal()`** for unsupported projection units, so pipelines and wrapper scripts can detect failure via a non-zero exit code and the message goes through GRASS's message infrastructure.
- **Generalize projected-CRS support** to all linear units (feet, US survey feet, nautical miles, etc.) by using the `meters` conversion factor from `g.proj -g` instead of string-matching on `"meters"`/`"Meters"` only. This is the same approach used by `r.in.usgs` and `r.futures.calib`.
- **Defensive `.get()` on the `units` key** to avoid a `KeyError` on unusual CRSes that don't report a units key.
- **Remove four unused imports** (`os`, `glob`, `garray`, `VectorTopo`).
- **Add a testsuite** covering: projected CRS m² and km² values with known cell sizes, overwrite protection, geographic CRS spherical formula, US survey feet CRS via the meters conversion factor, and XY (unprojected) location fatal error.

## Test plan

- [ ] `TestRCellAreaMeters` — runs in the NC sample dataset; verifies 10 m × 10 m cells → 100 m², 1000 m cells → 1 km², overwrite protection, overwrite flag
- [ ] `TestRCellAreaDegrees` — spawns a temporary EPSG:4326 project; verifies spherical formula and km²/m² consistency
- [ ] `TestRCellAreaFeet` — spawns a temporary EPSG:2249 (Massachusetts State Plane, US survey feet) project; verifies 100 ft × 100 ft cell ≈ 929.034 m²
- [ ] `TestRCellAreaXY` — spawns a temporary XY project; verifies non-zero exit code

### DIFF
--- a/src/raster/r.cell.area/r.cell.area.py
+++ b/src/raster/r.cell.area/r.cell.area.py
@@ -93,18 +93,12 @@ def main():
         # that feet, US survey feet, and all other linear units work correctly.
         m = float(projinfo.get("meters", 0))
         if m <= 0:
-            gs.fatal(
-                _("Projection units '{}' are not supported".format(projunits))
-            )
+            gs.fatal(_("Projection units '%s' are not supported") % projunits)
         if units == "m2":
-            gs.mapcalc(
-                "{out} = nsres() * ewres() * {m2}".format(out=output, m2=m**2)
-            )
+            gs.mapcalc("{out} = nsres() * ewres() * {m2}".format(out=output, m2=m**2))
         elif units == "km2":
             gs.mapcalc(
-                "{out} = nsres() * ewres() * {m2} / 1.e6".format(
-                    out=output, m2=m**2
-                )
+                "{out} = nsres() * ewres() * {m2} / 1.e6".format(out=output, m2=m**2)
             )
 
 
--- a/src/raster/r.cell.area/r.cell.area.py
+++ b/src/raster/r.cell.area/r.cell.area.py
@@ -43,14 +43,10 @@
 ##################
 
 # PYTHON
-import os
-import glob
 import numpy as np
 
 # GRASS
 import grass.script as gs
-from grass.script import array as garray
-from grass.pygrass.vector import VectorTopo
 
 
 def main():
--- a/src/raster/r.cell.area/r.cell.area.py
+++ b/src/raster/r.cell.area/r.cell.area.py
@@ -73,7 +73,7 @@ def main():
                 + "' already exists. Use '--o' to overwrite."
             )
 
-    projunits = str(projinfo["units"])  # Unicode to str
+    projunits = str(projinfo.get("units", ""))
     # Then compute
     if (projunits == "meters") or (projunits == "Meters"):
         if units == "m2":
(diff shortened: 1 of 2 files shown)


echoix:
@awickert You might want to update the settings of your fork to have grass8 the default branch, and update that grass8 branch on your fork. So the commits would not conflict as much.

currently, even if I would approve, the conflicts prevent me from being able to merge.
The conflicts seem to show that there are some changes in grass7 folder, what existed several years ago, and we can’t merge

awickert:
@echoix Fortunately the problem was simpler: I simply had left the default "OSGeo:master" for the target branch. I just switched that. Hopefully it looks good to bring in now.

echoix:
> @echoix Fortunately the problem was simpler: I simply had left the default "OSGeo:master" for the target branch. I just switched that. Hopefully it looks good to bring in now.

Oh, forgot to see that. I might have caught it if there weren’t conflicts, as the CI checks wouldn’t have been the same (and likely all fail to run on older discontinued runners). But I saw that your default branch is « master », and that is problematic for you in some way, anyway

awickert:
"master" might well have been the default when I first forked grass-addons :D

Let's hope that the checks work...

echoix:
> "master" might well have been the default when I first forked grass-addons :D
> 
> Let's hope that the checks work...

It was a while ago, and was causing confusion, so it got reorganized a big while ago

echoix:
Updating the pr to get ci to run (nothing at all was ran yet)

echoix (approved):

PR #1697, opened 2026-05-25
r.boxplot: remove debug print statement and add testsuite

Valyrian-Code:
## Summary

- Remove a leftover `print(vect_name)` debug statement in `bxp_nozones()` that leaked an internal Python list directly to stdout, bypassing normal GRASS output handling.
- Add a testsuite covering:
  - basic output
  - zonal output
  - horizontal orientation
  - missing required input
  - rejection of floating-point zones

## Test plan

- [ ] Verify all tests pass:

  ```bash
  python src/raster/r.boxplot/testsuite/test_r_boxplot.py
  ```

- [ ] Verify pre-commit checks pass on modified files

## Notes

A separate issue (#1696) tracks MASK loss when `zones=` is used and the script exits with an error. That issue is unrelated to this PR.

PR #1729, opened 2026-06-08
r.boxplot: remove debug print statement and add testsuite

Valyrian-Code:
The previous PR (#1697) was accidentally opened from my fork's default branch, which caused problems. This re-submits the same change from a dedicated branch, as suggested there.

It removes a leftover debug `print(vect_name)` in `bxp_nozones` (which leaked an internal Python list to stdout) and adds a testsuite covering basic, zonal, and flag cases for r.boxplot.


ecodiv:
@Valyrian-Code this can be merged I think

echoix (approved):

PR #1735, opened 2026-06-08
t.strds.copy: new addon

ecodiv:
Addon to copy a STRDS and its registered maps  from a source mapset to the current mapset.

### DIFF
--- a/src/temporal/t.strds.copy/Makefile
+++ b/src/temporal/t.strds.copy/Makefile
@@ -1,0 +1,7 @@
+MODULE_TOPDIR = ../..
+
+PGM = t.strds.copy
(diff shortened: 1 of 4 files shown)


veroandreo:
How is this new add-on different from `t.copy` in core? t.copy also allows you to copy the maps together with the stds and can copy over mapsets as long as they are accessible.

ecodiv:
That is a very good question. I could not find t.copy, it does not show up in my menu or tools. But it is there, I can indeed open it from the command line or console. Thanks, will remove this one.

ecodiv:
replicates existing funtionality, closing

PR #1737, opened 2026-06-08
t.rast.line: bugfix handle with timestamps with fractional seconds

ecodiv:
The module failed when registered raster maps had timestamps that included fractional seconds. Fixed

### DIFF
--- a/src/temporal/t.rast.line/t.rast.line.py
+++ b/src/temporal/t.rast.line/t.rast.line.py
@@ -208,20 +208,9 @@ import atexit
 import os
 import sys
 from datetime import datetime
-
-
-def _parse_dt(value):
-    """Parse a datetime string, tolerating optional fractional seconds."""
-    for fmt in ("%Y-%m-%d %H:%M:%S.%f", "%Y-%m-%d %H:%M:%S"):
-        try:
-            return datetime.strptime(value, fmt)
-        except ValueError:
-            continue
-    raise ValueError("Unrecognized datetime format: %s" % value)
-
-
 import grass.script as gs
 from grass.exceptions import CalledModuleError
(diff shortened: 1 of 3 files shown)


src/temporal/t.rast.line/t.rast.line.py
```diff
@@ -208,6 +208,18 @@
...
+
+
+def _parse_dt(value):
+    """Parse a datetime string, tolerating optional fractional seconds."""
+    for fmt in ("%Y-%m-%d %H:%M:%S.%f", "%Y-%m-%d %H:%M:%S"):
+        try:
+            return datetime.strptime(value, fmt)
+        except ValueError:
+            continue
+    raise ValueError("Unrecognized datetime format: %s" % value)
```
echoix: Don't we have a function like that in the grass code? I think I already saw that somewhere.

ecodiv: I think you refer to grass.temporal.datetime_math.string_to_datetime()? That is something I would never have found myself, thanks. Changed it.

echoix: Oh, you're better than me and found it. And ya, it was in that file. Does it do what you want it to do?

ecodiv: I think so, it was about dealing with fractional seconds. The function uses, if installed,the datutil library. I tested it, and it works. But also if datutil is not installed, it should support fractional seconds, if I understand it correctly.

```
In case datutil is not installed the supported ISO string formats are:
    - YYYY-mm-dd HH:MM:SS.s
    - YYYY-mm-ddTHH:MM:SS.s
```

echoix (approved):

Issue #1741, opened 2026-06-09
Several addons change the computational region without restoring it (use_temp_region) — tracking issue


Valyrian-Code:
## Summary

Several Python addons change the computational region (`g.region` with extent or resolution parameters) without using `gs.use_temp_region()`, so the user's region is left changed after the module runs, and in some cases is corrupted when the module exits with an error. As suggested in #1732, this consolidates the individual reports into a single tracking issue instead of one issue per module.

The fix is the same in each case: scope the region change with `gs.use_temp_region()` (and `del_temp_region`), which restores the user's original region automatically on exit, including on error. This is the standard GRASS pattern for modules that set their own region.

## Affected addons found so far

Region never restored (changed on every run):
- [ ] r.forcircular
- [ ] r.soillossbare
- [ ] i.image.bathymetry

Region not restored on error (restored only on the happy path):
- [ ] r.earthworks
- [ ] r.confusionmatrix
- [ ] v.habitat.dem

Conditional, only with the `-e` flag and anisotropic resolution:
- [x] r.maxent.predict
- [x] v.maxent.swd

(Related, but a different cleanup-on-error pattern: active MASK loss, tracked separately in #1730, already fixed in r.boxplot and r.vif.)

## Plan

Per the discussion in #1732: fix one good example first (r.earthworks, whose author is active), get the `use_temp_region()` approach reviewed there, and then work through the rest at a reasonable pace, coordinating with each addon's author. I will keep this list updated as items are fixed.

ecodiv:
Thanks @Valyrian-Code I implemented your suggestion for v.maxent.swd and r.maxent.predict in #1825.

PR #1742, opened 2026-06-09
r.soillossbare: restore the user's region with a temporary region

Valyrian-Code:
Part of #1741.

r.soillossbare sets the computational region to its own working resolution (`g.region -a res=...`) as the first thing it does, but it never restores it, so after the module runs the user is left with their region silently changed.

This wraps the change in `gs.use_temp_region()`, which restores the user's original region automatically on exit (including on error). This is the standard GRASS pattern for modules that set their own region.

I also added a testsuite. It fails without the fix (the region is left at the module's resolution) and passes with it. Verified locally on GRASS 8.5.

This is the first example from the #1741 cleanup. Happy to work through the rest at whatever pace suits, coordinating with each addon's author.


ecodiv (approved):
Good change, and test set works as advertised (fails without change, passes with change)

ecodiv:
@martinzbinden agree to merge?

Valyrian-Code:
> Good change, and test set works as advertised (fails without change, passes with change)

Thanks ✨

Valyrian-Code:
Gentle ping on this one. The included test fails without the change and passes with it, and it is the same use_temp_region pattern already merged for r.vif and r.boxplot. Happy to rebase if it has gone stale. Anything else needed before it can go in?

echoix (approved):

PR #1743, opened 2026-06-09
r.earthworks: use a temporary region to restore the user's region on exit

Valyrian-Code:
Part of #1741. Same `use_temp_region()` pattern as #1742 (r.soillossbare).

In `main()`, r.earthworks saves the current region as `source`, changes the region while modeling each quadrant, and restores it in `series()`. On a normal run that works fine. But the `finally` block only registers cleanup of the temporary maps, it does not restore the region, so on an error between the region change and the `series()` restore the user can be left with a changed region (and the saved `source` region gets removed by the cleanup too).

Wrapping the region handling in `gs.use_temp_region()` at the start of `main()` restores the user's original region automatically on exit, including on error. It is additive: the existing `source` save/restore still drives the per-quadrant and cumulative extents.

In practice `series()` restores on the normal path, so this mainly hardens the error path. @baharmon, since this is your module, does this look reasonable to you? Happy to adjust.


ecodiv:
Looks good to me. @baharmon , can this be merged?

ecodiv (approved):
Looks good

baharmon:
> Looks good to me. @baharmon , can this be merged?

@ecodiv Yes, tested and ready to be merged.

src/raster/r.earthworks/r.earthworks.py
```diff
@@ -1249,6 +1249,10 @@ def main():
 
     # Run processes
     try:
+        # Use a temporary region so the user's computational region is
```
baharmon, line 1252: Consider shortening comment to just "Use temporary region" to match style

Change
```
# Use a temporary region so the user's computational region is
# restored on exit, including when the module exits with an error
```
To
```
# Use temporary region
```

baharmon (changes requested):

baharmon:
I found and fixed a bug related to quadtree segmentation in r.earthworks with PR #1827. With the bug fix, I improved error handling and cleanup which required considerable refactoring. I added the temporary region, but removed the try finally statement, so the code has diverged quite a bit from this PR #1743. I recommend closing #1743 in favor of #1827.

Valyrian-Code:
Thanks @baharmon, that sounds great, and nice catch on the quadtree bug. Glad the temporary region fit into your refactor. Closing this in favor of #1827.

PR #1746, opened 2026-06-10
r.skyview: add parallel processing and memory option

ecodiv:
Add nprocs and memory options.
Only used for version 8.5 and up.


src/raster/r.skyview/r.skyview.py
```diff
@@ -120,6 +124,14 @@ def main():
     elev = options["input"]
     output = options["output"]
     n_dir = int(options["ndir"])
+    nprocs = int(options["nprocs"])
+    memory = int(options["memory"])
+    version = gcore.version()["version"].split(".")
+    major = int(version[0])
+    minor = int("".join(c for c in version[1] if c.isdigit()))
```
petrasovaa: This line looks unnecessarily complicated no?

src/raster/r.skyview/r.skyview.py
```diff
@@ -120,6 +124,14 @@ def main():
...
+    if (major, minor) < (8, 5):
+        nprocs = None
+        memory = None
```
petrasovaa, line 134: I guess a warning might be useful to notify user parallelization is not used.

src/raster/r.skyview/r.skyview.py
```diff
@@ -218,12 +235,32 @@ def main():
         )
         return 1
     if colorized_output:
+        sa_params = {}
+        if nprocs is not None:
+            sa_params["nprocs"] = nprocs
+        if memory is not None:
```
petrasovaa: nprocs and memory go together, so it's enough to test for one

src/raster/r.skyview/r.skyview.py
ecodiv: A number of addons use things like
```
gs.version()["version"] >= "8.2.0":
    r_patch_kwargs = {"nprocs": nprocs, "memory": memory}
```

But that would not work with 8.5dev I think? and it will not work for e.g., "10.0.0" >= "8.2.0"

petrasovaa (approved):

PR #1747, opened 2026-06-10
v.flexure: fix crash and add fatal for empty input vector map

awickert:
## Summary

Follow-up to #1745. Two bugs in the zero-load-point edge case:

- `get_points_xy` crashed with `IndexError` when passed a vector map containing no points: `np.array([], dtype=float)` has shape `(0,)`, so `coords[:, 0]` raises `IndexError: too many indices for array`. Fixed by returning empty 1-D arrays directly when there are no rows.
- A zero-point input would have silently produced all-zero deflection (gFlex treats it as no load, which is mathematically correct but almost certainly a user error). Added a `gs.fatal` immediately after coordinate loading so the user gets a clear message rather than a flat raster with no warning.

Note: the summary in #1745 incorrectly claimed "validates minimum of 2 load points" — that check was never implemented, and the claim was a hallucination in the AI-generated PR description. The minimum is 1 point (a single-point load works correctly and is the primary test fixture). This PR adds the zero-point guard only.

## Test plan

- [ ] `v.flexure` testsuite: 15/15 tests pass (14 from #1745 + `test_empty_input` added here)

### DIFF
--- a/src/vector/v.flexure/testsuite/test_v_flexure.py
+++ b/src/vector/v.flexure/testsuite/test_v_flexure.py
@@ -579,6 +579,7 @@ class TestVFlexureInputErrors(TestCase):
     """Input-validation errors raised inside main() — gFlex must be importable."""
 
     loads = "test_vflex_ierr_loads"
+    empty_loads = "test_vflex_ierr_empty"
 
     @classmethod
     def setUpClass(cls):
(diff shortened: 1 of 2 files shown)


echoix:
@awickert The previous commit (at least 3 months ago) had CI failures. I refreshed it to be able to see if it still fails. Would you be able to take a look at it?

echoix (approved):

PR #1758, opened 2026-06-15
r.curvenumber: fix missing NLCD cat 23 (Medium intensity developed)

petrasovaa:
NLCD category 23 (medium intensity developed) was missing from the lookup table, resulting in nulls. This completes the table with values based on table 9-5 and eq. 9-1 in https://irrigationtoolbox.com/NEH/Part630_Hydrology/H_210_630_09.pdf

The values for fair condition also match values in an external project:
https://github.com/WikiWatershed/tr-55/blob/develop/tr55/tables.py

Attached AI-generated script computing the values:
[cn.py](https://github.com/user-attachments/files/28965281/cn.py)

The script also shows inconsistencies in the other developed classes, perhaps due to different methodology?
<details>
================================================================
NEW: NLCD class 23 (Developed, Medium Intensity)  P_imp=65%
================================================================
  POOR  A=88 B=91 C=94 D=95
  FAIR  A=81 B=88 C=91 D=93
  GOOD  A=77 B=85 C=90 D=92

lut.c records to insert (between class 22 and class 24):

    {23, 1, HC_POOR, 88},  {23, 2, HC_POOR, 91},  {23, 3, HC_POOR, 94},
    {23, 4, HC_POOR, 95},  {23, 1, HC_FAIR, 81},  {23, 2, HC_FAIR, 88},
    {23, 3, HC_FAIR, 91},  {23, 4, HC_FAIR, 93},  {23, 1, HC_GOOD, 77},
    {23, 2, HC_GOOD, 85},  {23, 3, HC_GOOD, 90},  {23, 4, HC_GOOD, 92},

Method vs. existing lut.c (formula = open-space + P_imp via Eq. 9-1):

  class 21  (P_imp=0%)
    POOR formula=[68, 79, 86, 89] existing=[68, 79, 86, 89] exact
    FAIR formula=[49, 69, 79, 84] existing=[49, 69, 79, 84] exact
    GOOD formula=[39, 61, 74, 80] existing=[39, 61, 74, 80] exact

  class 22  (P_imp=38%)
    POOR formula=[79, 86, 91, 92] existing=[74, 86, 90, 92] diff [5, 0, 1, 0]
    FAIR formula=[68, 80, 86, 89] existing=[64, 79, 86, 89] diff [4, 1, 0, 0]
    GOOD formula=[61, 75, 83, 87] existing=[55, 74, 82, 86] diff [6, 1, 1, 1]

  class 24  (P_imp=85%)
    POOR formula=[94, 95, 96, 97] existing=[98, 98, 98, 98] diff [-4, -3, -2, -1]
    FAIR formula=[91, 94, 95, 96] existing=[97, 97, 97, 97] diff [-6, -3, -2, -1]
    GOOD formula=[89, 92, 94, 95] existing=[95, 95, 95, 95] diff [-6, -3, -1, 0]
</details>

@mabdazzam, any idea about the discrepancies? They are fairly minor, so I am not too worried about it, but I would be curious.


mabdazzam:
@petrasovaa Thanks for catching this. The missing class 23 values were likely just an oversight when I converted the original Python implementation (#1403) into C. 

use of Eq. 9-1 is probably more justified. In the original implementation, I used the Table 9-5 values directly because there was some concern that assuming the median impervious percentage is not always ideal. Also, for paved/impervious surfaces, the usual practice is often to stay on the higher/conservative side.

Secondly, the main idea was to provide baseline CN values and the range across the different condition classes, so any value within that range is still defensible. That is why we did not go too far in trying to calculate the most “accurate” initial value. The current picks are justified from Table 9-5 , but I agree that the equation backed approach is cleaner here.
A CN difference of around 6 in some parts of a watershed is unlikely to significantly affect runoff estimates, so my understanding is that either value is fine (and also for every class, a certain  `P_imp` instead of mean would make the difference 0).

PR #1769, opened 2026-06-30
tests: Add pytest in CI

wenzeslaus:
Run pytest-based addon tests in CI, in addition to the existing
gunittest runs. A new CI step discovers each addon's tests/ directory
and runs pytest per addon in a separate process for isolation. When an
addon ships a requirements.txt, it is installed first, and tools that
need network access (e.g. r.estimap.recreation) are skipped. A directory
with no collected tests is not treated as a failure.

Add the pytest configuration (import mode, test-file patterns, markers,
and a per-test timeout) to pyproject.toml, and add pytest and
pytest-timeout to the CI requirements.

This is one step toward running pytest for addons; fixes for individual
tools follow in separate pull requests.

The CI configuration was generated with the help of Claude and tested
locally.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -149,7 +149,8 @@ jobs:
           # Addons that bundle a Python package install it under etc/<tool>/;
           # add those dirs so their tests can import the package.
           etc_paths="$(find "$GISBASE/etc" -maxdepth 1 -type d -name '*.*' | paste -sd: -)"
-          export PYTHONPATH="$(grass --config python_path):${etc_paths}:${PYTHONPATH}"
+          PYTHONPATH="$(grass --config python_path):${etc_paths}:${PYTHONPATH}"
+          export PYTHONPATH
           export LD_LIBRARY_PATH="${GISBASE}/lib:${LD_LIBRARY_PATH}"
           status=0
           while IFS= read -r tests_dir; do
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -141,6 +141,35 @@ jobs:
           cd grass-addons/src
           ../.github/workflows/test.sh --config ../.gunittest.cfg
 
+      - name: Run pytest tests
+        timeout-minutes: 20
+        run: |
+          cd grass-addons/src
+          GISBASE="$(grass --config path)"
+          # Addons that bundle a Python package install it under etc/<tool>/;
+          # add those dirs so their tests can import the package.
+          etc_paths="$(find "$GISBASE/etc" -maxdepth 1 -type d -name '*.*' | paste -sd: -)"
+          export PYTHONPATH="$(grass --config python_path):${etc_paths}:${PYTHONPATH}"
+          export LD_LIBRARY_PATH="${GISBASE}/lib:${LD_LIBRARY_PATH}"
+          status=0
+          while IFS= read -r tests_dir; do
+            addon="$(dirname "$tests_dir")"
+            # r.estimap.recreation uses downloads data
+            # over the network, so we skip it for CI.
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -141,6 +141,36 @@ jobs:
...
+          cd grass-addons/src
+          GISBASE="$(grass --config path)"
+          # Addons that bundle a Python package install it under etc/<tool>/;
+          # add those dirs so their tests can import the package.
+          etc_paths="$(find "$GISBASE/etc" -maxdepth 1 -type d -name '*.*' | paste -sd: -)"
+          PYTHONPATH="$(grass --config python_path):${etc_paths}:${PYTHONPATH}"
+          export PYTHONPATH
+          export LD_LIBRARY_PATH="${GISBASE}/lib:${LD_LIBRARY_PATH}"
+          status=0
+          while IFS= read -r tests_dir; do
```
echoix, line 156: Where is `tests_dir` defined? Or how is the list of directories for iterating through the addons is collected?

.github/workflows/ci.yml
```diff
@@ -141,6 +141,36 @@ jobs:
...
+            addon="$(dirname "$tests_dir")"
+            # r.estimap.recreation uses downloads data
+            # over the network, so we skip it for CI.
```
echoix, line 159: TODO: Add a task item for that

.github/workflows/ci.yml
```diff
@@ -141,6 +141,36 @@ jobs:
...
+            echo "::group::pytest $addon"
+            # For increased isolation, run each addon in its own process.
+            rc=0
+            pytest --verbose --color=yes -ra "$tests_dir" || rc=$?
+            echo "::endgroup::"
+            # rc 5 means no tests were collected; only other codes are failures.
+            if [ "$rc" -ne 0 ] && [ "$rc" -ne 5 ]; then status=1; fi
+          done < <(find . -type d -name tests | sort)
+          exit "$status"
+
```
echoix, line 173: How do we distinguish between an addon that has no test (pytest) vs an addon that thinks they have some pytests, but due to naming conventions, they have none collected? (Oh, but even then, we couldn't catch the case where existing pytest tests are present+working and then a new test file is added without respecting the naming conventions and would not collect these new tests and we wouldn't see that it didn't work.

pyproject.toml
```diff
@@ -2,6 +2,24 @@
...
+minversion = "8.2"
+python_files = """
+    */tests/*_test.py
+    */tests/test_*.py
+"""
+markers = [
+    "slow: marks tests as slow (deselect with '-m \"not slow\"')",
+    "needs_solo_run: marks tests that must be run without any other tests running in parallel",
+]
+timeout = 300
```
echoix, line 21: TODO (future): Reduce that timeout, a test shouldn't take 5 minutes by its own.

echoix (approved):
If none of these are useful, lets start with this and fix after

pyproject.toml
wenzeslaus: See #1789

.github/workflows/ci.yml
wenzeslaus: See #1789

.github/workflows/ci.yml
wenzeslaus: `read` here is a shell build-in, so it can modify the tests_dir variable (that's the confusing part to me, but it is true). The variable `tests_dir` here is assigned the results of `done < <(find . -type d -name tests | sort)` line by line.

While by running pytest from the top directory, we can avoid this strangeness, for collecting requirements.txt files, we will need to do something like this (in some language).

PR #1770, opened 2026-06-30
v.rast.move: Fix import of exception

wenzeslaus:
The import of CalledModuleError was wrong, this changes it to the standard import pattern (CalledModuleError is not part of grass.script, but of grass.exceptions).

### DIFF
--- a/src/vector/v.rast.move/tests/test_v_rast_move.py
+++ b/src/vector/v.rast.move/tests/test_v_rast_move.py
@@ -3,6 +3,7 @@
 import pytest
 
 import grass.script as gs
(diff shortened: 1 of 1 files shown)


echoix (commented):
Was it in grass.exceptions in 8.4?

wenzeslaus:
> Was it in grass.exceptions in 8.4?

Yes, there might have been some compatibility thing this was relying on or it was always wrong.

On the releasebrach_8_4:

```text
python/grass/exceptions/__init__.py:58:class CalledModuleError(subprocess.CalledProcessError):
```

PR #1774, opened 2026-07-01
i.lmf: fix compiler errors (-Wreturn-mismatch) and warnings

neteler:
Fixed warnings, removed unused variables:

- `-Wreturn-mismatch`: `fitting` (`fitting.c:85`), `invert_matrix` (`invert_matrix.c:123`) — bare return in non-void function → changed all 5 functions to `void`
- `-Wabsolute-value`: `invert_matrix.c:43` - `abs()` on `double `→ `fabs()`
- `-Wunused-variable` / `-Wunused-but-set-variable`: removed 15 unused variables across `lmf.c`, `main.c`, `make_matrix.c`


nilason:
It seems to me changing the return type to void could be an alternative.

neteler:
- changed to use void
- fixed various compile warnings

Fixes assisted by opencode (DeepSeek V4 Flash Free, via OpenCode Zen)

nilason (approved):
Looks good to me (didn't test). I assume "unused variables" or similar was among the warnings. Would be good to at least list the warnings addressed in the commit list.

PR #1779, opened 2026-07-08
i.eb.hsebal95: fix numerical instability

YannChemin:
Known numerical limitation of Monin-Obukhov similarity theory at near-calm wind: operational SEBAL/METRIC implementations universally floor 2m wind speed at some minimum (commonly ~1.0 m/s) specifically to avoid this singularity, since the log-wind-profile assumptions themselves break down physically at near-zero wind.


echoix:
Is it something to write in the manual page too?

echoix (approved):

PR #1784, opened 2026-07-13
r.clip: Add pytest tests

christianbraun:
Adds a pytest suite for the **r.clip** addon, which had none.

Uses `grass.tools.Tools`, targeting GRASS 8.5. A module-scoped fixture builds the input raster once per CRS; each test gets an isolated mapset via `grass.experimental.TemporaryMapsetSession`.

**7 cases:** default clip preserves input cell size/alignment; `-r` resamples to region resolution; clipped values match the source; overwrite protection; region larger than input yields nulls outside; misaligned region snaps to the input grid; works in projected (UTM) and geographic (lat/lon) CRS.


src/raster/r.clip/tests/conftest.py
```diff
@@ -0,0 +1,47 @@
...
+
+def dataset_session(tmp_path, name, *, epsg, extent, res):
+    """Create a project with input raster ``name`` and yield a handle to it.
+
+    The raster is a 10x10 grid of ``col()`` values in PERMANENT; each test runs
+    in its own temporary mapset, reached through the yielded ``env``.
+    """
+    project = tmp_path / "project"
+    gs.create_project(project, epsg=epsg)
+    with gs.setup.init(project):
```
wenzeslaus: This should use env=os.environ.copy() to ensure that it will be isolated. This should work at least for 8.5. (But maybe 8.4 too)

src/raster/r.clip/tests/conftest.py
```diff
@@ -0,0 +1,47 @@
...
+        gs.run_command("g.region", res=res, **extent)
+        gs.mapcalc(f"{name} = col()")
+        with TemporaryMapsetSession() as session:
```
wenzeslaus: With fixture scope == module this will be even more powerful.

wenzeslaus (changes requested):
Moving to 8.5 as the minimum may make this little smoother. 8.5 is tested in CI, but not 8.4.

src/raster/r.clip/tests/r_clip_test.py
```diff
@@ -0,0 +1,88 @@
...
+
+def test_r_flag_resamples_to_region_resolution(clip_ll):
+    """With -r the output takes the current region resolution."""
+    tools = clip_ll.tools
+    tools.g_region(n=5, s=0, e=5, w=0, res=0.5)
+    tools.r_clip(input=clip_ll.input, output="clip_resample", flags="r")
+
+    info = tools.r_info(map="clip_resample", format="json")
+    assert info["nsres"] == pytest.approx(0.5)
+    assert info["ewres"] == pytest.approx(0.5)
```
wenzeslaus: It would be good to add here a tolerance value for clarity.

wenzeslaus: If it is actually needed or not, I don't know, but I'm fine with having the approx here.

src/raster/r.clip/tests/r_clip_test.py
```diff
@@ -0,0 +1,88 @@
...
+def test_clipped_values_match_input(clip_ll):
+    """Clipped cells equal the source values."""
+    tools = clip_ll.tools
+    tools.g_region(n=5, s=0, e=5, w=0, res=clip_ll.res)
+    tools.r_clip(input=clip_ll.input, output="clip_values")
+
+    tools.r_mapcalc(expression=f"clip_diff = {clip_ll.input} - clip_values")
+    stats = tools.r_univar(map="clip_diff", format="json")
+    assert stats["min"] == 0
+    assert stats["max"] == 0
```
wenzeslaus: Make the one line trick here as well so that we see both values during the failure. Also a note (comment perhaps) which says this asserts the difference with existing values would help (I was first confused why both min and max are zero and then I looked one the line above that).

wenzeslaus (changes requested):
This is almost ready to be merged.

You can consider a follow up PR which would change the behavior tested by test_region_larger_than_input_produces_nulls_outside. The clip now extends the extent instead of using overlap, so that's unexpected. This is an addon, so we can change it any time.

wenzeslaus (approved):
This is great. Thank you!

PR #1786, opened 2026-07-14
r.centroids: Add pytest suite

christianbraun:
Adds a pytest suite under `tests/` for r.centroids.

Covers: centroid count per clump (background excluded), removal of r.volume's extra columns, overwrite protection, and centroid location within its source clump extent.

### DIFF
--- a/src/raster/r.centroids/tests/conftest.py
+++ b/src/raster/r.centroids/tests/conftest.py
@@ -34,7 +34,7 @@ def clumps_session(tmp_path_factory):
 
 
 @pytest.fixture
-def centroids(clumps_session):
+def clump_setup(clumps_session):
     """Isolated per-test mapset + Tools handle over the module-scoped clumps raster."""
     with TemporaryMapsetSession(env=clumps_session.env) as session:
         with Tools(session=session) as tools:
--- a/src/raster/r.centroids/tests/conftest.py
+++ b/src/raster/r.centroids/tests/conftest.py
@@ -1,0 +1,44 @@
+"""Fixtures for the r.centroids tests."""
+
+import os
+from types import SimpleNamespace
+
+import pytest
+
+import grass.script as gs
+from grass.experimental import TemporaryMapsetSession
+from grass.tools import Tools
+
+
+@pytest.fixture(scope="module")
+def clumps_session(tmp_path_factory):
+    """Project with a raster of 4 separated same-valued blocks in PERMANENT.
+
+    Region is 10x10 at resolution 1; the blocks sit in the four corners
+    with background 0 between them, so each stays a single clump.
+    """
+    project = tmp_path_factory.mktemp("r_centroids") / "project"
+    gs.create_project(project, epsg="4326")
+    with gs.setup.init(project, env=os.environ.copy()) as session:
+        with Tools(session=session) as tools:
+            tools.g_region(n=10, s=0, e=10, w=0, res=1)
+            tools.r_mapcalc(
+                expression=(
+                    "clumps = if(row() <= 3 && col() <= 3, 1,"
+                    " if(row() <= 3 && col() >= 8, 2,"
+                    " if(row() >= 8 && col() <= 3, 3,"
+                    " if(row() >= 8 && col() >= 8, 4, 0))))"
+                )
+            )
+        yield session
+
+
+@pytest.fixture
+def centroids(clumps_session):
+    """Isolated per-test mapset + Tools handle over the module-scoped clumps raster."""
+    with TemporaryMapsetSession(env=clumps_session.env) as session:
+        with Tools(session=session) as tools:
+            # A new mapset starts with its own default region, so align it
+            # with the raster built in PERMANENT.
+            tools.g_region(raster="clumps")
+            yield SimpleNamespace(tools=tools, input="clumps")
--- a/src/raster/r.centroids/tests/r_centroids_test.py
+++ b/src/raster/r.centroids/tests/r_centroids_test.py
@@ -1,0 +1,57 @@
+"""Tests for the r.centroids addon."""
+
+import pytest
+
+from grass.exceptions import CalledModuleError
+
+# cat -> (x range, y range) of the source quadrant, for the loose
+# extent-containment check below.
+QUADRANTS = {
+    1: ((0, 3), (7, 10)),
+    2: ((7, 10), (7, 10)),
+    3: ((0, 3), (0, 3)),
+    4: ((7, 10), (0, 3)),
+}
+
+
+def test_output_has_one_point_per_clump_excluding_background(centroids):
+    """One centroid is produced per non-zero clump value; 0 is background."""
+    tools = centroids.tools
+    tools.r_centroids(input=centroids.input, output="out_centroids")
+
+    info = tools.v_info(map="out_centroids", flags="t", format="json")
+    assert info["points"] == 4
+
+    records = tools.v_db_select(map="out_centroids", format="json")["records"]
+    assert {row["cat"] for row in records} == {1, 2, 3, 4}
+
+
+def test_volume_columns_are_dropped(centroids):
+    """r.volume's volume/sum/count/average columns are removed."""
+    tools = centroids.tools
+    tools.r_centroids(input=centroids.input, output="out_columns")
+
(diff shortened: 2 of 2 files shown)


echoix:
Do you know if these new pytest tests actually run in CI now? I’m not sure yet, unless it’s very recent

src/raster/r.centroids/tests/r_centroids_test.py
```diff
@@ -0,0 +1,57 @@
...
+
+    columns = tools.v_info(map="out_columns", flags="c", format="json")
+    assert {col["name"] for col in columns} == {"cat"}
+
+
+def test_overwrite_protection(centroids):
+    """Reusing an output name without --overwrite fails."""
+    tools = centroids.tools
+    tools.r_centroids(input=centroids.input, output="out_overwrite")
+    with pytest.raises(CalledModuleError):
```
wenzeslaus: Consider catching `ToolError` (from `grass.tools`) instead of the broader `CalledModuleError`, since these tests drive the `Tools` interface, which raises `ToolError`:

```python
from grass.tools import ToolError
...
with pytest.raises(ToolError):
    tools.r_centroids(input=centroids.input, output="out_overwrite")
```

`ToolError` subclasses `CalledModuleError`, so the current code works — but asserting the specific type is more precise and self-documenting. `from grass.tools import ToolError` resolves via the package's lazy `__getattr__`, on the same GRASS versions as `Tools` (8.5+), so it adds no version constraint.

_🤖 AI-drafted, read by me before posting._

src/raster/r.centroids/tests/conftest.py
```diff
@@ -0,0 +1,44 @@
...
+                    " if(row() <= 3 && col() >= 8, 2,"
+                    " if(row() >= 8 && col() <= 3, 3,"
+                    " if(row() >= 8 && col() >= 8, 4, 0))))"
+                )
+            )
+        yield session
+
+
+@pytest.fixture
+def centroids(clumps_session):
```
wenzeslaus: Minor naming: the fixture is named `centroids`, but it yields `SimpleNamespace(tools=..., input="clumps")` — a session handle plus the input *clumps* raster, not centroids. Since "centroids" is also the module's output concept, it reads a little confusingly at the call sites (`centroids.tools`, `centroids.input`, where `.input` is the clumps map). A name like `clump_setup`, `clumps`, or `session` would be clearer.

_🤖 AI-drafted, read by me before posting._

src/raster/r.centroids/tests/r_centroids_test.py
```diff
@@ -0,0 +1,57 @@
+"""Tests for the r.centroids addon."""
+
+import pytest
+
+from grass.exceptions import CalledModuleError
+
+# cat -> (x range, y range) of the source quadrant, for the loose
+# extent-containment check below.
+QUADRANTS = {
```
wenzeslaus: Because the fixture builds symmetric 3×3 blocks, each centroid is fully deterministic, so this could assert exact coordinates instead of a loose bounding box — catching placement regressions the bbox check would miss. Exact expected points (region 10×10, res 1):

```python
EXPECTED = {1: (1.5, 8.5), 2: (8.5, 8.5), 3: (1.5, 1.5), 4: (8.5, 1.5)}
...
points = {int(cat): (float(x), float(y))
          for x, y, cat in (line.split("|") for line in lines)}
assert points == pytest.approx(EXPECTED)
```

An `EXPECTED` dict like this also subsumes the per-clump count / cat-set assertion and removes the need for `QUADRANTS` entirely.

Caveat worth a comment in the test: exact values couple the test to r.volume's centroid algorithm. For these symmetric odd-sized blocks it's stable (the distance-weighted centroid lands on the clump's middle cell and is used directly), but even-sized or irregular clumps can hit r.volume's counting/adjusted fallback and shift the point.

_🤖 AI-drafted, read by me before posting._

wenzeslaus (commented):
_These review comments were drafted with AI assistance (Claude Code) and read by me (a human) before posting._

---

**Overlap with the existing `testsuite/` — worth analyzing before removing anything**

This adds a pytest suite alongside the existing gunittest `testsuite/`, so it's worth being explicit about the intent: is this meant to eventually *replace* the gunittest tests, or coexist? A couple of things to weigh:

1. **Only the gunittest suite runs in CI today.** The addons CI (`.github/workflows/test.sh`) runs `grass.gunittest.main`, which discovers `testsuite/` directories only — there is no pytest step in the repo's workflows. So this new suite won't execute in CI as-is, and deleting `testsuite/` now would *drop* real coverage rather than move it.
2. **The two suites cover genuinely different things.** The gunittest test is a golden-reference (`assertVectorEqualsAscii`) check against the real NC `basins` raster — irregular, real-world clumps. The new suite tests granular wrapper behaviors (column drop, one-point-per-clump, overwrite) on synthetic 3×3 blocks. Those symmetric blocks only exercise r.volume's distance-weighted centroid path; irregular clumps can hit the counting/adjusted fallback path, so the golden test has coverage the synthetic fixture doesn't reproduce.

Suggested path: keep `testsuite/` for now and note the migration intent. If the goal is to converge on one framework, do it as a follow-up that (a) adds a pytest job to CI first, then (b) *ports* the real-data/golden coverage into pytest rather than just deleting it.

christianbraun:
Confirmed — this pytest suite doesn't run in CI yet. No pytest step exists in .github/workflows/; test.sh only runs grass.gunittest.main against testsuite/. Same situation as the r.clip pytest suite (#1784), which was merged with the same gap.

christianbraun:
Agreed — no intent to remove or replace testsuite/ here. It stays as the CI-running golden-data check against the real NC basins raster; this new suite just adds finer-grained coverage of the wrapper's own behavior (column drop, one-point-per-clump, overwrite) on synthetic data. Converging the two frameworks (CI job first, then porting the golden coverage) is a separate follow-up, not in scope for this PR.

wenzeslaus:
On pytest in CI: I'm working on it in #1769 (and other PRs).

wenzeslaus (commented):
_AI review:_

One new thing to fix, introduced by the exact-coordinate change — see the inline comment on the `pytest.approx` assertion.

src/raster/r.centroids/tests/r_centroids_test.py
```diff
@@ -0,0 +1,53 @@
...
+    """Each centroid lands on its source clump's exact expected point."""
+    tools = clump_setup.tools
+    tools.r_centroids(input=clump_setup.input, output="out_extent")
+
+    lines = tools.v_out_ascii(input="out_extent", format="point").stdout.splitlines()
+    points = {
+        int(cat): (float(x), float(y))
+        for x, y, cat in (line.split("|") for line in lines)
+    }
+    assert points == pytest.approx(EXPECTED)
```
wenzeslaus: _🤖 AI written:_

`pytest.approx` doesn't recurse into the tuple values of a dict, so `points == pytest.approx(EXPECTED)` is effectively an **exact** comparison — the tolerance is dead. I confirmed on pytest 7.4.4: with one coordinate perturbed by `1e-10` (well within the default `rel=1e-6`), `dict == pytest.approx(dict-of-tuples)` returns `False`, not `True`. It passes today only because the coordinates are exact half-integers that `v.out.ascii` prints exactly; if that formatting ever changed to e.g. `1.4999999`, the test would fail despite being within any sane tolerance.

(That one's on me — my earlier suggestion used this form.) Two clean options:

Apply `approx` per clump — tuples work correctly as sequences, so the tolerance is real:
```python
assert points.keys() == EXPECTED.keys()
for cat, expected_point in EXPECTED.items():
    assert points[cat] == pytest.approx(expected_point)
```

Or, since the values are exact and stable, drop `approx` and compare directly:
```python
assert points == EXPECTED
```

I'd lean toward the per-key `approx` form to stay robust against `v.out.ascii`'s output precision.

wenzeslaus: Result of in-person discussion: We want that approx, so we need to recurse here (option 1).

wenzeslaus:
On testsuite (discussed in-person): Keep it as is. Currently, it makes use of NC SPM real-world data and has specific reference test data to check regressions.

wenzeslaus (approved):
Great. Thank you. Let's go with this.

PR #1787, opened 2026-07-14
v.median: Fix stdout output and add pytest suite

christianbraun:
Fixes a bug in `v.median`'s stdout output and adds a pytest suite under `tests/`.

**Fix:** the `-` output mode wrote the result via `grass.message()`, which routes to stderr, not stdout. Switched to `print()` so the module's data output actually reaches stdout. `v.median out=-` previously produced no stdout at all.

**Tests:** cover the median written to a vector map, the median printed to stdout, that stdout mode creates no map, overwrite protection, and the even-count averaging case.

### DIFF
--- a/src/vector/v.median/v.median.py
+++ b/src/vector/v.median/v.median.py
@@ -71,7 +71,9 @@ def main():
     overwrite = os.getenv("GRASS_OVERWRITE")
     # if output is not set return to stdout
     if map_name == "-":
-        grass.message(output)
+        # Data output goes to stdout via print(); grass.message() is for
+        # informational messages, which are sent to stderr.
+        print(output)
     # else
     else:
         # output file
--- a/src/vector/v.median/v.median.py
+++ b/src/vector/v.median/v.median.py
@@ -71,8 +71,6 @@ def main():
     overwrite = os.getenv("GRASS_OVERWRITE")
     # if output is not set return to stdout
     if map_name == "-":
-        # Data output goes to stdout via print(); grass.message() is for
-        # informational messages, which are sent to stderr.
         print(output)
     # else
     else:
--- a/src/vector/v.median/tests/conftest.py
+++ b/src/vector/v.median/tests/conftest.py
@@ -1,0 +1,48 @@
+"""Fixtures for the v.median tests."""
+
+import os
+from types import SimpleNamespace
+
+import pytest
+
+import grass.script as gs
+from grass.experimental import TemporaryMapsetSession
+from grass.tools import Tools
+
+# Distinct x's and y's so the per-axis median lands on one input point (odd
+# count) or the average of the two middle points (even count).
+ODD_POINTS = [(1, 10), (2, 20), (3, 30), (4, 40), (5, 50)]
+EVEN_POINTS = [(1, 10), (2, 20), (3, 30), (4, 40)]
+
+
+def base_session(tmp_path_factory):
+    """Build a project with odd- and even-count point maps once per module."""
+    project = tmp_path_factory.mktemp("v_median") / "project"
+    gs.create_project(project, epsg="4326")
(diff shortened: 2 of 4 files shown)


src/vector/v.median/v.median.py
```diff
@@ -71,7 +71,9 @@ def main():
     overwrite = os.getenv("GRASS_OVERWRITE")
     # if output is not set return to stdout
     if map_name == "-":
-        grass.message(output)
+        # Data output goes to stdout via print(); grass.message() is for
+        # informational messages, which are sent to stderr.
```
wenzeslaus: Just drop this comment. No need. This is info for commit message, but not to preserve in the code.

wenzeslaus (changes requested):

wenzeslaus:
Looks good. Thanks. Here is a review I co-prepared with AI:

The fix checks out against core: `grass.message()` runs `g.message`, which writes to **stderr** (`grass/script/core.py`), while `Tools(...).text` reads **stdout** only (`grass/tools/support.py`), so the old `-` path really did drop the data output on the floor. The switch to `print()` is the correct GRASS pattern, and `test_stdout_output_prints_median` genuinely fails without it. The suite is also well isolated — module-scoped base project plus a throwaway `TemporaryMapsetSession` per test.

A few small suggestions:

### 1. Frame this as a fix, not only tests
The `print()` change is a user-facing bug fix (`v.median out=-` produced no stdout before), so it deserves to be visible in the history rather than buried under "Add pytest suite". Consider retitling, e.g.:

> **v.median: Fix stdout output and add pytest suite**

and mentioning the fix first in the description.

### 2. Also assert the stdout path writes no map
Nice complement to the fix. With a real `output=` name the module writes a single-point vector (one point at the median x/y, via `v.in.ascii`); with `output="-"` it should print that line and create nothing. Since each test runs in its own mapset and the base maps live in PERMANENT, the current mapset should stay empty:

```python
def test_stdout_output_creates_no_map(median):
    median.tools.v_median(input="odd_points", output="-")
    assert median.tools.g_list(type="vector", mapset=".").text_split() == []
```

### 3. Name the file like its siblings
The other `vector/*/tests/` suites use the `test_*.py` form (`test_v_rast_move.py`, `test_v_in_ags.py`). Renaming `v_median_test.py` → `test_v_median.py` keeps the directory consistent. (Both forms match core's `python_files`, so this is purely for consistency.)

### 4. Inline the single-use conftest helpers
`base_session` and `median_fixture` are each short and called from exactly one fixture, so the extra indirection isn't buying anything — folding them into `module_session`/`median` reads a bit more directly:

```python
@pytest.fixture(scope="module")
def module_session(tmp_path_factory):
    """Build a project with odd- and even-count point maps once per module."""
    project = tmp_path_factory.mktemp("v_median") / "project"
    gs.create_project(project, epsg="4326")
    with gs.setup.init(project, env=os.environ.copy()) as session:
        with Tools(session=session) as tools:
            for name, points in (("odd_points", ODD_POINTS), ("even_points", EVEN_POINTS)):
                ascii_file = tmp_path_factory.mktemp(name) / "points.txt"
                ascii_file.write_text("\n".join(f"{x}|{y}" for x, y in points))
                tools.v_in_ascii(input=str(ascii_file), output=name, separator="pipe")
        yield session


@pytest.fixture
def median(module_session):
    """Isolated per-test mapset + Tools handle over the module-scoped base session."""
    with TemporaryMapsetSession(env=module_session.env) as session:
        with Tools(session=session) as tools:
            yield SimpleNamespace(tools=tools, env=session.env)
```

All minor — the PR is in good shape. Thanks for catching and fixing the stderr/stdout bug while you were in there.

wenzeslaus (approved):
Thank you. Great standardization of the behavior and good test for the tool.

PR #1796, opened 2026-07-17
CI: New workflow for compiling addons on Windows

ant9000:
Added a new workflow for compiling the grass-addons bundle using Github CI. Used the following script for the actual module compilation:

https://github.com/landam/wingrass-maintenance-scripts/blob/master/addons/compile.sh

The script itself could use some updating - it survived from the SVN era - but for now is included as is.


echoix:
Will need https://github.com/OSGeo/grass-addons/pull/1797 and a little tweak too

.github/workflows/ci-windows.yml
```diff
@@ -0,0 +1,205 @@
...
+            install: >-
+              bison
+              diffutils
+              dos2unix
+              flex
+              git
+              libintl
+              make
+              tar
+              tree
```
echoix:
```suggestion
```

.github/workflows/ci-windows.yml
```diff
@@ -0,0 +1,204 @@
...
+              python3-core
+              python3-jupyter
+              python3-matplotlib
+              python3-numpy
+              python3-pip
+              python3-pywin32
+              python3-wxpython
+              sqlite3-devel
+              zstd-devel
+
```
echoix:
```suggestion
      - name: MSYS2 setup
        uses: msys2/setup-msys2@66cd2cce69caa17b53920067426061ca1de3a884 # v2.32.0
        id: msys2
        with:
          path-type: inherit
          location: D:\
          update: true
          msystem: MINGW64
          install: >-
            bison
            diffutils
            dos2unix
            flex
            git
            libintl
            make
            tar
            zip
          pacboy: >-
            bzip2
            ccache
            fftw
            gcc
            gettext
            libiconv
            libsystre
            libtre
            libwinpthread-git
            openblas
            pcre
            pkgconf
            toolchain
            zlib

      - name: Setup OSGeo4W environment
        uses: echoix/setup-OSGeo4W@1e711832882e03603e9013181a87885b10af8a98 # v0.3.0
        id: osgeo4w
        with:
          package-dir: "D:/OSGeo4W_pkg"
          packages: |
            cairo-devel
            freetype-devel
            gdal-devel
            geos-devel
            libjpeg-turbo-devel
            liblas-devel
            libpng-devel
            libpq-devel
            libtiff-devel
            netcdf-devel
            proj-devel
            python3-core
            python3-jupyter
            python3-matplotlib
            python3-numpy
            python3-pip
            python3-pywin32
            python3-wxpython
            sqlite3-devel
            zstd-devel

```

echoix (commented):
Dropping the use of parallel, as linting tools aren't ready yet here

.github/workflows/ci-windows.yml
```diff
@@ -0,0 +1,202 @@
...
+             for f in `ls bin/*.bat 2> /dev/null` ; do
+               echo $f
+               sed "s/GISBASE/GRASS_ADDON_BASE/" $f > tmp
+               mv tmp $f
+             done
+             rm -rf grass-addons
+             zip -r ../$mod.zip *
+             cd ..
+             rm -rf $mod
+             md5sum $mod.zip > ${mod}.md5sum
```
echoix:
```suggestion
            mod=${d%%/}
            if [ $mod == "logs" ] ; then
              continue
            fi
            cd $mod
            echo $mod
            for f in `ls bin/*.bat 2> /dev/null` ; do
              echo $f
              sed "s/GISBASE/GRASS_ADDON_BASE/" $f > tmp
              mv tmp $f
            done
            rm -rf grass-addons
            zip -r ../$mod.zip *
            cd ..
            rm -rf $mod
            md5sum $mod.zip > ${mod}.md5sum
```
Change indenting by 2 instead of 3

.github/workflows/ci-windows.yml
```diff
@@ -0,0 +1,202 @@
...
+        include:
+          - grass-version: main
+            python-version: "3.11"
+            os: windows-2022
+          - grass-version: releasebranch_8_5
+            python-version: "3.11"
+            os: windows-2022
+      fail-fast: false
+
+
```
echoix:
```suggestion

```

echoix (approved):

ecodiv:
@echoix  @ant9000 Compiling of addons fail with the message:

`Error: File:[/github/workspace/.github/workflows/windows_addons_compile.sh] is not executable`

echoix:
> @echoix  @ant9000 Compiling of addons fail with the message:
> 
> 
> 
> `Error: File:[/github/workspace/.github/workflows/windows_addons_compile.sh] is not executable`

On what OS? It's supposed to run on an msys2 / mingw64 shell, on windows.

echoix:
> @echoix  @ant9000 Compiling of addons fail with the message:
> 
> 
> 
> `Error: File:[/github/workspace/.github/workflows/windows_addons_compile.sh] is not executable`

Either way, the solution would be to `chmod +x` that file, and commit that. It is the only Linux file mode that is preserved by git. But it isn't useful on Windows. There's a way to do it using git only, but I don't remember the exact syntax

ecodiv:
> > @echoix  @ant9000 Compiling of addons fail with the message:
> > `Error: File:[/github/workspace/.github/workflows/windows_addons_compile.sh] is not executable`
> 
> Either way, the solution would be to `chmod +x` that file, and commit that. It is the only Linux file mode that is preserved by git. But it isn't useful on Windows. There's a way to do it using git only, but I don't remember the exact syntax

The [General linting / GitHub Super Linter test on Github fails with that message

echoix:
Should be fixed now

PR #1802, opened 2026-07-22
r.stream.basins: fix ad-hoc size limitation of the outlets

mastershubham:
I wanted to adapt the analysis for country level or even continental level DEM processing. The earlier limit was ad-hoc and served for small to medium area of interest analysis.


mastershubham:
I have included the change and submitted another pull request.

PR #1805, opened 2026-07-23
r.stream.basins: removed ad-hoc limits on stream size to facilitate large DEM processing

mastershubham:
r.stream.basins/basins_inputs.c had a bug where there was ad-hoc limits placed on the largest size of the stream network. Also, the variable meant to be updated was not done so in a complete manner. This made this module work for smaller basins but for basins of the order of ~150,000 sq.kms. and larger(at resolution of 30 m), this module was throwing an error. 
I handled the update of the variable and now this module is suitable even for larger basins.


mastershubham:
@neteler Please allow the workflow approval for this PR and review the changes.

neteler:
> review the changes.

The PR lacks a description.

mastershubham:
> > review the changes.
> 
> The PR lacks a description.

I have provided a description. Please now have a look. Thanks for your time.

mastershubham:
@wenzeslaus Please verify and merge this PR. Thanks.

cwhite911 (changes requested):
@mastershubham the fix is correct. `out_max` is the allocated capacity of outlets, and the original code realloc'd to `out_max * 6` without writing the new size back, so after the first growth out_max no longer described the allocation. Capacity stayed frozen at the initial `ncols + nrows` and the fatal at `6 * (out_max - 1)` became a hard ceiling on node count. r.stream.distance/distance_init.c has the same structure with the `out_max *= 4` line present. However, I'm not sure if this limit was intentionally or not.

Two things worth adding to the record.

First, this is not specific to large DEMs. The ceiling scales with region perimeter while node count scales with network density, so it is reachable on a small region. 

On 400x400 the cap is 6 * 799 = 4794 nodes:

```sh
g.region rows=400 cols=400 res=1
r.surf.fractal output=elev dimension=2.05 seed=42
r.watershed elevation=elev threshold=5 stream=streams drainage=drain
r.stream.basins direction=drain stream_rast=streams basins=b
```
That gives about 6100 stream categories. Before this PR it fails with "Stream and direction maps probably do not match" in both the default and -m paths, on maps that match perfectly. With the PR both complete. At threshold=10 the same region stays under the cap and the output is identical before and after, so there is no change for inputs that already worked.

Second, on removing the fatal: _r.stream.stats/stats_prepare.c_ already uses exactly this pattern, `out_max *= 2` with a realloc and no node-count check. This brings _r.stream.basins_ in line with the family.

Could you add a regression test? The module has no tests/ or testsuite/ today, and the recipe above runs in a few seconds, so a pytest case under _src/raster/r.stream.basins/tests/_ that builds a dense network and asserts the tool completes would cover this directly.

One more: grass-addons squashes on merge and uses the PR title as the commit message. GRASS uses the imperative mood, so something like _"r.stream.basins: Remove arbitrary limit on outlet count"_ fits the convention better than "removed".

PR #1830, opened 2026-08-17
i.landsat8.swlst: fix namedtuple use with duplicated keys

pesekon2:
Since the creation of the addon, the metadata have changed and they include some duplicated keys. However, these keys are not used within the addon - this simply renames them to make the addon work.


NikosAlexandris:
Please go ahead and apply the fix.

pesekon2:
Thanks @NikosAlexandris.

PR #1843, opened 2026-09-02
t.rast.stl: Improving error message

ecodiv:
Improving error message when using the wrong input for frequency parameter


echoix (approved):

ecodiv:
Thanks @echoix

PR #1846, opened 2026-09-12
r.stream.basins: Fix heap buffer overflow for out-of-bounds cats

### LINKED ISSUE #1828: [Bug] r.stream.basins: Heap buffer overflow when cats= is one past the maximum stream category
Passing a stream category one greater than the map's maximum writes four bytes past the end of the categories array. Both the RAM and segmented paths are affected.

_main.c_ `computes max_number_of_streams = (int)map_streams.max + 1` (lines 191 and 250) and passes it as `number_of_streams`. _basins_inputs.c_ allocates that many ints, so valid indices are 0 through map_streams.max:

```c
categories = G_malloc((number_of_streams) * sizeof(int));
memset(categories, -1, (number_of_streams) * sizeof(int));

for (i = 0; cat_list[i] != NULL; ++i) {
    cat = atoi(cat_list[i]);
    if (cat < 1 || cat > number_of_streams)  
        G_fatal_error(_("Stream categories must be > 0 and < maximum "
                                    "stream category"));
    categories[cat] = cat;
}
```
The bound is inclusive, so `cat == number_of_streams` passes validation and writes to `categories[map_streams.max + 1]`, one element past the allocation. This is _basins_inputs.c_:114 in `ram_process_streams` and _basins_inputs.c:207_ in `seg_process_streams`.

## Reproduction

```sh
g.region rows=60 cols=60 res=1
r.surf.fractal output=elev dimension=2.05 seed=7
r.watershed elevation=elev threshold=20 stream=streams drainage=drain
r.info -r streams        # max=134 in this example
r.stream.basins direction=drain stream_rast=streams basins=b cats=135
```

Built with `-fsanitize=address`:

```sh
ERROR: AddressSanitizer: heap-buffer-overflow
WRITE of size 4
     #0 ram_process_streams basins_inputs.c:114
```

The same happens at basins_inputs.c:207 with -m. Passing cats=134 is clean in

## Suggested fix

Make the bound exclusive, since number_of_streams is already max + 1:

```c
if (cat < 1 || cat >= number_of_streams)
```
After the fix the accepted range is 1 throug inclusive, which "< maximum stream category" understates. Something like ` _("Stream category must be between 1 and %d")` with the actual maximum would be clearer.

Found while reviewing #1805, which does not touch this code.
GRASS 8.6.0dev, addon r.stream.basins, current grass8.

bhuvan-somisetty:
## Description
In `r.stream.basins`, `categories` is allocated with size `number_of_streams` (where `number_of_streams = (int)map_streams.max + 1`). Valid index bounds for the `categories` array are `0` through `number_of_streams - 1` (up to `map_streams.max`).

Previously, the category bounds validation in `ram_process_streams` and `seg_process_streams` used an inclusive upper bound check:
```c
if (cat < 1 || cat > number_of_streams)
    G_fatal_error(_("Stream categories must be > 0 and < maximum stream category"));
categories[cat] = cat;
```
When `cat == number_of_streams` (`map_streams.max + 1`), it passed validation and executed `categories[number_of_streams] = cat;`, writing 4 bytes past the allocated heap buffer and causing AddressSanitizer heap buffer overflow errors.

This PR:
1. Fixes the upper bound check to `cat >= number_of_streams` in both `ram_process_streams` and `seg_process_streams`.
2. Updates the fatal error message to explicitly state the valid category range `1` to `number_of_streams - 1`:
   ```c
   G_fatal_error(_("Stream category must be between 1 and %d"),
                 number_of_streams - 1);
   ```

## Fixes
Fixes #1828

## Testing & Validation
- Verified bounds checking logic across edge cases (`cat < 1`, `cat == max_category`, `cat == max_category + 1`, and `cat > max_category + 1`).
- Checked code formatting and line endings via `pre-commit` (`clang-format`, `editorconfig-checker`, etc.).

Signed-off-by: bhuvan-somisetty <somisettybhuvan5@gmail.com>


echoix (approved):
Seems correct, and an rough AI review thinks the same as I do (by looking back at main.c)

PR #1848, opened 2026-09-15
r.stream.basins: Fix silent basin -1 for outlet points without category

### LINKED ISSUE #1847: [Bug] r.stream.basins: outlet points without a category in layer 1 silently become basin -1
**Name of the addon**
r.stream.basins

**Describe the bug**
When outlets are given as a vector points map (`points=`), the code reads each point's category with `Vect_cat_get()` but never checks the return value. If a point has no category in layer 1, `Vect_cat_get()` sets the category to -1 and returns 0 to signal that, but that -1 is written straight into the output basin raster instead of being caught as an error.

In `src/raster/r.stream.basins/basins_inputs.c`, function `process_vector()`:

```c
Vect_cat_get(cats, 1, &cat);

outlets[i].r = (int)Rast_northing_to_row(sites->y[0], &window);
outlets[i].c = (int)Rast_easting_to_col(sites->x[0], &window);
outlets[i].val = cat;
```

That value is later written directly into the output map in `basins_fill.c`:

```c
basins[outlets[i].r][outlets[i].c] = outlets[i].val;
```

So a point with no category quietly becomes basin -1 instead of producing an error. Worse, if more than one outlet point lacks a category, they all collide on -1 and get merged into a single basin, silently producing wrong results. The manual page says "Every point shall have its own unique category," but this isn't enforced anywhere.

The same pattern (unchecked `Vect_cat_get()`) also shows up in `read_points()` in `src/raster/r.stream.snap/points_io.c`, so it may be worth checking there too.

**To Reproduce**

1. Set up a region and generate a stream network:
   ```sh
   g.region rows=60 cols=60 res=1
   r.surf.fractal output=elev dimension=2.05 seed=7
   r.watershed elevation=elev threshold=20 stream=streams drainage=drain
   ```
2. Take a points map of outlets and remove the category from at least one point, e.g.:
   ```sh
   v.category input=some_outlets output=outlets_nocat option=del cat=2
   ```
3. Run:
   ```sh
   r.stream.basins direction=drain points=outlets_nocat basins=b
   r.category b
   ```
4. See a basin labeled -1 in the output instead of an error or warning about the missing category.

**Expected behavior**
`r.stream.basins` should reject or warn about outlet points that have no category in layer 1, instead of silently writing -1 into the basin output. Multiple uncategorized points should not be allowed to collide and merge into one basin.

- Operating System: any
- GRASS GIS version: current grass8 / addons main branch

**Additional context**
Found while going over `basins_inputs.c` after fixing the heap buffer overflow in the `cats=` bound check (#1828), which is a different function in the same file.

bhuvan-somisetty:
## Summary

`r.stream.basins` reads outlet categories from a vector points map with `Vect_cat_get()`, but the return value was never checked. When a point has no category in layer 1, `Vect_cat_get()` sets the category to -1 and returns 0 to signal that, and that -1 was written straight into the outlet list and then into the output basin raster.

A single outlet point with no category silently produced a basin labeled -1 instead of an error. If more than one outlet point lacked a category, they all collided on the same -1 value and got merged into a single basin, which is a silently wrong result. This contradicts the documented requirement that every point must have its own unique category.

This adds a check on the return value of `Vect_cat_get()` in `process_vector()` and fails with a clear error message when an outlet point has no category in layer 1.

Fixes #1847

## Test plan

- Reviewed the fix against the GRASS core implementation of `Vect_cat_get()` (`lib/vector/Vlib/cats.c`) to confirm it returns 0 and sets the category to -1 when no category is found for the given field.
- Ran `clang-format` on the changed file, no differences.
- Manually traced the reproduction from the issue: generate a stream network, remove the category from an outlet point with `v.category option=del`, run `r.stream.basins` with `points=`, and confirmed the code path now hits the new `G_fatal_error()` instead of writing -1 into the output.


bhuvan-somisetty:
@echoix PTAL

echoix:
I don't understand enough how it works to judge well. Can someone else read a bit ? It's an addon, it doesn't need to be perfect

bhuvan-somisetty:
Fair enough, thanks for taking a look. I went ahead and also fixed the same unchecked `Vect_cat_get()` pattern in `r.stream.snap/points_io.c`, which the linked issue flagged but wasn't in scope of the original diff. Should be a straightforward, low-risk change now - happy to have someone else confirm before merging.

